PR #9, opened 2013-09-14
Autocomplete

mark-rushakoff:
Add a way to enable bash autocompletion.

Right now it only autocompletes the first command, but there's nothing explicitly stopping us from enabling tab-completions on arguments or options to sub-commands too.


tsaleh:
Would love to see this picked up again.  Not sure if generating the completion code at runtime is the right path - maybe at compile time would be better (which is what the [codegangsta/cli](https://github.com/codegangsta/cli#bash-completion) code seems to be doing).

PR #315, opened 2014-12-17
Improve french i18n

bclozel:
Here are some i18n improvements in the fr_FR translation.
Some choice of words are quite hard to make, since some english words are quite common amongst developers while their french translation "sound weird".

Note: I'm a french native speaker and a Pivotal employee.


simonleung8:
Thanks for the Pull Request. Merged

PR #317, opened 2014-12-23
Fix the invalid memory address during bind service

SrinivasChilveri:
Solution to the bug:- [#79267756]
During Bind a service instance to an app if cli receives a non http error
then err parameter is nil, so cli is panic because its accessing the nil pointer err
so to over come that use a new httperr variable inside the if condition


jberkhahn:
Thanks for the PR, merged.

PR #319, opened 2015-01-05
Fixed issue 282

diatmpravin:
cf push commands will print usage if you use them without arguments now.


simonleung8:
Thanks for the PR, however, we cannot merge this PR for the following reason.

Currently the `cf push` command is designed to allow no argument. It uses the manifest file when no argument is supplied and we don't intend to change the current behavior as it would be a breaking change to many users.

Issue #318, opened 2015-01-04
cf curl don't pass custom headers to new request for 302 respose


tsl0922:
I'm using the http `Range` headers on `cf curl` to get partial of a file, but it doesn't work.

`CF_TRACE` output:

```
cf curl /v2/apps/e0a39c2b-aab7-4a19-8463-7ec0dbbfba14/instances/0/files/logs/staging_task.log -H 'Range: bytes=10-1024' -v

REQUEST: [2015-01-04T11:11:38+08:00]
GET /v2/apps/e0a39c2b-aab7-4a19-8463-7ec0dbbfba14/instances/0/files/logs/staging_task.log HTTP/1.1
Host: api.coding.io
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
Content-Type: application/json
Range: bytes=10-1024
User-Agent: go-cli 6.8.0-b15c536 / darwin

REQUEST: [2015-01-04T11:11:38+08:00]
GET /instance_paths/4a02cda394e942fabc6a6a1169d66f07?hmac=696448e2c35e2f3076a9305d725ede77611c907f6fb2104c6be64ed6688a325bf9cc4e67f05de91cfead258159070715390828eb6ebe72d09639e8a3bc60c9e3&path=logs%2Fstaging_task.log&timestamp=1420341098 HTTP/0.0
Host: 75414404eac549d9b713608b8987038a.coding.io
Authorization: [PRIVATE DATA HIDDEN]

RESPONSE: [2015-01-04T11:11:38+08:00]
HTTP/1.1 200 OK
Content-Length: 2320
Accept-Ranges: bytes
Content-Type: text/plain; charset=utf-8
Date: Sun, 04 Jan 2015 03:11:38 GMT
Last-Modified: Sat, 03 Jan 2015 15:41:22 GMT
```

jberkhahn:
This issue has been fixed with the merging of https://github.com/cloudfoundry/cli/pull/321.

PR #321, opened 2015-01-06
Copy original request's headers when handling redirect

uzzz:
Hi

This PR is about #318 issue. I think it's good idea to copy all headers (except POST-specific) from original request, at least curl does it when flag L (Follow redirects) given. What do you think about it?

```
curl -v -L -H 'X-Test-Header: value' http://google.com

> GET / HTTP/1.1
> User-Agent: curl/7.37.1
> Host: google.com
> Accept: */*
> X-Test-Header: value
> 
< HTTP/1.1 302 Found
< Location: http://www.google.by/
[...omited...]

> GET / HTTP/1.1
> User-Agent: curl/7.37.1
> Host: www.google.by
> Accept: */*
> X-Test-Header: value
> 
< HTTP/1.1 200 OK
< Date: Tue, 06 Jan 2015 09:07:22 GMT
[...omited...]
```

### DIFF
--- a/cf/net/http_client.go
+++ b/cf/net/http_client.go
@@ -33,10 +33,19 @@ func PrepareRedirect(req *http.Request, via []*http.Request) error {
 	}
 
 	prevReq := via[len(via)-1]
-	req.Header.Set("Authorization", prevReq.Header.Get("Authorization"))
+	copyHeaders(prevReq, req)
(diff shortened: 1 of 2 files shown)


jberkhahn:
Thanks for the pull request, merged.

PR #322, opened 2015-01-07
Updating go vet location in install-dev-tools target, as it has moved.

robdaemon:
Fixes this error:

can't load package: the code.google.com/p/go.tools/cmd/vet command has
moved; use golang.org/x/tools/cmd/vet instead.

### DIFF
--- a/makefile
+++ b/makefile
@@ -65,7 +65,7 @@ install-dev-tools:
 	@echo Installing development tools into $(GODEPS)
 	go get github.com/onsi/ginkgo/ginkgo
 	go get github.com/onsi/gomega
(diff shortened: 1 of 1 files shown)


robdaemon:
Individual CLA for HP is already signed and on file, I'll get my individual one signed and emailed in tonight.

simonleung8:
Thanks for the PR, merged.

robdaemon:
@simonleung8 - this looks like it was closed without merging the change?

simonleung8:
You are right, my mistake, it is really merged this time.
thanks for the look out.

robdaemon:
I've done that more than once. Thanks!

PR #333, opened 2015-01-19
CLI sends async request for service instance provisioning

jeffh:
CLI displays the state when viewing all or detailed information of service instances.


simonleung8:
looks good. merged.

PR #338, opened 2015-01-22
Fix makefile for go-bindata 3.x

aaronorosen:
go-bindata removed the option ignore and thus if one was building with
go-bindata 3.x it would result in the error:
"flag provided but not defined: -ignore go-bindata"


simonleung8:
Can you give more context on how to reproduce the `-ignore` flag error? We are trying reproduce the problem on v3.x and `-ignore` seems to be a valid flag.

jberkhahn:
Hey there,

We haven't heard back from you so we assume everything is working on your end. Please feel free to reopen the issue if you continue to experience this error.

aaronorosen:
Hi, sorry I missed the first message. If you try to build this on 14.04 it should fail without this change.

scottruitt:
@aaronorosen please reopen the issue and we'll take another look.

simonleung8:
@aaronorosen There is no mention of `-ignore` is deprecated in go-bindata's doc, it seems to be supported still.  https://github.com/jteeuwen/go-bindata. We also tried -ignore flag in 14.04 and it worked for us.

It seems to me that the error msg "flag provided but not defined: -ignore go-bindata" indicates `-ignore go-bindata` a flag as a whole

PR #339, opened 2015-01-22
Async Service Provisioning

jeffh:
- Changed formatting of state value
- Fixed bug where `cf services` did not parse the state and state_description from the spaces summary response
- Now with more colors


jberkhahn:
Looks good, merged.

PR #344, opened 2015-01-30
Adding a way to see Security Group Rules

SrinivasChilveri:
Before Change :  Users were not able to see Security Group Rules applied to their spaces
After Change : Users(Admin and Non-admin) will be able to see the rules associated with the security groups applied to their spaces.

$ ./cf space mySpace --security-group-rules
Getting info for space mySpace in org huawei as asif...
OK

mySpace
       Org:               huawei
       Apps:
       Domains:           10.244.0.34.xip.io
       Services:
       Security Groups:   dns
       Space Quota:

Getting rules for the security group  : dns...

   destination   :   0.0.0.0/0
   ports         :   53
   protocol      :   tcp

   destination   :   0.0.0.0/0
   ports         :   53
   protocol      :   udp

### DIFF
--- a/cf/commands/space/space_test.go
+++ b/cf/commands/space/space_test.go
@@ -112,6 +112,16 @@ var _ = Describe("space command", func() {
 			})
 		})
 
+		Context("when the security-group-rules flag is passed", func() {
+			It("it shows space information and security group rules", func() {
+				runCommand("--security-group-rules", "whose-space-is-it-anyway")
+
+				Expect(ui.Outputs).To(ContainSubstrings(
+					[]string{"Getting rules for the security group"},
+				))
+			})
+		})
+
 		Context("when the space has a space quota", func() {
 			It("shows information about the given space", func() {
 				runCommand("whose-space-is-it-anyway")
--- a/cf/commands/space/space_test.go
+++ b/cf/commands/space/space_test.go
@@ -68,8 +68,12 @@ var _ = Describe("space command", func() {
 			serviceInstance.Guid = "service1-guid"
 			services := []models.ServiceInstanceFields{serviceInstance}
 
-			securityGroup1 := models.SecurityGroupFields{Name: "Nacho Security"}
-			securityGroup2 := models.SecurityGroupFields{Name: "Nacho Prime"}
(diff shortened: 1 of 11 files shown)


asifhuawei:
This pull request solves the problem reported in https://www.pivotaltracker.com/n/projects/892938/stories/77733680

It was suggested to add this flag in either 'app' command or 'space' command but we choosed space command beacuse it was already having this information so we just need to format the data in human readable format and present it, where as in app command we were not having the security group rules information so we need to go one more round to CC to get these data which would be an overhead so we decide to go with the "space mySpace --security-group-rules".
And this flag will be accessible by both admin & non-admin users.

goehmen:
Hi Srinivas.  I move the story to the backlog & it's fairly high in priority.  I had a similar story about IP table visibilities and want to verify that your PR fulfills the requirements for that as well.

asifhuawei:
Hi @goehmen 
The present changes in this PR will show all the information of IPTables in the json format structure but if you want the data in the tabular format as illustrated in the IPTables story(#76616638)  then we will be happy to add the same feature for 'cf security-groups' and we will present the data there in tabular format.
Waiting for your approval to make changes in this PR......

goehmen:
The tacker story is now unblocked & we are good to go

cf/commands/space/space_test.go
```diff
@@ -112,6 +112,16 @@ var _ = Describe("space command", func() {
 			})
 		})
 
+		Context("when the security-group-rules flag is passed", func() {
+			It("it shows space information and security group rules", func() {
+				runCommand("--security-group-rules", "whose-space-is-it-anyway")
+
+				Expect(ui.Outputs).To(ContainSubstrings(
+					[]string{"Getting rules for the security group"},
```
simonleung8: Looks like the test only cover the title of the table. We would like the test to also assert the content of the table is correctly printed. Do you mind updating the test a little?

asifhuawei:
@simonleung8  Thanks for the review comment......we will be happy to make changes to this PR

asifhuawei:
@simonleung8  We have made the test case assertion more specific to the output, please review the changes

simonleung8:
@asifhuawei Thanks for updating the test, merged.

PR #345, opened 2015-01-31
Ginkgo matcher BeInDisplayOrder()

simonleung8:
Asserting string output order is always hard, but now it should be tons easier with this matcher.

### DIFF
--- a/cf/commands/servicebroker/service_brokers_test.go
+++ b/cf/commands/servicebroker/service_brokers_test.go
@@ -99,7 +99,15 @@ var _ = Describe("service-brokers command", func() {
 		}}
 
 		testcmd.RunCommand(cmd, []string{}, requirementsFactory)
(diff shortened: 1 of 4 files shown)


onsi:
@simonleung8 you may want to explore gbytes.  It allows you to make ordered assertions on a buffer.  For example:

```
buff := gbytes.BufferWithBytes(ui.Outputs)
Expect(buff).To(gbytes.Say("Getting service brokers as"))
Expect(buff).To(gbytes.Say("my-user"))
etc...
```

Docs are here:
http://onsi.github.io/gomega/#gbytes-testing-streaming-buffers

simonleung8:
@onsi Thanks for the pointer, the gbytes ordered assertions look good, except, maybe I missed it, but I don't see a way to assert 2 sub strings to appear in the same output line, which is useful for cli to assert output tables.

onsi:
Two things:
- it has no notion of lines
- it can take a regular expression

PR #349, opened 2015-02-10
Async

maximilien:
Adds more async goodness. Specifically around new response fashion (e.g., `last_operation`) and also service instance update and rename now pass `?accepts_incomplete=true`

Best,

@jeffh and maximilien


jberkhahn:
Ok, looks good. Merged.

PR #351, opened 2015-02-13
Finishes async work for CLI

jeffh:
- Changed "Creating service foo" to "Creating service instance foo"
- Changed "Updating service foo" to "Updating service instance foo"
- Suggested commands to run after creating a service are highlighted instead of quoted.
- `cf delete-service` passes along `accepts_incomplete=true` to perform async deprovision.

[#86668046, #88279874, #87584124]

@gtona & Jeff


simonleung8:
Looks good, merging

PR #357, opened 2015-02-20
Changed service instance commands to yellow (CommandColor).

jeffh:
[Fixes #86668046].

Added quotes around recommended commands (to be more like bind-service)


jberkhahn:
Looks good. Merged.

PR #360, opened 2015-02-23
Fix the requirmements issue in some of the application commands

SrinivasChilveri:
In some of the application related commands code the requirements are not handled properly
comparing other app commands for example cf apps return failure immediately if user has not
targetted the space as shown below but not in the following application commands
' cf delete' , 'cf rename' , 'cf logs', 'cf env','cf start', cf stop', ' cf restage'.

CLI is sending empty space guid & receiving error from cloud controller, So in ivalid conditions
CLI can validate & return error to user immediately.

htipl@cf5:~$ cf apps
 FAILED
 No space targeted, use 'cf target -s' to target a space

Before Fix: If cf user does 'cf delete testapp -f' without targetting the space

htipl@cf5:~$ cf delete testapp -f
 Deleting app testapp in org SriniOrg / space as admin...
 FAILED
 Server error, status code: 404, error code: 40004, message: The app space could not be found: apps

So CLI is sending empty space guid & receiving error from cloud controller which we can validate at CLI it self.

After Fix:If cf user does same steps
htipl@cf5:~/go/src/github.com/cloudfoundry/cli$ ./out/cf delete testapp

VERSION:
 BUILT_FROM_SOURCE

FAILED
 No space targeted, use 'cf target -s' to target a space

### DIFF
--- a/cf/commands/application/delete.go
+++ b/cf/commands/application/delete.go
@@ -47,7 +47,8 @@ func (cmd *DeleteApp) GetRequirements(requirementsFactory requirements.Factory, 
 		cmd.ui.FailWithUsage(c)
 	}
 
-	reqs = []requirements.Requirement{requirementsFactory.NewLoginRequirement()}
+	reqs = []requirements.Requirement{requirementsFactory.NewLoginRequirement(),
+		requirementsFactory.NewTargetedSpaceRequirement()}
 	return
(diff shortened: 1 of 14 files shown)


simonleung8:
Looks good, thanks for the PR. Merged

PR #363, opened 2015-02-25
Prevent updating service plans when the CC is less than v191.

jeffh:
v191 corresponds to CC API 2.16.0.

This is to prevent a bug with older CC and newer CLIs where plans can be
updated without talking to the service broker.

[#88798806, #88689444]

### DIFF
--- a/cf/api/applications/fakes/fake_application_repository.go
+++ b/cf/api/applications/fakes/fake_application_repository.go
@@ -2,6 +2,7 @@ package fakes
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/cloudfoundry/cli/cf/api/applications"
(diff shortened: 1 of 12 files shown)


jberkhahn:
Hey there,

It looks like you left a spare 'fmt' import in cf/api/applications/fakes/fake_application_repository.go  and it's causing the tests to fail. Could you remove that line?

dsabeti:
Hey Jonathan,

I've got a fix for this, but part of it is actually dependent on changes in a different pull request (https://github.com/cloudfoundry/cli/pull/357). Can you guys merge that one so I can rebase this change onto the updated master and re-submit?

Thanks,
David

jberkhahn:
Ok, #357 went through and this one looks good. Merged.

PR #365, opened 2015-02-26
Removed as admin.. clause from create-user since it is confusing.

SrinivasChilveri:
Story in CLI [#74893356].

When we create users using the CLI while logged in as the admin user, the CLI feedback for this uses the wording: Creating user as admin.... This was confusing at first glance as we though it meant that the new user was going to be an administrator rather than this meaning that we were logged in as admin.

Before Fix :
$ cf create-user abc def
Creating user abc as admin...
OK

After Fix :
 create-user abc1 def1
Creating user abc1...
OK


jberkhahn:
This looks good, merged.

PR #366, opened 2015-02-26
Fixed error message when there is a mismatch in the order of arguments for create-buildpack

SrinivasChilveri:
Story in CLI [#82598260].

Correct usage of create-buildpack is :
USAGE:
   cf create-buildpack BUILDPACK PATH POSITION [--enable|--disable]

If I mix up the order of arguments for create-buildpack its giving confusing error.

Before Fix :
 create-buildpack example-buildpack 1 some-buildpack.zip
Creating buildpack example-buildpack...
FAILED
Invalid position. strconv.ParseInt: parsing some-buildpack.zip: invalid syntax

After Fix :
 create-buildpack test 1 service.tzr
Creating buildpack test...
FAILED
Error service.tzr is being passed in as the argument for 'Position' but 'Position' requires an integer. For more syntax help, see cf create-buildpack -h.


jberkhahn:
Ok, this looks good. merged.

PR #368, opened 2015-02-27
Added api endpoint option in target command

SrinivasChilveri:
Now user could easily switch between CF instances by targeting that instance,
and could also declare org and space, as the target command already supports these arguments.
User can also use target command to set api endpoint eventhough user is not logged in but it will fail
if user try to use space or org's flag along with api's flag without login.

Story ID : #87983128


SrinivasChilveri:
Hi @goehmen ,
Do you have any comments on this story. We are ready to change if you have some comments.

goehmen:
Sure, this story is in the icebox with a label "needs-pm" because I need to
give it more thought.  I'm not convinced that adding an api endpoint in 'cf
target' is a good idea.  Especially with auth/refresh token issues.



Greg Oehmen
Cloud Foundry Product Manager
415.205.6596


On Thu, Mar 26, 2015 at 10:48 PM, SrinivasChilveri <notifications@github.com

camelpunch:
Closing due to inactivity for now. Please follow [the Tracker story](https://www.pivotaltracker.com/n/projects/892938/stories/89270100) for updates on this issue's priority.

PR #379, opened 2015-03-13
Added way to put user in org's space with 'cf target -o ORG' command if there is only one space

SrinivasChilveri:
cf target with [-o] flag will internally target org's space if there is only one space .
Story ID : #73568408
For Example :
If testOrg has only one space testspace then
Command line output with old Implementation :
$ cf target -o testOrg
API endpoint:   https://api.10.244.0.34.xip.io (API version: 2.22.0)
User:           admin
Org:            testOrg
Space:          No space targeted, use 'cf target -s SPACE'
Command line output with new Implementation:
$ cf target -o testOrg
API endpoint:   https://api.10.244.0.34.xip.io (API version: 2.22.0)
User:           admin
Org:            testOrg
Space:        testspace

### DIFF
--- a/cf/commands/target.go
+++ b/cf/commands/target.go
@@ -70,6 +70,11 @@ func (cmd Target) Run(c *cli.Context) {
 		err := cmd.setOrganization(orgName)
 		if err != nil {
 			cmd.ui.Failed(err.Error())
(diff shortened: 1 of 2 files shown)


jberkhahn:
Looks good. Merged.

PR #383, opened 2015-03-18
Add a flag to force a restage after bind/unbind service instance

SrinivasChilveri:
Solution to Story [82140190]
As of now user has to restage an app with 'cf restage SOME_APP' command after
binding or unbinding to ensure bind/unbind env variable changes take effect.
With --force-restage flag , User has option to restage an app with bind/unbind itself.

Note: Currently we have not handled this flag for if its already bind. If you want to handle that situation we are ready to change.


dkoper:
Thank you for your continuous contributions to the CLI.

I'm not clear on what problem this feature tries to solve. I assume it's not so the user needs to type less.
`cf bind-service myapp mydb -c ~/workspace/tmp/instance_config.json && cf rg myapp`
`cf bind-service myapp mydb -c ~/workspace/tmp/instance_config.json --force-restage`

I also understand that with most buildpacks a restart is sufficient, which is quicker than a restage.

It is the CLI's design philosophy for each command to do just one thing, which adds to my reluctance to accept this feature.

Cheers,
Dies Koper,
CF CLI PM

PR #384, opened 2015-03-18
Show blank last operation if the CC returns null last_operation in API r...

dsabeti:
...esponse

[fixes #90471958]


jberkhahn:
Looks good, merged.

PR #386, opened 2015-03-23
Adding the no-hostname attribute in create-app-manifest

SrinivasChilveri:
Solution: story [90554794]
Before : There was no support for --no-hostname attribute in  cf create-app-manifest 
After : If the app is mapped to the root domain then  cf create-app-manifest  will contain  no-hostname : true

### DIFF
--- a/cf/manifest/generate_manifest.go
+++ b/cf/manifest/generate_manifest.go
@@ -131,8 +131,14 @@ func (m *appManifest) Save() error {
 		}
 
 		if len(app.Routes) > 0 {
-			if _, err := fmt.Fprintf(f, "  host: %s\n", app.Routes[0].Host); err != nil {
-				return err
+			if app.Routes[0].Host == "" {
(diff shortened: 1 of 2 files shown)


jberkhahn:
So, I don't think this actually works. no-hostname is a Boolean flag - if it's in the manifest, none of the routes are going to have hostnames. With your code, if only one of the routes currently bound to the app lacks a hostname, ALL of the routes created with the generated manifest will be created with no-hostname, so you would start with something like:

Host | Domain | App
--- | --- | ---
color | foobar.com | MyApp
 | snafu.com | MyApp

You'll run `cf create-app-manifest MyApp` and get something like:
```
...
host: color
domain: foobar.com
...
domain: foobar.com
no-hostname: true
```

If you then push an app with this manifest, you'll get

Host | Domain | App
--- | --- | ---
 | foobar.com | MyApp
 | snafu.com | MyApp

So, this doesn't actually work.

PR #387, opened 2015-03-23
Fixing `spaces` displays name column heading even when requests fail

SrinivasChilveri:
Solution: story[90794162]
Before :  If apiErr happens still it will display the Spaces column heading
After  :  If apiErr happens then it will only display the apiErr message


goehmen:
Thanks for the PR but I don't think this issue needs to be addressed at this time.  I think the fix would only create uncertainty for users when they are traversing a large list of spaces.  The cmd will have the appearance of being hung in that case.

camelpunch:
Looks like an accidental re-opening.

PR #397, opened 2015-03-31
Print the "not authorized" error returned from CC when creating service key

xingzhou:
Fix a bug that only the spacedeveloper or admin can create a service key. CC will return "Not authorized error" and CLI need to report the error and print out the error message.

Besides, for other errors reported from CC side, we need also to report and print them in CLI side

this is a fix for story [#87480878]

Signed-off-by: Edward Zhang <zhuadl@cn.ibm.com>


maximilien:
@shalako this PR fixes the issue with story [#87480878] causing the rejection. Hopefully this change should allow you to accept story.

shalako:
Thank you, @maximilien. Please see my questions in the story: https://www.pivotaltracker.com/story/show/87480878

simonleung8:
Thanks, all good and merged

PR #399, opened 2015-04-03
minimum viable change to enable delegated managers

ozzyjohnson:
As stated in the comments...

    // Needed a way to allow OrgManagers to set and unset space roles
    // as the UAA call here fails for non-admins. Seems like this would
    // be better handled with a /v2/users that returns a combined collection 
    // of users for the caller's orgs, but this works.

TODO: Tests.

### DIFF
--- a/cf/api/users.go
+++ b/cf/api/users.go
@@ -31,6 +31,7 @@ var spaceRoleToPathMap = map[string]string{
 
 type UserRepository interface {
 	FindByUsername(username string) (user models.UserFields, apiErr error)
+	ListUsersInOrgWithNoUAA(orgGuid string) ([]models.UserFields, error)
 	ListUsersInOrgForRole(orgGuid string, role string) ([]models.UserFields, error)
 	ListUsersInOrgForRoleWithNoUAA(orgGuid string, role string) ([]models.UserFields, error)
 	ListUsersInSpaceForRole(spaceGuid string, role string) ([]models.UserFields, error)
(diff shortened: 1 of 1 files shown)


ozzyjohnson:
Wrong target.

cf/api/users.go
```diff
@@ -58,28 +59,76 @@ func NewCloudControllerUserRepository(config core_config.Reader, uaaGateway net.
...
 	var user models.UserFields
+
 	if apiErr != nil {
 		return user, apiErr
 	}
 
-	usernameFilter := neturl.QueryEscape(fmt.Sprintf(`userName Eq "%s"`, username))
-	path := fmt.Sprintf("%s/Users?attributes=id,userName&filter=%s", uaaEndpoint, usernameFilter)
-	users, apiErr := repo.updateOrFindUsersWithUAAPath([]models.UserFields{}, path)
+	// var myGUID string
```
maximilien: Why are we leaving what seems to be debugging code?

cf/api/users.go
```diff
@@ -58,28 +59,76 @@ func NewCloudControllerUserRepository(config core_config.Reader, uaaGateway net.
...
+			return user, errors.NewModelNotFoundError("User", username)
+		}
+
+		return users[0], apiErr
 	}
+	return user, apiErr
+}
 
-	return users[0], apiErr
+func (repo CloudControllerUserRepository) ListUsersInOrgWithNoUAA(orgGuid string) (users []models.UserFields, apiErr error) {
```
maximilien: Where is the test for this public method?

cf/api/users.go
```diff
@@ -58,28 +59,76 @@ func NewCloudControllerUserRepository(config core_config.Reader, uaaGateway net.
 
 func (repo CloudControllerUserRepository) FindByUsername(username string) (models.UserFields, error) {
```
maximilien: Where is the test for this?

Issue #402, opened 2015-04-06
Cannot use 'n' as a hostname without quoting it


wendorf:
When a manifest includes an unquoted 'n' hostname, it is for some reason not considered a string and the CLI refuses to use it.

This manifest:

```yaml
hosts:
  - n

applications:
  - name: hosts-test
```

Produces this error when run with `cf push manifest.yml`:

```
	Aww shucks.

	Something completely unexpected happened. This is a bug in cf.
	Please file this bug : https://github.com/cloudfoundry/cli/issues
	Tell us that you ran this command:

		cf push -f manifest.yml

	using this version of the CLI:

		6.10.0-b78bf10

	and that this error occurred:

		runtime error: invalid memory address or nil pointer dereference

	and this stack trace:

		goroutine 1 [running]:
	main.generateBacktrace(0xc21000a000, 0x4)
		/Users/pivotal/go-agent/pipelines/Mac-OSX-Unit-Tests/src/github.com/cloudfoundry/cli/main/main.go:174 +0xa9
	main.handlePanics(0x10265d8, 0xc2100684e0)
		/Users/pivotal/go-agent/pipelines/Mac-OSX-Unit-Tests/src/github.com/cloudfoundry/cli/main/main.go:158 +0x123
	runtime.panic(0x543c80, 0xdf5119)
		/usr/local/go/src/pkg/runtime/panic.c:248 +0x106
	github.com/cloudfoundry/cli/cf/manifest.mapToAppParams(0xc210177c40, 0x9, 0x1036440, 0xc21021b970, 0x0, ...)
		/Users/pivotal/go-agent/pipelines/Mac-OSX-Unit-Tests/src/github.com/cloudfoundry/cli/tmp/cli_gopath/src/github.com/cloudfoundry/cli/cf/manifest/manifest.go:140 +0x76b
	github.com/cloudfoundry/cli/cf/manifest.Manifest.Applications(0xc210177c40, 0x18, 0x1036440, 0xc21021b8e8, 0x0, ...)
		/Users/pivotal/go-agent/pipelines/Mac-OSX-Unit-Tests/src/github.com/cloudfoundry/cli/tmp/cli_gopath/src/github.com/cloudfoundry/cli/cf/manifest/manifest.go:43 +0x2ef
	github.com/cloudfoundry/cli/cf/commands/application.
```

Using the single "host" attribute, the error is a little nicer:

```yaml
host: n

applications:
  - name: hosts-test
```

```
FAILED
[...]

PR #406, opened 2015-04-09
Fix the hostname panic issue story [91881010]

SrinivasChilveri:
The sliceOrEmptyVal function can return nil, but in mapToAppParams function
we are invoking the function sliceOrEmptyVal as shown below.
So on returning nil its panicking because accessing value from nil pointer by (* nil) which is wrong.
hostsArr := *sliceOrEmptyVal(yamlMap, hosts, &errs)

After fixing the output as

@cf5:~/go/src/github.com/cloudfoundry/cli$ ./out/cf push -f ~/cf-downloads/simple-go-web-app/manifest.yml
FAILED
Error reading manifest file:
Expected hosts to be a list of strings


simonleung8:
Thanks for the PR.

However, we feel a more proper fix for the panic is simply not to return nil, instead, we could just return an empty slice.
We have introduced a fix while investigating issue #402. 
https://github.com/cloudfoundry/cli/commit/5580a6016ec4059c0e5368551da43972d1255138

Again, thanks for the effort.

PR #419, opened 2015-04-20
add show service key detail

zhang-hua:
[#87061876]

Signed-off-by: zhang-hua <zhuadl@cn.ibm.com>

### DIFF
--- a/cf/api/service_keys_test.go
+++ b/cf/api/service_keys_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cloudfoundry/cli/cf/configuration/core_config"
 	"github.com/cloudfoundry/cli/cf/errors"
+	"github.com/cloudfoundry/cli/cf/models"
 	"github.com/cloudfoundry/cli/cf/net"
 
 	. "github.com/cloudfoundry/cli/cf/api"
@@ -42,7 +43,7 @@ var _ = Describe("Service Keys Repo", func() {
 		repo = NewCloudControllerServiceKeyRepository(configRepo, gateway)
 	})
 
-	Describe("creating a service key", func() {
+	Describe("CreateServiceKey", func() {
 		It("makes the right request", func() {
 			setupTestServer(testapi.NewCloudControllerTestRequest(testnet.TestRequest{
 				Method:   "POST",
@@ -88,7 +89,7 @@ var _ = Describe("Service Keys Repo", func() {
 		})
 	})
 
-	Describe("listing service keys", func() {
+	Describe("ListServiceKeys", func() {
 		It("returns empty result when no service key is found", func() {
 			setupTestServer(testapi.NewCloudControllerTestRequest(testnet.TestRequest{
 				Method:   "GET",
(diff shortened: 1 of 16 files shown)


zhang-hua:
@simonleung8 please review this PR for getting service key detail. Thanks!

cc @maximilien @xingzhou

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,125 @@
...
+			Expect(callGetServiceKey([]string{"non-exist-service-instance"})).To(BeFalse())
+		})
+
+		It("fails when space is not targetted", func() {
+			requirementsFactory = &testreq.FakeReqFactory{LoginSuccess: true, TargetedSpaceSuccess: false}
+			Expect(callGetServiceKey([]string{"fake-service-instance", "fake-service-key-name"})).To(BeFalse())
+		})
+	})
+
+	Describe("requiremnts are satisfied", func() {
```
maximilien: small typo: requirements

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,125 @@
...
+						"password": "fake-password",
+						"host":     "fake-host",
+						"port":     "3306",
+						"database": "fake-db-name",
+						"uri":      "mysql://fake-user:fake-password@fake-host:3306/fake-db-name",
+					},
+				}
+			})
+
+			It("get service credential", func() {
```
maximilien: Minor: should read `It("gets service ...")` same also for rest. Reember the tests description should almost read like English proses... as much as you can. I know this is not super important but since we are there and I saw it, I am pointting them out...

Finally, I realize that English is not first language so some of these typos/grammar errors are normal. Cheers.

cf/commands/servicekey/service_key_test.go
zhang-hua: will fix it. thanks!

cf/commands/servicekey/service_key_test.go
zhang-hua: will fix it. thanks!

zhang-hua:
@simonleung8 have fixes the minor issues commented by Max. please help review this PR for getting service key detail. Thanks!

cc @maximilien @xingzhou

cf/api/service_keys_test.go
```diff
@@ -142,6 +142,33 @@ var _ = Describe("Service Keys Repo", func() {
...
+
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("username", "fake-username"))
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("password", "fake-password"))
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("host", "fake-host"))
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("port", float64(3306)))
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("database", "fake-db-name"))
+			Expect(serviceKey.Credentials).To(HaveKeyWithValue("uri", "mysql://fake-user:fake-password@fake-host:3306/fake-db-name"))
+		})
+	})
+
```
maximilien: What about the case where the right call is made but the key does not exist? I'd expect this is a valid `get service key` that we should test and verify that errors are displayed, something like `service key named ... is not valid` or similar.

Unless, of course, you are testing this case somewhere else.

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,125 @@
...
+		})
+
+		It("requires two arguments and one option to run", func() {
+			Expect(callGetServiceKey([]string{})).To(BeFalse())
+			Expect(callGetServiceKey([]string{"fake-arg-one"})).To(BeFalse())
+			Expect(callGetServiceKey([]string{"fake-arg-one", "fake-arg-two", "fake-arg-three"})).To(BeFalse())
+		})
+
+		It("fails when service instance is not found", func() {
+			requirementsFactory = &testreq.FakeReqFactory{LoginSuccess: true, ServiceInstanceNotFound: true}
```
maximilien: Shouldn't you still check that this message:

```
"No service key {{.ServiceKeyName}} found for service instance {{.ServiceInstanceName}}"
```

is printed to user?

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,125 @@
...
+			It("gets service guid", func() {
+				callGetServiceKey([]string{"--guid", "fake-service-instance", "fake-service-key"})
+				Expect(ui.Outputs).To(ContainSubstrings(
+					[]string{"Getting key", "fake-service-key", "for service instance", "fake-service-instance", "as", "my-user"},
+					[]string{"fake-service-key-guid"},
+				))
+			})
+		})
+
+		It("does not show service key detail when service key not exist", func() {
```
maximilien: This one should be in different Context since the current context is: `gets service key successfully`, or remove that context.

zhang-hua: The last test case is not in a context. But the first 3 test cases share the same context which requires setup fake service key repo. we should not remove the context.

cf/api/service_keys_test.go
zhang-hua: good point! I'll add one more test to cover this case for api side.  thanks!

zhang-hua:
Add one more test case for get the detail of non-exist service key suggested by Max. Thanks!

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,127 @@
...
+		return testcmd.RunCommand(cmd, args, requirementsFactory)
+	}
+
+	Describe("requirements", func() {
+		It("fails when not logged in", func() {
+			requirementsFactory = &testreq.FakeReqFactory{LoginSuccess: false}
+			Expect(callGetServiceKey([]string{"fake-service-key-name"})).To(BeFalse())
+		})
+
+		It("requires two arguments and one option to run", func() {
```
simonleung8: This should read "requires two arguments to run", the `--guid` flag is an option and is not required to run.

And to actually validate this test, add a new `Expect` to show that providing 2 arguments will indeed pass requirement "BeTrue()"

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,127 @@
...
+					[]string{"uri", "mysql://fake-user:fake-password@fake-host:3306/fake-db-name"},
+					[]string{"OK"},
+				))
+				Expect(serviceKeyRepo.GetServiceKeyMethod.InstanceId).To(Equal("fake-service-instance-guid"))
+			})
+
+			It("gets service guid", func() {
+				callGetServiceKey([]string{"--guid", "fake-service-instance", "fake-service-key"})
+				Expect(ui.Outputs).To(ContainSubstrings(
+					[]string{"Getting key", "fake-service-key", "for service instance", "fake-service-instance", "as", "my-user"},
```
simonleung8: The convention of CLI is to display only the "guid" whenever `--guid` is provided. A lot of CLI user leverage the guid feature for inline scripting, for example 
```
cf curl /v2/app/`cf app app_name --guid`
```
You can checkout other cf command with `--guid` flag as reference

cf/commands/servicekey/service_key_test.go
```diff
@@ -0,0 +1,127 @@
...
+					[]string{"host", "fake-host"},
+					[]string{"port", "3306"},
+					[]string{"database", "fake-db-name"},
+					[]string{"uri", "mysql://fake-user:fake-password@fake-host:3306/fake-db-name"},
+					[]string{"OK"},
+				))
+				Expect(serviceKeyRepo.GetServiceKeyMethod.InstanceId).To(Equal("fake-service-instance-guid"))
+			})
+
+			It("gets service guid", func() {
```
simonleung8: to better describe what the test does, this could read something like "gets service guid when '--guid' flag is provided"

cf/api/service_keys_test.go
```diff
@@ -142,6 +143,46 @@ var _ = Describe("Service Keys Repo", func() {
 		})
 	})
 
+	Describe("getting service key", func() {
```
simonleung8: Sorry for being a pedant, but this might be better to be Describe("GetServiceKey()"), since the describe block is essentially for GetServiceKey()

simonleung8:
Left some comments, nothing really major. I think the rest of it is good.

cf/commands/servicekey/service_key_test.go
zhang-hua: will fix it. thanks!

cf/commands/servicekey/service_key_test.go
zhang-hua: will fix this. Thanks!

cf/api/service_keys_test.go
zhang-hua: will fix it. thanks!

cf/commands/servicekey/service_key_test.go
zhang-hua: it make sense to do that. will fix it. thanks!

zhang-hua:
@simonleung8 thanks for your comments! I have updated this PR. please take a look.

simonleung8:
Thanks for addressing the issues so quickly. Merged.

PR #432, opened 2015-04-28
make accepts_incomplete configurable (i.e. for build jobs)

adrai:
I've never written go code but it would be really nice to choose to "accepts_incomplete" calls.
For example in build jobs this would be really nice.

### DIFF
--- a/cf/i18n/resources/en_US.all.json
+++ b/cf/i18n/resources/en_US.all.json
@@ -4505,6 +4505,26 @@
       "modified": false
    },
    {
+      "id": "Wait for update confirmation",
+      "translation": "Wait for update confirmation",
+      "modified": false
+   },
+   {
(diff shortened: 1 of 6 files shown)


cf/i18n/resources/en_US.all.json
```diff
@@ -4505,6 +4505,26 @@
       "modified": false
    },
    {
+      "id": "Wait for update confirmation",
```
maximilien: You need to add the same new strings to the other language resource files

maximilien:
This PR seems to have no new tests nor tests changes. I'd expect lots of changes for each. Plus you are failing Travis.

I'd suggest you close this and work some more on this before attempting to submit. It's also doubtful to be accepted since this feature needs to be discuss with @goehmen and @shalako.

Anyhow, my .02

PR #433, opened 2015-04-29
Restart command: load application(s) from manifest if no app name provided

uzzz:
Hi

I found it annoying to type app name every time when I run CF commands from an app dir with manifest.yml. I don't have to specify app name when I run "cf push", but I have to type it when I run cf restart/start/stop/restage/etc. This patch is not a merge-ready, at least because of missing translations in language files (which leads to failing ./bin/test). First of all I want to demonstrate an idea on restart command and get any feedback.

### DIFF
--- a/cf/commands/application/restart_test.go
+++ b/cf/commands/application/restart_test.go
@@ -1,9 +1,13 @@
 package application_test
 
 import (
+	testApplication "github.com/cloudfoundry/cli/cf/api/applications/fakes"
+	"github.com/cloudfoundry/cli/cf/manifest"
 	"github.com/cloudfoundry/cli/cf/models"
+	"github.com/cloudfoundry/cli/generic"
 	testcmd "github.com/cloudfoundry/cli/testhelpers/commands"
 	testconfig "github.com/cloudfoundry/cli/testhelpers/configuration"
+	testmanifest "github.com/cloudfoundry/cli/testhelpers/manifest"
 	testreq "github.com/cloudfoundry/cli/testhelpers/requirements"
 	testterm "github.com/cloudfoundry/cli/testhelpers/terminal"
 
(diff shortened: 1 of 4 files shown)


goehmen:
@uzzz My initial feedback is that there is a stark difference between push and stop/start/restart/restage and other commands like scale, bind (to a service).  When you are pushing an app, a failed or unintended result is greatly innocuous.  However, when you stop/start/etc, you are impact a running app and very possibly an app that is in production.  In that case, being specific and being deterministic are very good safeguards against an unintended action.  

If you want to get more feedback, you could post the same question out on vcap-dev and I'm sure the community will weigh in.

cf/i18n/resources/en_US.all.json
```diff
@@ -910,13 +910,28 @@
...
+   },
+   {
       "id": "CF_NAME restage APP_NAME",
       "translation": "CF_NAME restage APP_NAME",
       "modified": false
    },
    {
-      "id": "CF_NAME restart APP_NAME",
-      "translation": "CF_NAME restart APP_NAME",
+      "id": "   CF_NAME restart APP_NAME\n",
```
maximilien: Have those strings changed? Why are we not seeing the same changes for the other language files?

I am assuming they were not changed but somehow the `en_US` file was touched? Otherwise, the other files need to change as well.

cf/commands/application/restart_test.go
```diff
@@ -63,27 +63,62 @@ var _ = Describe("restart command", func() {
...
+				Expect(application).To(Equal(app))
+				Expect(orgName).To(Equal(config.OrganizationFields().Name))
+				Expect(spaceName).To(Equal(config.SpaceFields().Name))
+
+				Expect(appRepo.ReadArgs.Name).To(Equal("my-app"))
+			})
+		})
 
-			Expect(requirementsFactory.ApplicationName).To(Equal("my-app"))
+		Context("app name is not provided", func() {
```
maximilien: Should we have a test for app name not provided and no manifest or lack of app name in manifest?

PR #434, opened 2015-04-29
Add (failing) test case for deep merge where two arrays of hashes are merged

dlapiduz:
I am running into an issue with manifest inheritance where if I have the app specified in the main manifest and try to override it from the child one, two apps are created.

It is reproduced with the following test but I am not sure if this is an issue that we should solve at the merging level.

Some manifest examples:

```
---
# main.yml
memory: 512M
services:
- mydb
applications:
- name: myapp
  path: .
```

```
---
# dev.yml
inherit: main.yml
applications:
- name: myapp
  host: myapp-dev
```

And the merged result is something like:
```
memory: 512M
services:
- mydb
applications:
- name: myapp
  path: .
- name: myapp
  host: myapp-dev
```

When I was expecting:
```
memory: 512M
services:
- mydb
applications:
- name: myapp
  path: .
  host: myapp-dev
```

### DIFF
--- a/generic/map_test.go
+++ b/generic/map_test.go
@@ -19,6 +19,12 @@ func init() {
 				"nest2": []interface{}{
 					"nest2Val1",
 				},
+				"nest3": []interface{}{
+					map[interface{}]interface{}{
+						"nestKey1": "nest3Val1",
+						"nestKey2": "nest3Val2",
+					},
+				},
 			})
 
 			map2 := NewMap(map[interface{}]interface{}{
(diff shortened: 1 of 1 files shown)


jberkhahn:
Hey,

This is working as intended. App host inheritances isn't going to override the definition in the original manifest because having multiple hosts/domains is a valid use cause, so it cats them rather than overriding. If you wanted to override it, try using a command line argument when you actually push the app.

Also, in the future if you encounter a bug like this, please submit a github issue rather than a failing test with no solution.

dlapiduz:
Sorry @jberkhahn I didn't think this would be an issue. Basically I made the test to make sure that the use case that we were experiencing got through and since it was code I just sent it in a PR. Should have done a gist in an issue instead and have a conversation there instead of having it in a PR.

Moved to #479 so we can have the conversation there.

PR #441, opened 2015-05-06
Arbitrary params for create-service

dsabeti:
Also includes a fix to a different bug where `async: true` was mistakenly being sent in the post body.

### DIFF
--- a/cf/api/services_test.go
+++ b/cf/api/services_test.go
@@ -46,7 +46,9 @@ var _ = Describe("Services Repo", func() {
 	})
 
 	AfterEach(func() {
-		testServer.Close()
+		if testServer != nil {
+			testServer.Close()
+		}
 	})
 
 	Describe("GetAllServiceOfferings", func() {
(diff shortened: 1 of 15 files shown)


cf/i18n/resources/fr_FR.all.json
```diff
@@ -5323,5 +5323,10 @@
       "id": "service key",
       "translation": "clef de service",
       "modified": false
+   },
+   {
+      "id": "Valid JSON object containing service-specific configuration parameters, provided either in-line or in a file. For a list of supported configuration parameters, see documentation for the particular service offering.",
+      "translation": "Valid JSON object containing service-specific configuration parameters, provided either in-line or in a file. For a list of supported configuration parameters, see documentation for the particular service offering.",
```

```
"Objet JSON valide contenant les paramètres de configuration spécifiques au service, fourni soit en ligne ou dans un fichier. Pour une liste des paramètres de configuration valide, voir la documentation de l'offre de service particulier."
```

cf/commands/service/create_service.go
```diff
@@ -60,6 +66,13 @@ func (cmd CreateService) Run(c *cli.Context) {
 	serviceName := c.Args()[0]
 	planName := c.Args()[1]
 	serviceInstanceName := c.Args()[2]
+	params := c.String("c")
+
+	paramsMap := make(map[string]interface{})
+	err := json.Unmarshal([]byte(params), &paramsMap)
+	if err != nil && params != "" {
+		cmd.ui.Failed("Invalid JSON provided in parameters argument")
```
maximilien: This string (if it might end up shown to end user) should be in `T("...")`

Here is the French translation, if you do add it:

```
"JSON non-valide fourni dans l'argumentation des paramètres"
```

cf/api/services_test.go
```diff
@@ -193,26 +195,54 @@ var _ = Describe("Services Repo", func() {
...
+				paramsMap := make(map[string]interface{})
+				paramsMap["data"] = "hello"
+
+				err := repo.CreateServiceInstance("instance-name", "plan-guid", paramsMap)
+				Expect(testHandler).To(HaveAllRequestsCalled())
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+
+		Context("and there is a failure during serialization", func() {
```
maximilien: What if the parameters string is almost valid JSON, so something like: `{"data": "hello}`, so not a type serialization error but rather unmarshalling error? Is this also covered by this test?

dsabeti: Notice that the value here is `paramsMap`, a map instead of a string. The user's input actually gets unmarshalled earlier, so if the json is invalid, it will raise an error there. By the time we get to this part of the code, we know for a fact that the object is not "almost" valid json. Look at https://github.com/cloudfoundry/cli/pull/441/files#diff-e39828f6ca182457b7f7a8842b138547R110

maximilien: Cool. The `make(chan bool)` thingie was kind of odd to use for test but fair enough since any marshaling error will be caught. The test might be clearer but no biggie. Cheers.

PR #452, opened 2015-05-19
Arbitrary params

rmasand:
As discussed, we've included all of our arbitrary params work in a single PR. This feature allows the user to provide arbitrary JSON params in either raw JSON or file format. We've added this for create-service, bind-service, update-service, and create-service-key.

This covers:
- https://www.pivotaltracker.com/story/show/89843658
- https://www.pivotaltracker.com/story/show/88670540
- https://www.pivotaltracker.com/story/show/88670566
- https://www.pivotaltracker.com/story/show/89843656 
- https://www.pivotaltracker.com/story/show/89843654
- https://www.pivotaltracker.com/story/show/88670578
- https://www.pivotaltracker.com/story/show/90163330
- https://www.pivotaltracker.com/story/show/90163332

Let us know if you have any feedback!


json/json_parser.go
```diff
@@ -46,6 +46,21 @@ func ParseJsonHash(path string) (map[string]interface{}, error) {
 	return stringMap, nil
 }
 
+func ParseJsonFromFileOrString(fileOrJson string) (map[string]interface{}, error) {
+	var jsonMap map[string]interface{}
+	var err error
+
+	jsonMap, err = ParseJsonHash(fileOrJson)
+	if err != nil {
```
simonleung8: The error message is hidden from the user in case of the file contains invalid string, because it is over-written by the error message at line 56. If it was actually a file containing invalid json, the error message explaining that will never get displayed to the user. They'll see an error message about how their filename was invalid json instead. There's a go command to validate a filepath that you can use to check if it's a string or filepath, and switch from there.

cf/commands/service/bind_service.go
```diff
@@ -65,6 +88,12 @@ func (cmd *BindService) GetRequirements(requirementsFactory requirements.Factory
 func (cmd *BindService) Run(c *cli.Context) {
 	app := cmd.appReq.GetApplication()
 	serviceInstance := cmd.serviceInstanceReq.GetServiceInstance()
+	params := c.String("c")
+
+	paramsMap, err := json.ParseJsonFromFileOrString(params)
+	if err != nil {
+		cmd.ui.Failed("Invalid JSON provided in -c argument")
```
simonleung8: In general, we try to bubble actual errors up to the user. In this case, totally overwriting the error is bad because it doesn't tell the user what went wrong: did the file contain invalid json? did the file not exist? who knows? Just bubble up errors and let the user figure stuff out.

simonleung8:
@rmasand: We made some comments, nothing really major, mostly just stuff that could improve the UX.

on the side note, inputing json strings requires different format on Windows/Linux/OSX, it might be helpful to user to document the different formats. We have an example in command `create-user-provided-service`'s help text.

jberkhahn:
This looks good, merged.

dave-chen:
Hi,
I need to use arbitrary params features.
Can someone tell me which CF version is compatible to CLI v.6.12.0?

Thanks,
Dave

rmasand:
Hi @dave-chen,

Arbitrary params are in CC API version 2.26.0, which is in v213. The feature may be present in some form as early as 211, but I'd recommend starting with 213 to be safe.

dave-chen:
Thanks Raina!

PR #453, opened 2015-05-20
Last operation timestamps

tanglisha:
- Add started and updated timestamps to service instance operations
- Updated the CLI to not return a Started date if the service/operation does not have a CreatedAt in it's JSON.

### DIFF
--- a/cf/api/resources/service_instances_test.go
+++ b/cf/api/resources/service_instances_test.go
@@ -36,7 +36,9 @@ var _ = Describe("ServiceInstanceResource", func() {
         "last_operation": {
           "type": "create",
           "state": "in progress",
-          "description": "fake state description"
+          "description": "fake state description",
+          "created_at": "fake created at",
+          "updated_at": "fake updated at"
         },
         "service_plan": {
           "metadata": {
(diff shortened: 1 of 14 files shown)


cf/i18n/resources/fr_FR.all.json
```diff
@@ -4050,6 +4050,11 @@
       "modified": false
    },
    {
+      "id": "Started: {{.Started}}",
+      "translation": "Started: {{.Started}}",
```
```
"Commencez: {{.Started}}"
```

cf/i18n/resources/fr_FR.all.json
```diff
@@ -4270,6 +4275,11 @@
       "modified": false
    },
    {
+      "id": "Updated: {{.Updated}}",
+      "translation": "Updated: {{.Updated}}",
```
maximilien:
```
"Changez: {{.Updated}}"
```

cf/api/resources/service_instances_test.go
```diff
@@ -125,7 +127,127 @@ var _ = Describe("ServiceInstanceResource", func() {
...
+				Expect(fields.LastOperation.UpdatedAt).To(Equal("fake updated at"))
 			})
+
+			Context("When created_at is null", func() {
+				It("unmarshalls the service instance resource model", func() {
+					var resourceWithNullCreatedAt ServiceInstanceResource
+
+					err := json.Unmarshal([]byte(`
+						{
+							"metadata": {
```
maximilien: This seems very noisy when inline. Did you consider defining once and modifying string (using `strings` package for instance) later for each case. Especially since you also need to do the same for `updated_at` ?

Gerg:
Added test fixture changes and rebased.

PR #455, opened 2015-05-21
Updated cf service-access and cf service-brokers so that they only pass through the 403 error, rather than giving specific lookup information.

tanglisha:
[#91452714]

### DIFF
--- a/cf/commands/serviceaccess/service_access.go
+++ b/cf/commands/serviceaccess/service_access.go
@@ -105,7 +105,7 @@ func (cmd *ServiceAccess) Run(c *cli.Context) {
 
 	brokers, err := cmd.actor.FilterBrokers(brokerName, serviceName, orgName)
 	if err != nil {
-		cmd.ui.Failed(T("Failed fetching service brokers.\n{{.Error}}", map[string]interface{}{"Error": err}))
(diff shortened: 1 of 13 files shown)


cf/commands/serviceaccess/service_access.go
```diff
@@ -105,7 +105,7 @@ func (cmd *ServiceAccess) Run(c *cli.Context) {
 
 	brokers, err := cmd.actor.FilterBrokers(brokerName, serviceName, orgName)
 	if err != nil {
-		cmd.ui.Failed(T("Failed fetching service brokers.\n{{.Error}}", map[string]interface{}{"Error": err}))
+		cmd.ui.Failed(T("{{.Error}}", map[string]interface{}{"Error": err}))
```
maximilien: Since you are passing the error string as is, I fail to see why you'd i18n the string. So instead, until you have additional strings for format the error, the line above should be:

```
cmd.ui.Failed(err.Error())
```

I am assuming that this change is the correct thing to do. A different discussion.

Gerg:
Updated error printing and rebased.

simonleung8:
We added a commit on top to update the error printing in service_brokers.go
other than that, everything looks good and merged.

PR #456, opened 2015-05-23
Fixed misleading error when quotes are used to wrap json for cf curl -d 

diatmpravin:
notice the error returned when single ticks are not used to wrap the argument for -d.

When I was debugging code, I realized that flag 'X' was missing in flags list. 

ALL Flags-----> [i v H d output]
 
Story details: https://www.pivotaltracker.com/n/projects/892938/stories/95206110

### DIFF
--- a/cf/command_factory/factory.go
+++ b/cf/command_factory/factory.go
@@ -378,6 +378,8 @@ func (f concreteFactory) GetCommandFlags(command string) []string {
 			flags = append(flags, t.Name)
 		case flag_helpers.StringFlagWithNoDefault:
 			flags = append(flags, t.Name)
+		case cli.StringFlag:
+			flags = append(flags, t.Name)	
 		case cli.BoolFlag:
 			flags = append(flags, t.Name)
 		}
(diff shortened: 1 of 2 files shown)


maximilien:
@diatmpravin,

I am sure I speak for everyone in CF in appreciating your contributions. However, as you know, we test-drive code here in CF and related projects. 

This means that when you make submissions (as this one and others) YOU MUST also submit tests. If there was no tests to start with (unlikely) but even in those cases, you need to backfill and add new tests.

Thanks,

dr.max

diatmpravin:
@maximilien I will write test case and resubmit. Thanks for your comment.

diatmpravin:
@maximilien Added test case!

camelpunch:
Hi @diatmpravin. Sorry it's been a while since you added your test case. It looks like CLI's architecture has moved around since this PR was made. I'm not entirely sure, but it looks like the flags you mentioned are now at least in the help text:

```
cf curl -d
FAILED
Incorrect Usage

No value provided for flag: -d

NAME:
   curl - Executes a raw request, content-type set to application/json by default

USAGE:
   cf curl PATH [-iv] [-X METHOD] [-H HEADER] [-d DATA] [--output FILE]
   For API documentation, please visit http://apidocs.cloudfoundry.org

OPTIONS:
   -H            Custom headers to include in the request, flag can be specified multiple times
   -d            HTTP data to include in the request body
   -output       Write curl body to FILE instead of stdout
   -X            HTTP method (GET,POST,PUT,DELETE,etc)
   -i            Include response headers in the output
   -v            Enable CF_TRACE output for all requests and responses
```

Please re-open and rebase, or make your change again if you think it's still needed. Thanks!

PR #465, opened 2015-06-02
Expose api errors for service broker commands

rmasand:
The API was returning a 403 for unauthorized service broker operations, but the CLI was making it seem like a 404. We're now going to expose all CC errors that occur. The previous behavior is retained only for the case where there is no CC error, but the list of service brokers is empty.

https://www.pivotaltracker.com/story/show/94892746

Thanks!

Raina and @Gerg 
CF Services API

### DIFF
--- a/cf/api/service_brokers.go
+++ b/cf/api/service_brokers.go
@@ -56,7 +56,7 @@ func (repo CloudControllerServiceBrokerRepository) FindByName(name string) (serv
 			return false
 		})
 
-	if !foundBroker {
+	if !foundBroker && (apiErr == nil) {
 		apiErr = errors.NewModelNotFoundError("Service Broker", name)
 	}
 
--- a/cf/api/service_brokers_test.go
+++ b/cf/api/service_brokers_test.go
@@ -135,6 +135,27 @@ var _ = Describe("Service Brokers Repo", func() {
 			Expect(apiErr).To(HaveOccurred())
 			Expect(apiErr.Error()).To(Equal("Service Broker my-broker not found"))
 		})
+
+		It("returns an error when listing service brokers returns an api error", func() {
+			req := testapi.NewCloudControllerTestRequest(testnet.TestRequest{
+				Method: "GET",
+				Path:   "/v2/service_brokers?q=name%3Amy-broker",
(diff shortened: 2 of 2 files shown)


cf/api/service_brokers_test.go
```diff
@@ -135,6 +135,27 @@ var _ = Describe("Service Brokers Repo", func() {
 			Expect(apiErr).To(HaveOccurred())
 			Expect(apiErr.Error()).To(Equal("Service Broker my-broker not found"))
 		})
+
+		It("returns an error when listing service brokers returns an api error", func() {
```
maximilien: This description reads odd... `"returns... returns..."`

cf/api/service_brokers.go
```diff
@@ -56,7 +56,7 @@ func (repo CloudControllerServiceBrokerRepository) FindByName(name string) (serv
 			return false
 		})
 
-	if !foundBroker {
+	if !foundBroker && (apiErr == nil) {
```
maximilien: Also, since you have now added another variable to the conditional, I expected at least two more tests or really four totals: `foundBroker and apiErr nil`, "foundBroker and apiErr not nil`, `!foundBroker and apiErr nil`, and `!foundBroker and apiErr not nil`.

Are you compacting the tests or am I missing something?

rmasand: I think the test coverage is sufficient; there's no case where we'd have a list of service brokers to search as well as an apiErr from CC. If we get a CC error, we return from ListPaginatedResources before ever searching for a potential service broker. foundBroker will always remain false if there's a CC error.

This block covers a weird case: the CC may return 200 with an empty list, so we make our own error in this if-block. The additional check is a bug fix; the previous behavior was ignoring existing CC errors and would overwrite them with the ModelNotFoundError.

maximilien: Hmm, OK. I guess I'd then simplify the conditional into two. However, we are splitting hair here so if others are OK then I am happy to be cool with it. Cheers.

jberkhahn:
Hmm weird, looks like gitbot didn't make us a story for this one. So didn't see this until now.
Link to the new story: https://www.pivotaltracker.com/story/show/96804244

PR #467, opened 2015-06-04
fix the error message when service key does not exist

xingzhou:
[#87062548]
https://www.pivotaltracker.com/story/show/87062548

### DIFF
--- a/cf/commands/servicekey/delete_service_key.go
+++ b/cf/commands/servicekey/delete_service_key.go
@@ -86,7 +86,7 @@ func (cmd DeleteServiceKey) Run(c *cli.Context) {
 	}
 
 	serviceKey, err := cmd.serviceKeyRepo.GetServiceKey(serviceInstance.Guid, serviceKeyName)
-	if err != nil {
+	if err != nil || serviceKey.Fields.Guid == "" {
 		cmd.ui.Ok()
 
 		cmd.ui.Say(T("Service key {{.ServiceKeyName}} does not exist for service instance {{.ServiceInstanceName}}.",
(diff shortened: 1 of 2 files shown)


zhang-hua:
@simonleung8 @jberkhahn : this PR is to fixed the bug of error message of deleting service key for story [#87062548](https://www.pivotaltracker.com/story/show/87062548). Please help us to review and merge it. thanks!

@shalako @maximilien

cf/commands/servicekey/delete_service_key.go
```diff
@@ -86,7 +86,7 @@ func (cmd DeleteServiceKey) Run(c *cli.Context) {
 	}
 
 	serviceKey, err := cmd.serviceKeyRepo.GetServiceKey(serviceInstance.Guid, serviceKeyName)
-	if err != nil {
+	if err != nil || serviceKey.Fields.Guid == "" {
```
maximilien: Shouldn't we check first specifically for `serviceKey.Fields.Guid == ""` since that should always produce an error? Since if the key exists the GUID cannot be `""`.  Correct?

zhang-hua: the Guid could be empty if the service key doesn't  exist but empty result is returned without an error by director. The error need also be checked.

zhang-hua:
@simonleung8 @jberkhahn : this PR is to fixed the bug of error message of deleting service key for story #87062548. Please help us to review and merge it. thanks!

jberkhahn:
Hey there,

We can't automatically merge this, I would guess because of a merge conflict in the translation files. Would you mind rebasing on top of master?

zhang-hua:
@jberkhahn this bug has been fixed by Greg in the patch [#06b816](https://github.com/cloudfoundry/cli/commit/06b81638c06bd2789ef179ea45c4f13f5576a7b3). So we don't need this PR. So we will close it.

PR #470, opened 2015-06-04
Fix flaky test for go 1.4 where map iteration order is randomized.

Gerg:


simonleung8:
Thanks, looks good, merged

PR #474, opened 2015-06-08
Update user-provided service tags

rmasand:
This PR includes changes to add the "-t" flag for update-service. We've also fixed that command so users can update instance tags and/or arbitrary params without supplying an updated plan. Some cleanup of translated strings is included as well.

https://www.pivotaltracker.com/story/show/72117050


cf/api/services_test.go
```diff
@@ -203,7 +203,7 @@ var _ = Describe("Services Repo", func() {
 				Response: testnet.TestResponse{Status: http.StatusCreated},
 			}))
 
-			err := repo.CreateServiceInstance("instance-name", "plan-guid", nil)
+			err := repo.CreateServiceInstance("instance-name", "plan-guid", nil, nil)
```
maximilien: I kind of hate seeing these `nil`s all over the place when we could use empty string array...  But eh, that's just me.

Though hopefully we are testing for case when tags is empty array vs nil...

cf/commands/service/bind_service.go
```diff
@@ -36,37 +38,36 @@ func NewBindService(ui terminal.UI, config core_config.Reader, serviceBindingRep
 }
 
 func (cmd *BindService) Metadata() command_metadata.CommandMetadata {
-	return command_metadata.CommandMetadata{
-		Name:        "bind-service",
-		ShortName:   "bs",
-		Description: T("Bind a service instance to an app"),
-		Usage: T(`CF_NAME bind-service APP_NAME SERVICE_INSTANCE [-c PARAMETERS_AS_JSON]
+	baseUsage := T("CF_NAME bind-service APP_NAME SERVICE_INSTANCE [-c PARAMETERS_AS_JSON]")
```
maximilien: Yeah, thanks for the nicely formatted string!

cf/i18n/resources/fr_FR.all.json
```diff
@@ -4545,6 +4565,11 @@
       "modified": false
    },
    {
+      "id": "User provided tags",
+      "translation": "User provided tags",
```

```
"Étiquettes fournis par l'utilisateur"
```

jberkhahn:
Thanks a bunch for fixing this. Merged.

PR #478, opened 2015-06-11
Allow update service instances with empty tags

Gerg:
Tracker story: https://www.pivotaltracker.com/story/show/96329216


simonleung8:
Thanks, everything looks good, merged!

PR #480, opened 2015-06-15
Missing service key coloring message from dsk  now matches the coloring from ds

tanglisha:
- ui type is now `Warn` instead of `Say`
- Keyword highlight is now switched off

Tracker link: https://www.pivotaltracker.com/n/projects/1211680/stories/94220156


simonleung8:
looks good, merged.

PR #483, opened 2015-06-18
cf service-access performance

Gerg:
To improve performance of the `service access` command we changed the way that we:
1. Get service offerings for brokers
2. Get service plans for service offerings
3. Get organization names for service plan visibilities

Each change has its own commit with further details in the commit message. Let us know if you have any questions about our changes.

Tracker Story: https://www.pivotaltracker.com/story/show/96912380

### DIFF
--- a/cf/api/service_plan.go
+++ b/cf/api/service_plan.go
@@ -44,10 +44,11 @@ func (repo CloudControllerServicePlanRepository) Update(servicePlan models.Servi
 	return repo.gateway.UpdateResource(repo.config.ApiEndpoint(), url, strings.NewReader(body))
 }
 
-func (repo CloudControllerServicePlanRepository) ListPlansFromManyServices(serviceGuids []string) (plans []models.ServicePlanFields, err error) {
+func (repo CloudControllerServicePlanRepository) ListPlansFromManyServices(serviceGuids []string) ([]models.ServicePlanFields, error) {
 	serviceGuidsString := strings.Join(serviceGuids, ",")
+	plans := []models.ServicePlanFields{}
 
-	err = repo.gateway.ListPaginatedResources(
+	err := repo.gateway.ListPaginatedResources(
 		repo.config.ApiEndpoint(),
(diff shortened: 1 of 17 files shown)


cf/api/service_plan.go
```diff
@@ -43,6 +44,22 @@ func (repo CloudControllerServicePlanRepository) Update(servicePlan models.Servi
 	return repo.gateway.UpdateResource(repo.config.ApiEndpoint(), url, strings.NewReader(body))
 }
 
+func (repo CloudControllerServicePlanRepository) ListPlansFromManyServices(serviceGuids []string) (plans []models.ServicePlanFields, err error) {
```
jberkhahn: We try to avoid using inline return variable declarations (like plans is here) because it can be hella confusing. Would you guys be ok with changing this to just have a generic return type and then declare plans in the function?

cf/api/services.go
```diff
@@ -255,6 +256,22 @@ func (repo CloudControllerServiceRepository) FindServicePlanByDescription(planDe
 	return planGuid, apiErr
 }
 
+func (repo CloudControllerServiceRepository) ListServicesFromManyBrokers(brokerGuids []string) (services []models.ServiceOffering, err error) {
```
jberkhahn: same thing here, inline return variable declaration

jberkhahn:
Hey guys, just had a few stylistic issues. We try to shy away from inline declaration of return variables because it can be hard to parse.

Gerg:
Your word is my command.

cf/api/services.go
rmasand: This is how most functions in the file (including the one we were modeling ours off of) were returning, so we thought that was desirable. We can go ahead and change it if that's the convention you're moving toward.

jberkhahn: Yeah, it's another one of those things we're trying to change by changing it wherever we see it. Sorry for these sorts of hidden trip ups.

Gerg:
We fixed the style issues.

PR #484, opened 2015-06-19
Change api endpoint for listing service keys

zhang-hua:
CLI should use the endpoint `/v2/service_instances/:fake-guid/service_keys`
to list service keys instead of using endpoint `/v2/service_keys?q=service_instance_guid:fake-guid`

[#87481016]

### DIFF
--- a/cf/api/service_keys.go
+++ b/cf/api/service_keys.go
@@ -59,13 +59,13 @@ func (c CloudControllerServiceKeyRepository) CreateServiceKey(instanceGuid strin
 }
 
 func (c CloudControllerServiceKeyRepository) ListServiceKeys(instanceGuid string) ([]models.ServiceKey, error) {
-	path := fmt.Sprintf("/v2/service_keys?q=service_instance_guid:%s", instanceGuid)
+	path := fmt.Sprintf("/v2/service_instances/%s/service_keys", instanceGuid)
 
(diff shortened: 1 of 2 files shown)


zhang-hua:
@shalako @maximilien this PR is to change the api endpoint to list service keys.
@simonleung8 @jberkhahn : please help us review the code and merge it. Thanks!

simonleung8:
@zhang-hua This new endpoint only exists in CF release 206 and up, the code path should only use this endpoint with the newer releases, else it should revert back to the old endpoint.

Here is an example of CLI checking the CC API version
https://github.com/cloudfoundry/cli/blob/master/cf/commands/service/update_service.go#L149

We are happy to take another look once this change is made. 
Thanks

zhang-hua:
@simonleung8 : I have checked the [API doc](http://apidocs.cloudfoundry.org/206) of CF release 206 and didn't found service key feature support. So it doesn't make sense to revert to old endpoint for service key. In addition, both of the endpoints are supported since service key features are developed. This change is just related to special use case to behave differently for SpaceAuditor and SpaceManager.  Please clarify your concern. Thanks!

simonleung8:
@zhang-hua , you are right, thanks for pointing that out.
The PR looks good, merged.

PR #487, opened 2015-06-22
Improve performance of disable-service-access

Gerg:
- It was making an `async=true` delete request for each
  service_plan_visibility. This meant each delete would take at least 5
  seconds due to polling.
- Deleting service plan visibilities does not interact with
  the broker and can be completed synchronously in ~.5s
- Add new http test matcher for testing empty query strings.

Tracker Story: https://www.pivotaltracker.com/story/show/96912324


simonleung8:
Thanks, all good and merged!

PR #490, opened 2015-06-24
Reduce API calls when creating,listing and getting details of service…

zhang-hua:
… keys

Leveraging existing API calls in ServiceInstanceRequirement to find service
instance info by name so that no need to send the same request twice.

[#93578300]


simonleung8:
thanks, merged

PR #505, opened 2015-07-01
Reduce API calls when CRU operations of service keys

zhang-hua:
Leveraging existing API calls in ServiceInstanceRequirement to find service
instance info by name so that no need to send the same request twice.

[#93578300]


zhang-hua:
@simonleung8 @jberkhahn would please help to review this new PR to fix this bug? Thanks!

simonleung8:
Thanks for the fix, merged.

PR #506, opened 2015-07-01
SpaceManager and SpaceAuditor should receive 403

zhang-hua:
Defined a new error NotAuthorizedError to capture CC returned
403 error

[#87481016]

### DIFF
--- a/cf/i18n/resources/fr_FR.all.json
+++ b/cf/i18n/resources/fr_FR.all.json
@@ -5368,5 +5368,10 @@
       "id": "{{.Usage}} {{.FormattedMemory}} x {{.InstanceCount}} instances",
       "translation": "{{.Usage}} {{.FormattedMemory}} x {{.InstanceCount}} instances",
       "modified": false
+   },
+   {
+      "id": "Server error, status code: 403, error code: 10003, message: You are not authorized to perform the requested action",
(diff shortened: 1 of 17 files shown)


shalako:
Does this take into account that the CLI team intends to revert PR #490?
https://www.pivotaltracker.com/story/show/98152534

So rather than just adding a fix, you may need to resubmit #490

cf/i18n/resources/fr_FR.all.json
```diff
@@ -5368,5 +5368,10 @@
       "id": "{{.Usage}} {{.FormattedMemory}} x {{.InstanceCount}} instances",
       "translation": "{{.Usage}} {{.FormattedMemory}} x {{.InstanceCount}} instances",
       "modified": false
+   },
+   {
+      "id": "Server error, status code: 403, error code: 10003, message: You are not authorized to perform the requested action",
+      "translation": "Server error, status code: 403, error code: 10003, message: You are not authorized to perform the requested action",
```
maximilien: Late, but here is the French translation for this new string:

```
"Erreur, code: 403, code de l'erreur: 10003, méssage: Vous n'êtes pas autorisé pour effectuer l'action demandée"
```

simonleung8: thanks Max.

PR #514, opened 2015-07-15
Fix create-app-manifest only includes one host [92530254]

SrinivasChilveri:
After fixing the issue cli create-app-manifest will cosiders
hosts,domains based on multiple routes to an app,so when ever
has more then one host or domain uses the hosts & domains
key word  respectively.
the changed code doesn't have perf impact if an app has single route


jberkhahn:
Neat! These tests are much clearer. Merged.

PR #523, opened 2015-07-22
Unmarshal the extra field, get documentation url

zachgersh:
- Also prepares for any other fields we'd like to
pull off of extra

Signed-off-by: Evan Farrar <efarrar@pivotal.io>


simonleung8:
Looks good. merged.

PR #566, opened 2015-09-01
remove inappropriate error message

EngineerBetter:
Remove a joke that could be at best confusing, and at worst culturally insensitive.

I've just seen this error message occur for a user whose first language was not English (making debugging harder for them), and who lives and works in a country that was once under British colonial rule: between 200,000 and 500,000 people died in the independence movement.


camelpunch:
Thanks for spotting this!

PR #584, opened 2015-09-16
Update README to add additional steps around preparing dev environment.  

### LINKED ISSUE #583: Instructions for creating/initializing development environment in README.md are incomplete
Some steps for Go newbies around navigating the directory structure are missing.

mcowger:
Fixes #583

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,13 +94,17 @@ Forking the repository for development
 ======================================
 
 1. Install [Go](https://golang.org)
-1. [Ensure your $GOPATH is set correctly](http://golang.org/cmd/go/#hdr-GOPATH_environment_variable)
-1. Install [godep](https://github.com/tools/godep)
+  * on Mac OS X this could be `brew install go` if you have [Homebrew](http://brew.sh/) installed.
+1. Create a directory for your source & dependencies ex. `mkdir ~/dev/cli && cd ~/dev/cli`
+1. [Ensure your $GOPATH is set correctly](http://golang.org/cmd/go/#hdr-GOPATH_environment_variable).  ex. `export $GOPATH=~/dev/cli`
+1. Install [godep](https://github.com/tools/godep) (usually `go get github.com/tools/godep`)
 1. Get the cli source code: `go get github.com/cloudfoundry/cli`
   * (Ignore any warnings about "no buildable Go source files")
 1. Run `godep restore` (note: this will modify the dependencies in your $GOPATH)
-1. Fork the repository
+1. Fork the repository on GitHub
 1. Add your fork as a remote: `cd $GOPATH/src/github.com/cloudfoundry/cli && git remote add your_name https://github.com/your_name/cli`
+1. Optionally add build output location to your path: ex. `export PATH=$GOPATH/src/github.com/cloudfoundry/cli/out/:$PATH`
+
 
 Building
 ========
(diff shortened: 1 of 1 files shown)


mcowger:
1. I understand your point.  I add it in there to help newbies who aren't go people (newbies like me).  Adding the brew syntax does potentially increase maint. over time, but the brew syntax has not changed in years, so it seems safe to me.

2. It will, but the instructions say to set `$GOPATH`, which you can't do until you actually have a path to set it to.  Does that make sense.  You can't get $GOPATH until you have a path.  Thoughts on this?

3.  Thanks for the catch.  Will fix.

4. The instructions as I've written do exactly that, although the example would result in a full path like `$~/dev/cli/src/github.com/cloudfoundry/cli`, which can be confusing.  Will adjust to make it clearer.

5. Understood and will adjust.

camelpunch:
1. Fair enough
2. Makes a bit more sense, but I think most Go devs use a GOPATH for all their Go development, not just one project. For example, I have ~/go as my GOPATH.
3. -
4. See above. Could be even more generic. I do feel like we're teaching people Go here, rather than CLI.

README.md
```diff
@@ -94,13 +94,16 @@ Forking the repository for development
 ======================================
 
 1. Install [Go](https://golang.org)
-1. [Ensure your $GOPATH is set correctly](http://golang.org/cmd/go/#hdr-GOPATH_environment_variable)
-1. Install [godep](https://github.com/tools/godep)
+  * on Mac OS X this could be `brew install go` if you have [Homebrew](http://brew.sh/) installed.
+1. Create a directory for your source & dependencies ex. `mkdir ~/dev/cloudfoundry-dev && cd ~/dev/cloudfoundry-dev`
```
camelpunch: I don't think there's any need to tell the user to `cd` here. It's not required by the rest of the instructions.

README.md
```diff
@@ -94,13 +94,16 @@ Forking the repository for development
...
+1. [Ensure your $GOPATH is set correctly](http://golang.org/cmd/go/#hdr-GOPATH_environment_variable).  ex. `export GOPATH=~/dev/cloudfoundry-dev`
+1. Install [godep](https://github.com/tools/godep) (usually `go get github.com/tools/godep`)
 1. Get the cli source code: `go get github.com/cloudfoundry/cli`
   * (Ignore any warnings about "no buildable Go source files")
 1. Run `godep restore` (note: this will modify the dependencies in your $GOPATH)
-1. Fork the repository
+1. Fork the repository on GitHub
 1. Add your fork as a remote: `cd $GOPATH/src/github.com/cloudfoundry/cli && git remote add your_name https://github.com/your_name/cli`
+1. Optionally add build output location to your path: ex. `export PATH=$GOPATH/src/github.com/cloudfoundry/cli/out/:$PATH`
```
camelpunch: If this step is designed for beginners, then we should also instruct the reader how to make the change persist.

simonleung8:
@mcowger Thanks for this PR, unfortunately we have decided not to pull it in for the following reasons.

- We don't want users to install from homebrew, it might not be the most updated version. Also the official Go installer works very good for Mac OS, you can install on top of an existing installation for version upgrade/downgrade.
- We don't want to provide terminal usage instructions, it is beyond the scope of CLI
- We have begun work on improving the README which conflicts with this PR

PR #587, opened 2015-09-17
better error reporting with feeded curl data is not enclosed.

Zouuup:
this will also help with this issue on PivotalTracker:
https://www.pivotaltracker.com/n/projects/892938/stories/95206110


Zouuup:
ahh, i just added a simple double quote and didn't checked test results afterward! :I let me fix this and reopen.

PR #588, opened 2015-09-17
better error reporting with feeded curl data is not enclosed.

Zouuup:
this will also help with this issue on PivotalTracker:
https://www.pivotaltracker.com/n/projects/892938/stories/95206110


Zouuup:
emm, i think there was a problem with cfdreddbot, my newer PR #590 has CLA passed, do i need to resend PR?

Zouuup:
there is a problem with corporate licence, i'll take care of that.

camelpunch:
Your username is on the bot's whitelist, not sure why it didn't pick it up.

camelpunch:
Minor: can we lower-case Argument in that error sentence please? Also, squashing the commits so we don't have undos in the history would be nice.

Zouuup:
Done sir ;)

PR #594, opened 2015-09-21
Fix Admin should be able to list apps in any space of current org

SrinivasChilveri:
Story in CLI [#91292742] for addition of the -s string flag in cf app……s command.

As of now this command lists only the apps in the currently targeted space. With this new flag, user can get the list of apps
in a space of current organization provided by the user.
Output after modification :
apps -s spaceName
Getting apps in org orgName / space spaceName as admin...
OK
name requested state instances memory disk urls
fib-cpu started 1/1 256M 1G fib-cpu.10.244.0.34.xip.io

### DIFF
--- a/cf/api/app_summary.go
+++ b/cf/api/app_summary.go
@@ -114,6 +114,7 @@ type DomainSummary struct {
 type AppSummaryRepository interface {
 	GetSummariesInCurrentSpace() (apps []models.Application, apiErr error)
 	GetSummary(appGuid string) (summary models.Application, apiErr error)
+	GetSpaceSummaries(spaceGuid string) (apps []models.Application, apiErr error)
 }
 
 type CloudControllerAppSummaryRepository struct {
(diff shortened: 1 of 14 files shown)


SrinivasChilveri:
Hi @camelpunch ,
I have fixed all your comments except one(comparing slices or structs with Equal(), ) .I feel fixing that is not a good idea, because the apps list is return value of the func which is local to every test case& we don't have a slice/struct which is defined globally. 

Any suggestions, comments are welcome.

Thanks & Regards,
SriniCH

camelpunch:
That particular suggestion isn't to do with test scope, but style. You would instantiate an expected object, then assert that it was Equal() with the actual one received from the implementation. Will have a chat with the team about the PR as it stands.

SrinivasChilveri:
Hi @camelpunch , Any updates on this. Do you want me to change according to your comment only for the new test case or for the old test cases also? please do update me so that i can update the PR

Thanks & Regards,
SriniCH

camelpunch:
@SrinivasChilveri I need to talk with the team about this. Will get back to you.

SrinivasChilveri:
Hi @camelpunch , Expecting a reply with the details to complete this PR.

Thanks & Regards,
SriniCH

nicw:
@SrinivasChilveri For more of a foundation, can you give us the user scenario for this? We have a potential solution that may solve another problem, but we need to know more about your workflow to make sure our solution won't be even worse for you. (see solution for listing the apps across all orgs that you manage) https://www.pivotaltracker.com/story/show/105045822).

Thanks! 
- CLI PM Team

SrinivasChilveri:
Hi @nicw , Thanks a lot for the reply. I had done this PR based on a story either in backlog/icebox of CLI tracker,not based on our work flow. But out of curiosity still even if we implement listing apps across all orgs of a manager still i feel need to have listing of apps related to a particular space.(may when we want to check apps of a a particular space not all spaces of org or all spaces of all orgs of a manager). But its up to you take decision.

Thanks & Regards,
SriniCH

krishicks:
@dkoper Would you please take a look at this and take action on it?

dkoper:
Closing:
I had left this PR open (for a year now!) to see if we would hear from more people sharing this need, and I haven't. We want to be careful not to support uncommon use cases that may end up cluttering the Interface.

PR #614, opened 2015-10-05
Improve messaging on zero space users

camelpunch:
...and refactor to separate printers.

My main concern with this change is that I've removed ui.Say()ing from the plugin case. Do plugins want those calls? There are no tests that say so.

### DIFF
--- a/cf/commands/user/space_users.go
+++ b/cf/commands/user/space_users.go
@@ -84,13 +84,11 @@ func (cmd *SpaceUsers) Execute(c flags.FlagContext) {
 
 	var usersMap = make(map[string]plugin_models.GetSpaceUsers_Model)
 
+	listUsers := cmd.getUserLister()
+
 	var users []models.UserFields
 	for role, displayName := range spaceRoleToDisplayName {
-		if cmd.config.IsMinApiVersion("2.21.0") {
-			users, apiErr = cmd.userRepo.ListUsersInSpaceForRoleWithNoUAA(space.Guid, role)
-		} else {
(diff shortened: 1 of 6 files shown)


camelpunch:
I also started iterating over a map, which is going to produce randomised ordering of role groups. Fix under way.

cf/commands/user/space_users.go
```diff
@@ -66,72 +83,94 @@ func (cmd *SpaceUsers) Execute(c flags.FlagContext) {
...
+func (cmd *SpaceUsers) getUserLister() func(spaceGuid string, role string) ([]models.UserFields, error) {
+	if cmd.config.IsMinApiVersion("2.21.0") {
+		return cmd.userRepo.ListUsersInSpaceForRoleWithNoUAA
+	}
+	return cmd.userRepo.ListUsersInSpaceForRole
+}
 
+func (p *pluginPrinter) printUsers(_ models.Organization, space models.Space, _ string) {
+	for role, _ := range p.spaceRoleToDisplayName {
+		users, _ := p.userLister(space.Guid, role)
```
maximilien: Why are you ignoring error returned by `func(spaceGuid string, role string) ([]models.UserFields, error)` function?

What if `p.userLister(space.Guid, role)` returned an error? Seems like ignoring is a bad idea...

camelpunch: Mainly because there are no tests that cover that case. I'll add a test before merging this. It's not finished, and will also go through team code review.

Issue #603, opened 2015-09-29
Plugin API GetCurrentSpace/Org weirdness


jberkhahn:
If you don't currently have a target, these methods return "" as the currently targeted space. While this is technically true, and I should probably be doing explicit validation on my end, wouldn't it be helpful to return an error like "no current target" or something?

camelpunch:
We did something similar for the UI recently. Please follow the Tracker story for updates on us getting this done for plugins.

PR #621, opened 2015-10-09
Error in plugin rpc when getting current space or org

jberkhahn:
Fix for https://github.com/cloudfoundry/cli/issues/603. Does this seem like a reasonable change? Currently its not possible for the error that function returns to ever be anything other than nil.


simonleung8:
@jberkhahn Thanks for this PR. Unfortunately we will not pull this in.

The API is returning the current state and no org/space being targeted is not an error.

PR #623, opened 2015-10-12
Refresh auth tokens from AccessToken endpoint

camelpunch:
Might need some help figuring out how to accept this story. I'll try to
construct a plugin that helps.


camelpunch:
Closing in favour of using the whole command, "oauth-token" instead of TokenRefresher.RefreshAuthToken()

PR #610, opened 2015-10-01
Add cf router-groups command

atulkc:
This PR is to add `cf router-groups` command to CF CLI. This command lets user list router groups.

This is first in the list of stories scheduled for supporting tcp routes in CF. Pivotal tracker story link for this PR: https://www.pivotaltracker.com/story/show/100975070

(Required OAuth scopes for CF CLI client are added to cf-release.)

Regards,
@atulkc and @fordaz
(CF Routing team)

PR #627, opened 2015-10-13
Add the type field to 'cf domains' command.

mdelillo:
This displays the type of router group for domains associated with router groups.

Below is the story in the routing team's tracker:
https://www.pivotaltracker.com/story/show/100981770

### DIFF
--- a/cf/api/resources/domains.go
+++ b/cf/api/resources/domains.go
@@ -11,6 +11,7 @@ type DomainEntity struct {
 	Name                   string `json:"name"`
 	OwningOrganizationGuid string `json:"owning_organization_guid,omitempty"`
 	SharedOrganizationsUrl string `json:"shared_organizations_url,omitempty"`
+	RouterGroupGuid        string `json:"router_group_guid,omitempty"`
 	Wildcard               bool   `json:"wildcard"`
 }
 
@@ -21,5 +22,6 @@ func (resource DomainResource) ToFields() models.DomainFields {
 		Guid: resource.Metadata.Guid,
 		OwningOrganizationGuid: resource.Entity.OwningOrganizationGuid,
 		Shared:                 !privateDomain,
+		RouterGroupGuid:        resource.Entity.RouterGroupGuid,
 	}
 }
--- a/cf/commands/domain/domains.go
+++ b/cf/commands/domain/domains.go
@@ -1,6 +1,8 @@
 package domain
 
 import (
+	"errors"
+
 	"github.com/cloudfoundry/cli/cf/api"
 	"github.com/cloudfoundry/cli/cf/command_registry"
 	"github.com/cloudfoundry/cli/cf/configuration/core_config"
@@ -12,10 +14,12 @@ import (
 )
 
 type ListDomains struct {
-	ui         terminal.UI
-	config     core_config.Reader
-	orgReq     requirements.TargetedOrgRequirement
-	domainRepo api.DomainRepository
+	ui              terminal.UI
+	config          core_config.Reader
+	orgReq          requirements.TargetedOrgRequirement
+	domainRepo      api.DomainRepository
+	routingApiRepo  api.RoutingApiRepository
+	routerGroupsMap map[string]models.RouterGroup
 }
 
 func init() {
@@ -47,6 +51,8 @@ func (cmd *ListDomains) SetDependency(deps command_registry.Dependency, pluginCa
 	cmd.ui = deps.Ui
 	cmd.config = deps.Config
 	cmd.domainRepo = deps.RepoLocator.GetDomainRepository()
+	cmd.routingApiRepo = deps.RepoLocator.GetRoutingApiRepository()
+
 	return cmd
 }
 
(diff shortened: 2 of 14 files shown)


krishicks:
This PR builds on code that's been since removed from the cli; see e68ce0f. @simonleung8, can you elaborate on why @atulkc's PR was removed?

simonleung8:
Hi, @atulkc's PR was reverted because the command is returning a poorly constructed error message to non-privileged users. A new PR is in the work to fix the issue. See #632 

We will have put this PR on hold till #632 is processed.

yuzhangcmu:
We created a new pull request: #645 , which include #610 and #632 . Please merge #645 first . Let us know if we need to resubmit the PR.

@markstgodard @yuzhangcmu

krishicks:
Hi @mdelillo,

I reviewed your PR and had a hard time following it:
* routerGroupsMap was being stored on cmd, which seems to be only so the unexported method `getRouterTypeFromGuid` would have access to it. 
* getRouterTypeFromGuid does a nil check on that very member to see if it should populate it, which I found surprising.
* A test which says it returns an error regarding an invalid router group guid was actually asserting the response was "Failed fetching router groups".

I'd like to suggest the following change (a diff is below).

In this version populating routerGroupsMap, which is now local to the func, happens initially, and the loop only has to care about looking up the value. This makes it very easy to comprehend, since the loop isn't relying on any more methods that might do surprising things.

It also changes one of the tests to show the correct output when a domain has a guid that can't be found in routingApiRepo.

```go
func (cmd *ListDomains) printDomainsTable(domains []models.DomainFields) {
	table := cmd.ui.Table([]string{T("name"), T("status"), T("routing")})

	routerGroupsMap := map[string]models.RouterGroup{}
	cb := func(routerGroup models.RouterGroup) bool {
		routerGroupsMap[routerGroup.Guid] = routerGroup
		return true
	}

	err := cmd.routingApiRepo.ListRouterGroups(cb)
	if err != nil {
		cmd.ui.Failed(T("Failed fetching router groups.\n{{.Err}}", map[string]interface{}{"Err": err.Error()}))
	}

	for _, domain := range domains {
		if domain.Shared {
			if len(domain.RouterGroupGuid) == 0 {
				table.Add(domain.Name, T("shared"))
			} else {
				routerGroup, ok := routerGroupsMap[domain.RouterGroupGuid]
				if !ok {
					cmd.ui.Failed(T("Invalid router group guid"))
				}

				table.Add(domain.Name, T("shared"), routerGroup.Type)
			}
		} else {
			table.Add(domain.Name, T("owned"))
		}
	}

	table.Print()
}
```

Thoughts?

Here's the diff, which you can apply by piping it through `sed` to add the trailing newline, and then tell `git` to apply it: `pbpaste | sed  -e '$a\' | git apply --ignore-whitespace`).

```diff
diff --git a/cf/commands/domain/domains.go b/cf/commands/domain/domains.go
index a23aa45..6225541 100644
--- a/cf/commands/domain/domains.go
+++ b/cf/commands/domain/domains.go
@@ -1,8 +1,6 @@
 package domain
 
 import (
-	"errors"
-
 	"github.com/cloudfoundry/cli/cf/api"
 	"github.com/cloudfoundry/cli/cf/command_registry"
 	"github.com/cloudfoundry/cli/cf/configuration/core_config"
@@ -14,12 +12,11 @@ import (
 )
 
 type ListDomains struct {
-	ui              terminal.UI
-	config          core_config.Reader
-	orgReq          requirements.TargetedOrgRequirement
-	domainRepo      api.DomainRepository
-	routingApiRepo  api.RoutingApiRepository
-	routerGroupsMap map[string]models.RouterGroup
+	ui             terminal.UI
+	config         core_config.Reader
+	orgReq         requirements.TargetedOrgRequirement
+	domainRepo     api.DomainRepository
+	routingApiRepo api.RoutingApiRepository
 }
 
 func init() {
@@ -83,52 +80,36 @@ func (cmd *ListDomains) fetchAllDomains(orgGuid string) (domains []models.Domain
 	return
 }
 
-func (cmd *ListDomains) getRouterGroupTypeFromGuid(guid string) (string, error) {
-	if cmd.routerGroupsMap == nil {
-		cmd.routerGroupsMap = map[string]models.RouterGroup{}
-		cb := func(routerGroup models.RouterGroup) bool {
-			cmd.routerGroupsMap[routerGroup.Guid] = routerGroup
-			return true
-		}
+func (cmd *ListDomains) printDomainsTable(domains []models.DomainFields) {
+	table := cmd.ui.Table([]string{T("name"), T("status"), T("routing")})
 
-		apiErr := cmd.routingApiRepo.ListRouterGroups(cb)
-		if apiErr != nil {
-			return "", apiErr
-		}
+	routerGroupsMap := map[string]models.RouterGroup{}
+	cb := func(routerGroup models.RouterGroup) bool {
+		routerGroupsMap[routerGroup.Guid] = routerGroup
+		return true
 	}
 
-	routerGroup, ok := cmd.routerGroupsMap[guid]
-	if ok == false {
-		return "", errors.New(T("Invalid router group guid"))
+	err := cmd.routingApiRepo.ListRouterGroups(cb)
+	if err != nil {
+		cmd.ui.Failed(T("Failed fetching router groups.\n{{.Err}}", map[string]interface{}{"Err": err.Error()}))
 	}
 
-	return routerGroup.Type, nil
-}
-
-func (cmd *ListDomains) printDomainsTable(domains []models.DomainFields) {
-	table := cmd.ui.Table([]string{T("name"), T("status"), T("routing")})
-
-	cmd.routerGroupsMap = nil
-
 	for _, domain := range domains {
 		if domain.Shared {
-			if domain.RouterGroupGuid == "" {
+			if len(domain.RouterGroupGuid) == 0 {
 				table.Add(domain.Name, T("shared"))
 			} else {
-				routerGroupType, err := cmd.getRouterGroupTypeFromGuid(domain.RouterGroupGuid)
-				if err != nil {
-					cmd.ui.Failed(T("Failed fetching router groups.\n{{.Err}}", map[string]interface{}{"Err": err.Error()}))
+				routerGroup, ok := routerGroupsMap[domain.RouterGroupGuid]
+				if !ok {
+					cmd.ui.Failed(T("Invalid router group guid"))
 				}
 
-				table.Add(domain.Name, T("shared"), routerGroupType)
+				table.Add(domain.Name, T("shared"), routerGroup.Type)
 			}
-		}
-	}
-
-	for _, domain := range domains {
-		if !domain.Shared {
+		} else {
 			table.Add(domain.Name, T("owned"))
 		}
 	}
+
 	table.Print()
 }
diff --git a/cf/commands/domain/domains_test.go b/cf/commands/domain/domains_test.go
index 8b03af1..28e4824 100644
--- a/cf/commands/domain/domains_test.go
+++ b/cf/commands/domain/domains_test.go
@@ -111,7 +111,7 @@ var _ = Describe("domains command", func() {
 					Expect(ui.Outputs).To(ContainSubstrings(
 						[]string{"Getting domains in org", "my-org", "my-user"},
 						[]string{"FAILED"},
-						[]string{"Failed fetching router groups"},
+						[]string{"Invalid router group guid"},
 					))
 				})
 			})

```

yuzhangcmu:
@krishicks:
I and @mdelillo refactored the code as your comment:
1. Rebased the master branch as there is some test error if don't rebase.
2. Instead put routerGroupsMap as a field of struct, just pass it as a parameter. Why we don't  populating routerGroupsMap in the function locally because if there is not any domain has RoutergroupGuid, then we don't need to populate it. So we only populate it when necessary. And we make a function to do it to separate the logic. That is one function do only one thing.
3. A test which says it returns an error regarding an invalid router group guid was actually asserting the response was "Failed fetching router groups".
-- We changed the test case and the code to display the right information. 

Thanks.

@mdelillo  @yuzhangcmu

PR #631, opened 2015-10-14
Fix 'cf stack' to display acurate error message for incorrect usage

acrlewis:
#### Before fix:
````
➜  cli git:(master) cf stack
FAILED
Incorrect Usage. Requires app name as argument

NAME:
   auth - Authenticate user non-interactively

USAGE:
   cf auth USERNAME PASSWORD

WARNING:

EXAMPLE:
   cf auth name@example.com "my password" (use quotes for passwords with a space)
   cf auth name@example.com "\"password\"" (escape quotes if used in password)
````
#### After fix:
````
➜  cli git:(master) cf stack
FAILED
Incorrect Usage. Requires app name as argument

NAME:
   stack - Show information for a stack (a stack is a pre-built file system, including an operating system, that can run apps)

USAGE:
   cf stack STACK_NAME

OPTIONS:
   --guid      Retrieve and display the given stack's guid. All other output for the stack is suppressed.
````

### DIFF
--- a/cf/commands/stack.go
+++ b/cf/commands/stack.go
@@ -28,7 +28,7 @@ func (cmd *ListStack) MetaData() command_registry.CommandMetadata {
 	return command_registry.CommandMetadata{
 		Name:        "stack",
 		Description: T("Show information for a stack (a stack is a pre-built file system, including an operating system, that can run apps)"),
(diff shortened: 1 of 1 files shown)


acrlewis:
failed travis ci, reviewing changes, the error message correction is accurate, need to validate the command

PR #637, opened 2015-10-17
Fix 'cf stack' and 'cf stacks' to display accurate error message

acrlewis:
##### Before Fix
```
➜  cli git:(master) ✗ cf stack
FAILED
Incorrect Usage. Requires app name as argument

NAME:
   auth - Authenticate user non-interactively

USAGE:
   cf auth USERNAME PASSWORD

WARNING:

EXAMPLE:
   cf auth name@example.com "my password" (use quotes for passwords with a space)
   cf auth name@example.com "\"password\"" (escape quotes if used in password)
```
```
➜  cli git:(master) ✗ cf stacks 1   
FAILED
Incorrect Usage. Requires app name as argument

NAME:
   stacks - List all stacks (a stack is a pre-built file system, including an operating system, that can run apps)

USAGE:
   cf stacks
```
##### After Fix
```
➜  cli git:(master) ✗ ./out/cf stack
FAILED
Incorrect Usage. Requires stack name as argument

NAME:
   stack - Show information for a stack (a stack is a pre-built file system, including an operating system, that can run apps)

USAGE:
   cf stack STACK_NAME

OPTIONS:
   --guid      Retrieve and display the given stack's guid. All other output for the stack is suppressed.
```

```
➜  cli git:(master) ✗ ./out/cf stacks 1
FAILED
Incorrect Usage. Does not require an argument

NAME:
   stacks - List all stacks (a stack is a pre-built file system, including an operating system, that can run apps)

USAGE:
   cf stacks
```

### DIFF
--- a/cf/commands/stack.go
+++ b/cf/commands/stack.go
@@ -36,7 +36,7 @@ func (cmd *ListStack) MetaData() command_registry.CommandMetadata {
 
 func (cmd *ListStack) Requirements(requirementsFactory requirements.Factory, fc flags.FlagContext) (reqs []requirements.Requirement, err error) {
 	if len(fc.Args()) != 1 {
-		cmd.ui.Failed(T("Incorrect Usage. Requires app name as argument\n\n") + command_registry.Commands.CommandUsage("auth"))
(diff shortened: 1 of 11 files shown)


krishicks:
Closing this in favor of https://github.com/cloudfoundry/cli/pull/635, which uses a more standard message when no argument is required ("No argument required").

PR #638, opened 2015-10-19
Fix cf ssh with more than required args

SrinivasChilveri:
Currently we are allowing the execution of command for more than required num of args , for example
cf ssh appname1  appname2  etc  by ignoring the other than one arg but considering other commands which supposed to make fail with invalid msg to user.

Any suggestions/commands are welcome.


krishicks:
Hi @SrinivasChilveri,

This looks good to me. Thanks!

PR #646, opened 2015-10-23
Make the edge binaries easily accessable

squeedee:
The edge binaries are now harder to find.
I think this offers a good compromise between "don't download unless you know what you're doing" and "I know what I'm doing, where's the download?"


dkoper:
Making the edge binaries harder to find is kinda intentional to make the binaries that matter to most users easier to find.

PR #648, opened 2015-10-23
create-route command now takes optional context path

crhino:
This PR adds an optional parameter, `-c`, to the `create-route` command in order to specify a context path for a route.

Notably, in this PR we are always sending the `path` key in the JSON request to the CC, by default an empty string. We also modify the `RouteCreator` interface to take in a new path parameter and send in the empty string for `map-route` at the current time.

[#104020104](https://www.pivotaltracker.com/story/show/104020104)

### DIFF
--- a/cf/commands/route/create_route.go
+++ b/cf/commands/route/create_route.go
@@ -1,6 +1,8 @@
 package route
 
 import (
+	"fmt"
+
 	"github.com/cloudfoundry/cli/cf/api"
 	"github.com/cloudfoundry/cli/cf/command_registry"
 	"github.com/cloudfoundry/cli/cf/configuration/core_config"
@@ -13,7 +15,7 @@ import (
 )
 
 type RouteCreator interface {
-	CreateRoute(hostName string, domain models.DomainFields, space models.SpaceFields) (route models.Route, apiErr error)
+	CreateRoute(hostName, path string, domain models.DomainFields, space models.SpaceFields) (route models.Route, apiErr error)
 }
 
 type CreateRoute struct {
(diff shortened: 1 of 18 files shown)


cf/i18n/resources/en_US.all.json
```diff
@@ -3645,6 +3645,16 @@
       "modified": false
    },
    {
+      "id": "Optional context path. Path must include at least one character following a leading forward slash (/). Trailing slashes will be stripped. Requests received with a trailing slash will match.",
```
nicw: @crhino can you explain this a little more? If trailing slashes are being removed, then what does " Requests received with a trailing slash will match." mean? Do you mean to say that they'll be matched after the trailing slashed is removed?

cf/commands/route/create_route.go
```diff
@@ -29,13 +31,15 @@ func init() {
 }
 
 func (cmd *CreateRoute) MetaData() command_registry.CommandMetadata {
+	contextPathHelp := T(`Optional context path. Path must include at least one character following a leading forward slash (/). Trailing slashes will be stripped. Requests received with a trailing slash will match.`)
```
dkoper: Having discussed the meaning of the message with Shannon and Nic, we'd like to suggest a small change to:

`Optional context path. Path must include at least one character following a leading forward slash (/). Trailing slashes will be stripped, but requests received with a trailing slash will match.`

We'll pull it in after your update.

crhino:
@dkoper Made the changes and rebased on master. Sorry for the wait, was on vacation last week and this got lost in the backlog.

shalako:
At CLI team's request, we'll add a validation to return a meaningful error when API doesn't support the new option.

shalako:
Per @krishicks, the CLI team has decided to address a version checking strategy themselves later. So this PR is ready for evaluation.

shashwathi:
Will Resubmit this PR after @shalako accepts https://www.pivotaltracker.com/story/show/104020104

PR #654, opened 2015-10-29
Add the type field to 'cf domains' command.

yuzhangcmu:
@krishicks:
This displays the type of router group for domains associated with router groups.

Below is the story in the routing team's tracker:
https://www.pivotaltracker.com/story/show/100981770

As we don't have the permission to modify repo of @mdelillo in #627, we create a new PR here to replace the it, please close #627.

This PR contains all the changes in #627 and do the refactor:
1. We rebased the master branch as there is some test error if don't rebase.
2. Instead of put routerGroupsMap as a field of struct, now we just pass it as a parameter. Why we don't populating routerGroupsMap in the function "printDomainsTable" locally because if no domain has RoutergroupGuid, then we don't need to populate it. So we only populate it when necessary. So it is better to put it in the for loop and initialize it when necessary. And we make a function to do it to separate the logic. That is one function do only one thing.
3. Test case and the error message: We changed the test case and the code to display the right information.

@mdelillo @yuzhangcmu

### DIFF
--- a/cf/api/resources/domains.go
+++ b/cf/api/resources/domains.go
@@ -11,6 +11,7 @@ type DomainEntity struct {
 	Name                   string `json:"name"`
 	OwningOrganizationGuid string `json:"owning_organization_guid,omitempty"`
 	SharedOrganizationsUrl string `json:"shared_organizations_url,omitempty"`
+	RouterGroupGuid        string `json:"router_group_guid,omitempty"`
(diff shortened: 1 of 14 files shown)


shalako:
We're going to check that in the case where CLI has to hit routing-api, and the uri is not in config.json, we prompt the user to login.

PR #662, opened 2015-11-02
add note about where to run `godep restore`

rndstr:


nicw:
Hi @rndstr, thanks for the contribution. We are actually in the process of revamping this part of the documentation. Rather than go into detail here, we are encouraging others to look at the widely accessible resources for compiling w/Go. 

Thanks,
CLI team.

PR #663, opened 2015-11-03
Show better error message when routing api endpoint is not defined.

atulkc:
This PR adds a specific error message when trying to execute "cf router-groups" and the Routing Api Endpoint is not defined in the ".cf/config.json". The i18n resources have been updated as well. For more details please check [#100975070]

Thanks,
Fermin Ordaz, Mike Smykowski, Atul Kshirsagar

### DIFF
--- a/cf/api/routing_api.go
+++ b/cf/api/routing_api.go
@@ -31,7 +31,7 @@ func (r routingApiRepository) ListRouterGroups(cb func(models.RouterGroup) bool)
 
 	routingApiEndpoint := r.config.RoutingApiEndpoint()
 	if routingApiEndpoint == "" {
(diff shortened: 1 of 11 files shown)


atulkc:
Added one more commit to this PR to change the error message to be displayed when routing api endpoint is not found in `~/.cf/config.json`

krishicks:
Hi @atulkc,

I'm unable to pull in this PR. Rather than adding error handling to routing_api, the command should be gated at the Requirements() phase.

I've added a new Requirement, RoutingAPIRequirement, which handles this case. I've also updated the wording, though that might yet change again. Please see 24d7634 for details.

Sorry to have made you do this PR on your own; I wasn't yet familiar with the code.

PR #673, opened 2015-11-09
Add support for port in create_route command

yuzhangcmu:
Add port flag "-p" to the create route command to be able to create a route with a given external port mapping. 

The story about the PR: https://www.pivotaltracker.com/story/show/100992242
 
Thanks,
@yuzhangcmu @fordaz

### DIFF
--- a/cf/api/fakes/fake_route_repo.go
+++ b/cf/api/fakes/fake_route_repo.go
@@ -16,15 +16,27 @@ type FakeRouteRepository struct {
 		Error error
 	}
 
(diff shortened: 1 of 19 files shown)


yuzhangcmu:
#674

shalako:
At CLI team's request, we'll add a validation to return a meaningful error when API doesn't support the new option.

shalako:
Per @krishicks, the CLI team has decided to address a version checking strategy themselves later. So this PR is ready for evaluation.

shalako:
I'll accept on a branch before we submit a PR.

PR #674, opened 2015-11-09
modify error message of missing routing-api endpoint

yuzhangcmu:
We are updating the error message of the story: https://www.pivotaltracker.com/story/show/100975070

The error message returned when routing api url is not found in config file:
current: "Routing API not found. Please log in.".
desired: "Routing API uri missing. Please log in again and retry."

@yuzhangcmu @fordaz

Thanks!

### DIFF
--- a/cf/i18n/resources/de_DE.all.json
+++ b/cf/i18n/resources/de_DE.all.json
@@ -4250,8 +4250,8 @@
       "modified": false
    },
    {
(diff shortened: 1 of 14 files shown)


dkoper:
There must have been a misunderstanding.
The current message (`login in again to set the URI automatically`) is actually an improvement @krishicks made to the message Shannon and I decided on (`log in again and retry`).

Issue #692, opened 2015-11-24
cf push does not work when pushing from a directory that is a symlink


jasonkeene:
I ran into this on bosh-lite just trying to deploy some demo apps. `null` ends up getting sent to `/v2/resource_match` which causes a 500. I'll open a pull request for the cli but the API should probably not return a 500 since it is a malformed request. I will file an issue on cloud_controller_ng as well.

```
$ cf --version
cf version 6.14.0+2654a47-2015-11-18
$ cd ~/workspace/
$ ln -s ~/workspace/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/assets/dora
$ cd dora
$ pwd
/Users/jasonkeene/workspace/dora
$ CF_TRACE=true cf push dora

....

Uploading dora...

REQUEST: [2015-11-23T16:59:24-07:00]
PUT /v2/resource_match HTTP/1.1
Host: api.bosh-lite.com
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
Content-Type: application/json
User-Agent: go-cli 6.14.0+2654a47 / darwin

null

RESPONSE: [2015-11-23T16:59:24-07:00]
HTTP/1.1 500 Internal Server Error
Content-Length: 99
Content-Type: application/json;charset=utf-8
Date: Mon, 23 Nov 2015 23:59:24 GMT
Server: nginx
X-Cf-Requestid: c119d2a6-f4cc-4391-5d83-051e1c6b8b70
X-Content-Type-Options: nosniff
X-Vcap-Request-Id: 26aec165-a0da-4053-4c56-b5ba0c1424e8::86676993-fe54-4232-a8f9-7400e09e3312

{
  "error_code": "UnknownError",
  "description": "An unknown error occurred.",
  "code": 10001
}

FAILED
Error uploading application.
Server error, status code: 500, error code: 10001, message: An unknown error occurred.
FAILED
Error uploading application.
Server error, status code: 500, error code: 10001, message: An unknown error occurred.
```

dkoper:
Hi @jasonkeene 

Thanks for your report and PR. We have fixed this in cf CLI release v6.15.0 published two days ago.
https://github.com/cloudfoundry/cli#downloads

Cheers,
Dies Koper
CF CLI PM

PR #693, opened 2015-11-24
Add support for pushing apps from directories that are symlinks

jasonkeene:
See #692


krishicks:
Hi @jasonkeene,

We've made modifications to GatherFiles that will require you to rebase before we can pull this in. See https://github.com/cloudfoundry/cli/blob/master/cf/actors/push.go#L39 for where this code has moved.

Also, can you please remove the TODO comment, and update this PR via an amend (don't add another commit on top).

krishicks:
This has been pulled in as be6ba73 (not from the PR).

jasonkeene:
@krishicks thanks!

Issue #564, opened 2015-08-30
[Issue/Question] .cfignore does not respect folders


jd-carroll:
Using the latest from mean.js, I installed the latest cli tools and I get the following:

```
$ cf push
Using manifest file c:\dev\artecha\inventory-management-meanjs\manifest.yml

Updating app mean in org inventory-management / space development as jdsalingerjr@gmail.com...
OK

Creating route mean-unsaccharine-arboriculture.cfapps.io...
OK

Binding mean-unsaccharine-arboriculture.cfapps.io to mean...
OK

Uploading mean...
FAILED
Error uploading application.
GetFileAttributesEx c:\dev\artecha\inventory-management-meanjs\node_modules\bower\node_modules\updat
e-notifier\node_modules\latest-version\node_modules\package-json\node_modules\got\node_modules\duple
xify\node_modules\end-of-stream\node_modules\once\node_modules\wrappy\package.json: The system canno
t find the path specified.
```

The `node_modules` is very clearly listed as an entry in the `.cfignore` file as `node_modules/`.

Is this a bug or am I missing something?

simonleung8:
The `node_modules/`will not be uploaded, however, CLI walks through all directories in order to determine what files need to be uploaded. Are you running into this issue under Windows? IIRC, Windows has a limit on how long a dir name can be.

KrekkieD:
I explicitly set `node_modules` to be ignored in `.cfignore` and still it iterates through this dir. This is a really annoying thing on windows which indeed occurs once paths become too long, which depends on the modules you have installed.

I think it should give the `.cfignore` paths more respect before attempting iterating over something that serves no purpose.

jd-carroll:
@simonleung8 That's completely ridiculous... Why would it care what's in an ignored directory? Can you please point me to where this code is and I will submit a PR to fix this issue.

petereberlein:
I ran into the same problem that the CLI cannot deal with long paths on Windows, however, in my use case the folder structure is intended, ie. there is no .cfignore in it because I need all the files in this way.

It is certainly a good idea to not traverse folders with a .cfignore in it and that solves the issue for the original reporter, but I'd appreciate it very much, if the actual problem (the use of ancient Windows APIs that cannot deal with long paths) was fixed as well.

culmor30:
[...]

PR #717, opened 2015-12-15
Add support for long filenames on windows - fixes #716

### LINKED ISSUE #716: cf push fails if paths are too long (more than about 260 chars) on windows
Pushing an app on windows with 'deep nested' paths (path length > 260) fails with 
```
FAILED
Error uploading application.
GetFileAttributesEx ...
```
This issue was already reported in #564 but is not fixed in client version 6.14.0: 
```
>cf --version
cf version 6.14.0+2654a47-2015-11-18
```
For reference please see:
- Go issue: https://github.com/golang/go/issues/3358
- Windows API (MAX_PATH):
 - https://msdn.microsoft.com/de-de/library/windows/desktop/aa364944%28v=vs.85%29.aspx
 - https://msdn.microsoft.com/de-de/library/windows/desktop/aa365247%28v=vs.85%29.aspx#maxpath

d024441:


krishicks:
This has been fixed in the latest edge release. See https://github.com/cloudfoundry/cli/issues/564#issuecomment-163785073

PR #722, opened 2015-12-18
Context path support

atulkc:
This PR adds support for specifying context path while creating route. It also consists of changes to list path in `cf routes` command.

Pivotal tracker stories:
https://www.pivotaltracker.com/story/show/104020832
https://www.pivotaltracker.com/story/show/104020104

Regards,
@atulkc and @fordaz
(CF Routing team)


shalako:
We'll close this PR when I've verified that the work we've done is available on edge.

Issue #734, opened 2016-01-20
Process @ params on cli command line by reading file/URL


mandarjog:
Enable @ params to enable command like the following
```
cf  set-env spring-music BLAH @abc.json
```

Where abc.json is
```
{
    "aa": 200,
    "bb": "xyz"
}
```

It also makes other command easy to use such as 
```
cf create-user-provided-service my-db-mine -p '{"username":"admin","password":"pa55woRD"}'
```
to
```
cf create-user-provided-service my-db-mine -p @crentials.json 
```

dkoper:
Hi @mandarjog 

Thanks for bringing this up.
We have a few stories around enabling a user to pass in input from a file instead of inline, the other way around, or either, summarized in #650. I have added set-env to the list.

We like consistency, and can't introduce changes in a CLI v6 release that could break existing scripts.
The '@' approach is consistent with the preferred solution for `cf curl` because of curl's use of `-d @file`.
However, it would break compatibility for users already scripting things like `cf set-env myapp domain-separator @`.
I will look into this further.

mandarjog:
The implementation only looks for patan value starting with a '@'
So domains will be fine.

It also ensures that if a file with that name is not available, the original value is retained. So existing scripts should not break.

Regards, 
Sent from my Ericsson Blaze

Mandar U Jog

dkoper:
How about `cf login -u user -p @abc123` - makes the CLI access the file system?
I don't think it's a good idea to introduce this a feature to *all* parameters, it's easy to overlook its impact - I'd prefer to introduce it to commands that users actually need it for.

mandarjog:
I understand from the backward compatibility standpoint it makes sense.

However the description
"If a param is prefixed with "@",  Cli treats it as a file indirection"

1. Is far more succinct.
2. It is also automatically applied to all cli / plugins. 
3. It is also *very* consistent way of dealing with indirections.
4. Leaves the door open for general URL indirection.

I understand the issue with unanticipated change in behavior, can that be addressed with a global flag like "param-indirection=false" ?


Regards, 
Sent from my Ericsson Blaze

Mandar U Jog

dkoper:
Few people know about/look for global flags. They're usually to configure the http proxy to use, set timeouts or debugging statements.

Users have only expressed a strong need for support for reading files for curl, cups/uups and now set-env.

I think we should go forward with the implementation of the feature, but only for these commands.

[...]

PR #735, opened 2016-01-20
re #743: Add @ param processing

mandarjog:
Add @ param processing in flags so it can be used
with any cli command

https://github.com/cloudfoundry/cli/issues/734

```
mjog@ mac ~/workspace/src/github.com/cloudfoundry/cli/out$ ./cf  set-env spring-music-1 BLAH @abc.json
Setting env variable 'BLAH' to '{ "aa": 200,
"bb": "xyz"
}
' for app spring-music-1 in org advancedservices / space dev as mjog@pivotal.io...
```

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -3,11 +3,14 @@ package flags
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"strconv"
 	"strings"
 
 	"github.com/cloudfoundry/cli/flags/flag"
 )
+
+const URLIndicatorPrefix = "@"
 
 type FlagSet interface {
 	fmt.Stringer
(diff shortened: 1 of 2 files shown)


mandarjog:
This saves customers from having to remember needless escape patterns for Mac/Linux, Windows and powershell.
This has caused confusion and errors with multiple customers.

chipchilders:
cla received @dkoper

krishicks:
We won't be pulling this in; it is too broad a change and would introduce confusion with commands that presently take a file parameter.

Additionally, I'm not keen on the implementation: a more appropriate thing for it to do would be to provide a Reader that could be used by a command. Reading the entire file at once isn't what we would necessarily want to do.

PR #747, opened 2016-01-29
add missing param route-path to manifest mapper

floriangosse:
I hope this is correct. It is my first line in go :tada: but untested.


floriangosse:
Could one of the maintainers check the failed test?

krishicks:
Sorry for the delayed response, @floriangosse. Normally someone should have responded sooner.

I'm not sure we're ready to pull in this feature yet. I'd like @dkoper, the PM of CLI, and @shalako to comment on whether this is something that can just be pulled in.

dkoper:
I had responded on #748. We are looking at a solution that will cater for the full range of routing features we are currently working on, rather than for one specific case.

PR #764, opened 2016-02-16
Avoid Risky operation of scaling 0 instances

ronakbanka:
Scaling application to 0 instances is risky , this PR just creates a prompt when scaling to 0 instances.

Output for this check.

```bash
./bin/build && ./out/cf scale hello-java-app -i 0

Generating Binary...

Are you sure you want to scale hello-java-app to 0 instances?>
```

### DIFF
--- a/cf/commands/application/scale.go
+++ b/cf/commands/application/scale.go
@@ -121,6 +121,15 @@ func (cmd *Scale) Execute(c flags.FlagContext) {
 
 	if c.IsSet("i") {
 		instances := c.Int("i")
+		if instances == 0 {
+			response := cmd.ui.Confirm(T("Are you sure you want to scale {{.AppName}} to 0 instances?",
+				map[string]interface{}{
+					"AppName": terminal.EntityNameColor(currentApp.Name),
+				}))
+			if !response {
+				return
+			}
+		}
 		params.InstanceCount = &instances
 	}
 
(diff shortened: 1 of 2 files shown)


dkoper:
Hi @ronakbanka 

`cf stop` basically has the same effect and we don't prompt for that?
Wouldn't your change risk existing scripts to suddenly stop working while waiting for someone to enter y/n, after a cf CLI upgrade?

ronakbanka:
@dkoper ``cf stop`` is specific for stopping app but with ``cf scale -i`` there is always risk for human error (maybe typing 0 instead of 10) . 

I understand that scripts need to be modified if someone is using cf scale for scaling down to 0 instances , but i think using cf stop is better option to do that operation ( as a command for stopping application).

Even for  ``cf stop`` prompt can be good option, what are your views on it?

CAFxX:
`cf app` will still return `requested state: started` though, not very clear...

```
$ cf scale myapp -i 0
Scaling app myapp in org test / space test as user...
OK

$ cf app myapp
Showing health and status for app myapp in org test / space test as user...
OK

requested state: started
instances: 0/0
usage: 512M x 0 instances
urls: myapp.example.com

   state   since   cpu   memory   disk   details

$
```

dkoper:
Apologies for the late follow-up.
I think the inconvenience to many script authors of having to revise their invocations of these commands to include a flag to suppress the confirmation, and the inconvenience to users of existing scripts when they find their scripts hanging on a new prompt, far outways the inconvenience of a few users (you're the only one bringing this up so far) who mistype the instance count or command name.
I don't believe it's common/necessary to put confirmation prompts on everything and the current policy (confirmation on delete/destroy only) is a reasonable one.

ronakbanka:
@dkoper Thanks , we can close this for now .

PR #804, opened 2016-03-28
Add support for route-path via the manifest file

tobocop:
Signed-off-by: Alpha Chen <achen@pivotal.io>

This code change will enable us to do blue-green deploys via the concourse autopilot resource.

### DIFF
--- a/cf/manifest/manifest.go
+++ b/cf/manifest/manifest.go
@@ -201,6 +201,7 @@ func mapToAppParams(basePath string, yamlMap generic.Map) (models.AppParams, err
 	appParams.EnvironmentVars = envVarOrEmptyMap(yamlMap, &errs)
 	appParams.HealthCheckType = stringVal(yamlMap, "health-check-type", &errs)
 	appParams.AppPorts = intSliceVal(yamlMap, "app-ports", &errs)
+	appParams.RoutePath = stringVal(yamlMap, "route-path", &errs)
 
 	if appParams.Path != nil {
 		path := *appParams.Path
--- a/cf/manifest/manifest_test.go
+++ b/cf/manifest/manifest_test.go
@@ -256,6 +256,7 @@ var _ = Describe("Manifests", func() {
 					"no-route":          true,
 					"no-hostname":       true,
 					"random-route":      true,
+					"route-path":        "my/route/path",
 				},
 			},
 		}))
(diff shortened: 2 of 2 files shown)


dkoper:
Hi @tobocop,

Note that we have a proposal (two actually) on supporting routes with paths in App Manifests to the CF Dev mailing list:
https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/message/N4ZDS7FVHBQFSGP724ZAXUOMISNVURIR/

Please take a look and join the discussion!

Closing this PR for now, its spec seems to make the problem the proposal tries to address worse.

lnhrdt:
@dkoper I've been trying to specify app routes with paths. After reading [the docs](https://docs.run.pivotal.io/devguide/deploy-apps/manifest.html#routes) I thought this would work:

```
```

I expected that what you proposed in "HTTP route example with proposal B:" would have worked, but

lnhrdt:
@dkoper I've been trying to specify app routes with paths. After reading [the docs](https://docs.run.pivotal.io/devguide/deploy-apps/manifest.html#routes) I thought this:

```
---
applications:
- name: my-app
  buildpack: java_buildpack
  path: path/to/my-app.jar
  routes:
  - route: my-domain.com/api
```

would result in my app being deployed with the route `my-domain.com/api`, but it seems the docs are incorrect because it's deployed with the route `my-app.cfapps.io` (in PCF).

I was about to post an issue since the docs suggested it would have worked but I found this issue here. In the doc you posted in the mailing list link above, you outlined this behavior as "HTTP route example with proposal B". Do you have any context on why the docs suggest this works but it reality it does not? Or have I encountered a bug?

I am using `cf version 6.21.1+6fd3c9f-2016-08-10`.

dkoper:
When I use the route mapping you listed above, I get `The route my-domain.com did not match any existing domains.`  as I would expect.
@lnhrdt Please submit a new issue to the CLI tracker. Comments on closed issues and PRs are easily lost.

lnhrdt:
In case anyone comes across this, I don't see this issue anymore. Declaring app routes with paths in manifest files works for me now. I'm unsure if when I experienced a problem there was an issue in the CLI or if it was just user error  on my part.

@dkoper I'll make sure to use the Tracker if I ever want to discuss an issue again.

PR #805, opened 2016-03-29
speed up marketplace command by making plan requests in parallel

theophoric:
Reduces response time from 1 minute 43 seconds to 7 seconds

### DIFF
--- a/cf/actors/service_builder/service_builder.go
+++ b/cf/actors/service_builder/service_builder.go
@@ -50,17 +50,35 @@ func (builder Builder) GetAllServicesWithPlans() ([]models.ServiceOffering, erro
 	if err != nil {
 		return []models.ServiceOffering{}, err
 	}
-
-	var plans []models.ServicePlanFields
+    
+    type planResponse struct {
+        index int
+        plans []models.ServicePlanFields
+        err error
+    }
+    
+    resChan := make(chan *planResponse, len(services));
+    
 	for index, service := range services {
-		plans, err = builder.planBuilder.GetPlansForService(service.Guid)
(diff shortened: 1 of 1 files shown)


ronakbanka:
+1

dkoper:
Hi @theophoric 

That is an impressive speed up!
I have a few concerns about performing API calls in parallel:
* what should API trace output look like: do we buffer request logs and log them together with the responses when they come in, or log them straight away (so you can see what's been sent before the call is made, but then more difficult to match to the response)
* what is reasonable default, and maximum numbers of calls to make in parallel, which don't make existing CF deployments fall over or worsen performance
* how to deal with API errors

I'd like to work with @SocalNick to see if the CC API can (or needs to) be enhanced to provide the information the CLI requires with lesser and/or faster API calls. First, we should take a look at the API calls we currently make and see if we can't optimize them using the existing CC API.

Regards,
Dies Koper
CF CLI PM

SocalNick:
Another way to speed this up considerably is page through `/v2/service_plans` instead of calling `/v2/service_plans?q=service_guid:<service_guid>` for each service offering.

cf/actors/service_builder/service_builder.go
```diff
@@ -50,36 +50,75 @@ func (builder Builder) GetAllServicesWithPlans() ([]models.ServiceOffering, erro
 	if err != nil {
```
Ruhassan: hi

dkoper:
Closing as we will not be merging this due to the concerns. We will explore ways to speed this up, incl. the different API call mentioned above.

PR #808, opened 2016-04-01
Compile linux binaries statically

### LINKED ISSUE #807: Compiled binaries are dynamically linked. Can we build them statically?
### CLI Version

Any version, including 6.16.1

### Platform Details

Ubuntu 14.04.3 64-bit, Alpine linux

### Error

The distributed binaries in the github releases are dynamically compiled with the libc. Because that, it must run in a system with the same libraries that the build server.

This prevents the users to use alternative systems and distros, like [alpine](https://hub.docker.com/_/alpine/).

You can test this with:

```
$ docker run -it alpine sh
alpine$ apk --update curl tar gunzip openssl ca-certificates
...
alpine$ curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.16.1&source=github-rel' | tar -xvzf -
...
alpine$ ldd cf-kkk 
	/lib64/ld-linux-x86-64.so.2 (0x55b92070a000)
	libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x55b92070a000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x55b92070a000)
alpine$  ls -l /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
ls: /lib64/ld-linux-x86-64.so.2: No such file or directory
ls: /lib64/ld-linux-x86-64.so.2: No such file or directory
ls: /lib64/ld-linux-x86-64.so.2: No such file or directory

alpine$ ./cf
sh: ./cf: not found
```

alpine says that `./cf: not found` but that is because it cannot exec the command.

# Proposed solution

Compile the project using `CGO_ENABLED=0`, as described in https://golang.org/cmd/cgo/

```
$ go get -d github.com/cloudfoundry/cli/
...
$ cd $GOPATH/src/github.com/cloudfoundry/cli
$ CGO_ENABLED=0 go build -o $GOPATH/bin/cf ./main/
...
$ ldd $GOPATH/bin/cf
	not a dynamic executable
$ file $GOPATH/bin/cf
/home/keymon/.gvm/pkgsets/go1.6/global/bin/cf: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), statically linked, not stripped
```

keymon:
In order to be able to use the linux binaries in different distributions
with different libc implementation (e.g. Linux Alpine), we must
compile it statically by setting CGO_ENABLED=0.

More info in https://golang.org/cmd/cgo/

Fixes #807

Refers to pivotal stories:
 * [#116760875](https://www.pivotaltracker.com/story/show/116760875)
 * [#115376361](https://www.pivotaltracker.com/story/show/115376361)

### DIFF
--- a/ci/tasks/build-binaries.yml
+++ b/ci/tasks/build-binaries.yml
@@ -24,13 +24,13 @@ run:
       bin/replace-sha
 
       echo "Building 32-bit Linux"
-      GOARCH=386 GOOS=linux go build -o out/cf-cli_linux_i686 ./main
+      CGO_ENABLED=0 GOARCH=386 GOOS=linux go build -o out/cf-cli_linux_i686 ./main
 
       echo "Building 32-bit Windows"
       GOARCH=386 GOOS=windows go build -o out/cf-cli_win32.exe ./main
 
       echo "Building 64-bit Linux"
-      GOARCH=amd64 GOOS=linux go build -o out/cf-cli_linux_x86-64 ./main
+      CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -o out/cf-cli_linux_x86-64 ./main
 
       echo "Building 64-bit Windows"
       GOARCH=amd64 GOOS=windows go build -o out/cf-cli_winx64.exe ./main


XenoPhex:
@keymon Thanks for your help with this problem! I'd imagine with a lot of people switching to Alpine, this would be a big problem for a lot of people.

Instead of using your pull request directly, we decided to go with a different approach, choosing to specify the linking behavior manually as opposed to relying on the `CGO_ENABLED`. We are worried about potential ramifications of disabling CGO, and are going to try a potentially safer approach first, as it also appears to generate a static binary.

Thanks again for your input and initiative in sending us this PR.

PR #811, opened 2016-04-05
Hcf 466 cli extended table

andreas-kupries:
Made as requested on the slack #cli channel by dkoper.


chendrix:
Hi @andreas-kupries , I think there might have been an issue introduced as part of this change

https://www.pivotaltracker.com/story/show/117404629

PR #812, opened 2016-04-06
added 3 min timeout for requests with 10 times retry

frankyhun:
We struggle with the issue, that the cloud doesn't responds to requests, and the cf tool just hangs. I added a timeout of 3 minutes to the requests, and increased the retry count to 10, and added command line logging.

### DIFF
--- a/cf/net/gateway.go
+++ b/cf/net/gateway.go
@@ -426,9 +426,10 @@ func (gateway Gateway) doRequest(request *http.Request) (response *http.Response
 
 	httpClient.DumpRequest(request)
 
-	for i := 0; i < 3; i++ {
+	for i := 0; i < 10; i++ {
 		response, err = httpClient.Do(request)
 		if response == nil && err != nil {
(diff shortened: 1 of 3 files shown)


chendrix:
Hi, @frankyhun! Thanks for this pull-request.

I'm trying to better understand your use case around this before merging/closing this PR. This PR seems like a band-aid over a larger networking issue you're having with your CF installation. 

* Is it only certain commands which are causing long hang times?
* Can you talk to your Cloud Controller API via other tools; is it just the CLI that's having issues? (Try running `curl YOUR.API.ADDRESS/v2/info`, e.g. `curl api.bosh-lite.com/v2/info`)
* Can you tell us the information you would have had to fill out in the [ISSUE_TEMPLATE](https://github.com/cloudfoundry/cli/blob/master/ISSUE_TEMPLATE.md)?

dkoper:
Closing. Feel free to open a new issue if you can share any of the info we asked for.

PR #821, opened 2016-04-19
Ref ticket 117404629. 

andreas-kupries:
Enabled the pending test, and added clarifying comments regarding the cause of the crash to the involved parts of the code.

The comments are future maintainers to know about the dependency between the two places and of having to keep them in sync.

### DIFF
--- a/cf/terminal/table.go
+++ b/cf/terminal/table.go
@@ -132,7 +132,23 @@ func (t *Table) calculateMaxSize(transformer rowTransformer, rowIndex int, row [
 		}
 
 		for i := range lines {
-			value := trim(Decolorize(transformer.Transform(columnIndex, lines[i])))
+			// (**) See also 'printCellValue' (pCV). Here
+			// and there we have to apply identical
+			// transformations to the cell value to get
+			// matching cell width information. If they do
+			// not match then pCV may compute a cell width
+			// larger than the max width found here, a
+			// negative padding length from that, and
+			// subsequently panic the Go runtime.  What
+			// was further missing is trimming before
+			// entering the user-transform. Especially
+			// with color transforms any trailing space
+			// going in will not be removable for print.
+			//
+			// This happened for
+			// https://www.pivotaltracker.com/n/projects/892938/stories/117404629
+
+			value := trim(Decolorize(transformer.Transform(columnIndex, trim(lines[i]))))
 			width := visibleSize(value)
 			if t.columnWidth[columnIndex] < width {
 				t.columnWidth[columnIndex] = width
@@ -171,8 +187,7 @@ func (t *Table) printRow(result io.Writer, transformer rowTransformer, rowIndex 
 				break
 			}
 
-			value := trim(transformer.Transform(columnIndex, row[columnIndex]))
-			t.printCellValue(line, columnIndex, last, value)
+			t.printCellValue(line, transformer, columnIndex, last, row[columnIndex])
 		}
 
 		fmt.Fprintf(result, "%s\n", trim(string(line.Bytes())))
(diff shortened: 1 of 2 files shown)


chendrix:
Hmm, looks like the test is failing on travis?

andreas-kupries:
That is weird. I ran ```ginkgo -p cf/terminal``` here before making the PR, and the newly enabled test succeeded, as did all the others.

And running ```CF_HOME=$(pwd)/fixtures ginkgo -r -randomizeAllSpecs``` ... Fail here too.
Reducing to ```CF_HOME=$(pwd)/fixtures ginkgo -r cf/terminal ```. Fail. Second try later on OK.
Reducing to ```ginkgo -r cf/terminal ```. Fail. Second try later on OK.
Changing to ```ginkgo -p cf/terminal ```. OK. After this trying the others above again, and its ok.

This feels like some race in the tests, maybe ?
I will investigate more.

andreas-kupries:
A bit more proof of the non-determinism, a copy of my terminal. I ran `cf/terminal` with `-race` active, thinking that it may be a race condition somewhere. While nothing is reported we clearly have an issue, with the first run succeeding the run immediately after reporting a test fail.

```
andreask@akucloud:~/Local/Go/src/github.com/cloudfoundry/cli> ginkgo -race cf/terminal
Running Suite: Terminal Suite
=============================
Random Seed: 1461172226
Will run 77 of 77 specs

•••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••
Ran 77 of 77 Specs in 0.383 seconds
SUCCESS! -- 77 Passed | 0 Failed | 0 Pending | 0 Skipped PASS

Ginkgo ran 1 suite in 17.344730763s
Test Suite Passed

andreask@akucloud:~/Local/Go/src/github.com/cloudfoundry/cli> ginkgo -race cf/terminal
Running Suite: Terminal Suite
=============================
Random Seed: 1461172257
Will run 77 of 77 specs

••
------------------------------
• Failure [0.002 seconds]
Table
/home/andreask/Local/Go/src/github.com/cloudfoundry/cli/cf/terminal/table_test.go:270
  REGRESSION: #117404629, having a space in one of the middle headers
  /home/andreask/Local/Go/src/github.com/cloudfoundry/cli/cf/terminal/table_test.go:59
    prints the table with the extra whitespace from the header stripped [It]
    /home/andreask/Local/Go/src/github.com/cloudfoundry/cli/cf/terminal/table_test.go:58

    expected to find "[watashi   no   atama!]" in actual:
    'watashi   no    atama!
    '
    

    /home/andreask/Local/Go/src/github.com/cloudfoundry/cli/cf/terminal/table_test.go:57
------------------------------
••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••

Summarizing 1 Failure:

[Fail] Table REGRESSION: #117404629, having a space in one of the middle headers [It] prints the table with the extra whitespace from the header stripped 
/home/andreask/Local/Go/src/github.com/cloudfoundry/cli/cf/terminal/table_test.go:57

Ran 77 of 77 Specs in 0.362 seconds
FAIL! -- 76 Passed | 1 Failed | 0 Pending | 0 Skipped --- FAIL: TestTerminal (0.51s)
FAIL

Ginkgo ran 1 suite in 3.196051409s
Test Suite Failed
```

Looking more closely at the failure it seems that for this case a space we expected to get trimmed, is not, causing the generation of 4 spaces between the columns instead of the expected 3. Unclear is, for now, why this happens only intermittently.

andreas-kupries:
I think I see the issue, although not yet why it is intermittent.

`printCellValue` puts the value into the output buffer as is, including possible trailing spaces (which may be hidden inside of colorization too). It then computes the padding based on the visible size, which excludes the trailing space(s) (And sees them due to `Decolorize`).

I have to look into the ordering of trimming and transforms in all places, it looks like. 
I am unhappy that I let this slip through.
OTOH, the non-determinism of the test result was unexpected.
Will make it difficult to asses if my changes are a proper fix or not.

andreas-kupries:
Updated the PR with new fix (additional trimming before transform).
This passed the full testsuite a 100 times without fail, making me believe that I actually caught it now.

challiwill:
@andreas-kupries thanks for the fix

PR #828, opened 2016-04-26
limit the -v effect position

jessexu20:
As we are talking about this...
In the old days, -v is the cf cli, I believe is short for the version... cf -v = cf --version...
however, a small changes like this will make it clear if user would like to see the cf debug information, or want to see the plugins' verbose result... 
with this change, user can specify `cf -v <plugin command> -v` and the first -v will handled by cf and show the cf debug info, while the second one is totally handled by the plugin itself :) @dkoper  

The same thing is also with `-h`, if the implementation could be done for `-h`, it could be better, since right now, the -h with plugin only shows the upper level help information that plugin registered with the cf in the metadata. However, maybe some plugin itself has the ability to show more detailed help info to the user... Limiting the scale of the global flags will improve that :) 

Thank you ...

Jesse

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -198,7 +198,7 @@ func handleVerbose(args []string) ([]string, bool) {
 		}
 	}
 
-	if idx != -1 && len(args) > 1 {
+	if idx == 1 && len(args) > 1 {
 		verbose = true
 		args = append(args[:idx], args[idx+1:]...)
 	}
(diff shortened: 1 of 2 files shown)


dkoper:
Global flags are global flags. Making them less global is confusing, as `-h` and `-v` would then not give the expected behaviour for some plugin commands.
Also, if a plugin wants to display a different type of help as is currently displayed for native and plugin commands, it can expose that with a different option.

I'm sorry, we've discussed it in the team and there was no support for changing the CLI global options.

Regards,
Dies Koper
CF CLI PM

jessexu20:
I understand... THank you Koper 👍

Issue #793, opened 2016-03-10
Cannot update a service_instance's plan that was made private after it was provisioned


markstgodard:
Given a service which supports "plan_updateable", 
User cannot update a service instance's plan if a service plan is made private after a service instance provisioned.

NOTE: @dkoper I can create a PR as well

(see related CC item: https://github.com/cloudfoundry/cloud_controller_ng/issues/561)

For example: (using bosh-lite)

Service has two plans, and another plan "fake-plan" which I made private (so does not show up here)

```
mark at evilcorp ~ $ cf m
Getting services from marketplace in org o / space s as mark...
OK

service                                             plans                                   description
fake-service-6dde3dd0-c875-4818-94fb-6b8d521bfa1a   fake-async-plan, fake-async-only-plan   fake service

TIP:  Use 'cf marketplace -s SERVICE' to view descriptions of individual plans of a given service.
```

I have a service instance, which was provisioned BEFORE fake-plan was set to private.
```
mark at evilcorp ~ $ cf s
Getting services in org o / space s as mark...
OK

name                  service                                             plan        bound apps   last operation
myservice             fake-service-6dde3dd0-c875-4818-94fb-6b8d521bfa1a   fake-plan                create succeeded
```

However when a service instance is updated to another plan (which is valid) it gets an error.

### Command

``` 
$ cf update-service myservice -p fake-async-plan
FAILED
Plan does not exist for the  service
```

### CLI Version

6.15.0+1076af9-2016-02-11

### Error

FAILED
Plan does not exist for the service 

The reason it is getting this error is likely due to: inline-relations-depth=1
```
GET /v2/spaces/1187a7d0-ba38-431c-a01a-0b657f741b7c/service_instances?return_user_provided_service_instances=true&q=name%3Amyservice&inline-relations-depth=1 HTTP/1.1
```

IMPORTANT NOTE: Simon and myself was looking at this, and Simon also created a CAPI GH Issue as well, because there is likely related discussion / potential changes to be made on the CC side..

### CF Trace

```
mark at evilcorp ~ $ CF_TRACE=true cf update-service myservice -p fake-async-plan

REQUEST: [2016-03-10T09:23:35-06:00]
GET /v2/spaces/1187a7d0-ba38-431c-a01a-0b657f741b7c/service_instances?return_user_provided_service_instances=true&q=name%3Amyservice&inline-relations-depth=1 HTTP/1.1
Host: api.bosh-lite.com
Accept: application/json
[...]

PR #831, opened 2016-05-04
Add optional service name to update-service

markstgodard:
Talked to @dkoper today regarding https://github.com/cloudfoundry/cli/issues/793 and submitting a PR to address this issue.

**IMPORTANT NOTE: This PR was updated to NOT require a new "-s" flag and we found a way to get the service guid from the service plan using /v2/spaces/guid/summary endpoint.
See comments below.**

Provide flag to specify the service name so that when updating a service instance that is currently associated with a private / disabled service plan it can be updated.

Without the -s flag, if you try to update a service instance that has a service plan that has been disabled, you will see an error like this:

```
cf update-service my-service -p fake-async-plan
FAILED
Plan does not exist for the  service
```

With this PR, you can specify the service offering name so that the CLI can properly fetch the current plans associated with the service offering. Previously, the service offering was being resolved via the current service plan. However in the case where a service plan has been made private, there is no service plan or associate service offering available in the response from /v2/service_instances/<guid>

New -s flag:

``` 
$ cf update-service my-service -p fake-async-plan -s fake-service
Updating service instance my-service as mark...
OK

Update in progress. Use 'cf services' or 'cf service my-service' to check operation status.
```

I also tested this locally on bosh-lite, if you wanted to integration test this what I did was:

- use CF acceptance tests asset to create a fake service broker
https://github.com/cloudfoundry/cf-acceptance-tests/tree/master/assets/service_broker
- provision a service instance using a regular (non admin user) using a fake-plan
- as admin, disable service access to that fake plan
- as regular user, try to cf update-service -p fake-sync-plan  (should get error)
- as regular user, use the new flag:  cf update-service -p fake-sync-plan -s fake-service-namewhatever
and it should update.

See full details related to this here:
https://github.com/cloudfoundry/cli/issues/793

Cheers


markstgodard:
Hi @dkoper and cli team

@shalako and I discussed this this morning. And Shannon found a way we could still update the service instance to a public plan when its current plan is private AND NOT have to add the new flag.

Using `/v2/spaces/:guid/summary` you can match the existing service instance name, and determine the associated service offering guid. You can you then call `/v2/service_plans?q=service_guid:guid` and match the desired plan name to its guid. Then you can call the `PUT /v2/service_instances/:guid` to update the plan.

With this approach we should be able to address the issue AND not have to include a new flag.

If you are ok with that, I can resubmit this PR with these changes.

Mark && Shannon

shalako:
@dieucao @SocalNick Seems like having `GET /v2/service_instances` return `service_plan_guid: null`  when plan is private may be unnecessarily restrictive. Is that a recent change? 

If the instance has a plan (private or not), and the client has permissions to list the instance, shouldn't the client be able to determine the plan without resorting to crazy summary endpoints or inline-relations-depth params?

SocalNick:
@shalako I agree, it seems like we don't need to hide that. However, I think we can accomplish this w/o a new flag and w/o jumping through all these hoops. I'm rushing to get this done and get to the bus, but here's current requests and proposed requests for accomplishing updating a plan when the current plan isn't visible:

```
Update from Plan that is Visible:

GET /v2/spaces/:space_guid/service_instances?return_user_provided_service_instances=true&q=name%3A:service_instance_name&inline-relations-depth=1
WHY Get the :service_instance_guid and (through the service_plan) get the :service_guid

GET /v2/services/:service_guid
WHY - Maybe for the service_plans_url, but it's not used in the next request

GET /v2/service_plans?q=service_guid%3A:service_guid
WHY Get visible service plans

GET /v2/organizations?q=name%3A:org_name&inline-relations-depth=1
WHY?

GET /v2/service_plan_visibilities
WHY - it's empty in this case

PUT /v2/service_instances/:service_instance_guid?accepts_incomplete=true -d '{"service_plan_guid":":service_plan_guid","tags":[]}'



Update from Plan that is NOT visible:
GET /v2/spaces/:space_guid/service_instances?return_user_provided_service_instances=true&q=name%3A:service_instance_name&inline-relations-depth=1
WHY Get the :service_instance_guid and (through the service_plan) get the :service_guid

GET /v2/service_plans?q=service_guid%3A
WHOOPS - Can't get the service_plan

GET /v2/organizations?q=name%3Ancalugar-a1&inline-relations-depth=1
WHY?

GET /v2/service_plan_visibilities
WHY - still empty

FAILED
Plan does not exist for the  service



Proposed requests:
GET /v2/spaces/:space_guid/summary
WHY - Get the :service_instance_guid and the :service_guid

GET /v2/services/:service_guid/service_plans
WHY - Find the desired :service_plan_guid

PUT /v2/service_instances/:service_instance_guid?accepts_incomplete=true -d '{"service_plan_guid":":service_plan_guid","tags":[]}'
```

markstgodard:
Hi @dkoper && @shalako 

I updated this PR to NOT require the "-s" flag and as discussed above, I was able to use the `/v2/spaces/guid/summary' CC API to fetch the service instances and match by service instance name.
I was able to get the service guid from the Service offering and most of the code changes were the same.

So implementation wise its cleaner and also from a UX perspective it does not change the CLI which is what we wanted to avoid in the first place.

As mentioned above, ideally it would be nice to still be able to get the service plan info for a service instance, even if it is now marked private.  However this PR as it is right now will work if the /v2/service_instances endpoint does not change.

Thanks to Shannon for pinging me with "I am pretty sure there is a way to get this information" :)

Cheers

shalako:
That looks good, Nick.

markstgodard:
Hi @SocalNick 

cc @shalako @dkoper 

Thanks for the update.

So as of now this PR is updated to:

- not introduce a new flag
- uses** /v2/spaces/<guid>/summary** to find the service instance's current plan's service guid :)
- (rest of the logic in the CLI is the same)

So Nick based on your comment about Proposed Requests, this updated PR is in line with how to get the service guid.

So there is no plan to change the response from GET /v2/service_instances/<guid> to return the service plan info (even if it is private) ?  Is that not what you originally were asking @shalako ?
If it was, then this PR would not be required and we would not have to resolve the service guid for a service instance using this /v2/spaces/guid/summary endpoint.

However I am not sure if the implications.

I am fine with going thru /v2/spaces/<guid>/summary. Its definitely better than adding a new flag which I did NOT want to do.


Cheers

SocalNick:
@markstgodard i'm glad we've figured a way to unblock this without adding the flag or changing the API. @shalako let's talk next week about the service plan visibility issue.

SocalNick:
Related: https://github.com/cloudfoundry/cloud_controller_ng/issues/529

dkoper:
@SocalNick @shalako Please let us know how you went. Mark will update the PR accordingly after which I'll add it to our backlog.

SocalNick:
We intend to fix the null service plan id in https://www.pivotaltracker.com/story/show/119333093

Up to you if you want to wait or fix now with the summary endpoint.

markstgodard:
Hi @SocalNick , @dkoper 

Sure, I can wait for the CC fix.

Just to clarify, was the CC change to fix the /v2/spaces/space_guid/service_instances/service_instance_guid to ensure that it returns the "service_plan" portion in the response?

If so, then the CLI would then require zero changes.

Right now the CLI is getting the "service_guid" from the "service_plan" section of the response.

I'll close this PR and we can wait for the CC fix.    You can also make a note or let me know, and we can re-test this scenario as part of acceptance for the CC fix.

Cheers

SocalNick:
@markstgodard I think (#119333093)[https://www.pivotaltracker.com/story/show/119333093] will work for you, feel free to follow the story.

markstgodard:
@dkoper see above for details ^^  resubmitting to fix this in the CF CLI.

I rebased and had to resubmit a new PR: https://github.com/cloudfoundry/cli/pull/963

Cheers

PR #839, opened 2016-05-11
Japanized Bind cancelled, Creating route, Creating service broker and…

MNitta:
Japanized Bind cancelled, Creating route, Creating service broker and some incorrect usages messages.


challiwill:
@MNitta Thanks!! I've merged this in. It will be included in the next release.

PR #881, opened 2016-07-05
Added possibility to use different OAuth client in cf cli

s-matyukevich:
The use case for this pull request is the following:

in 18F we are forced to limit user password expiration period to 1 month due to government restrictions. However we want to allow our users to set up special accounts for continuous integration. For this accounts we want to provide our users long lived tokens, that will have large expiration time (more that 1 month, probably 6 month) In order to do this we are going to set up different OAuth client and configure it appropriately. But in cf cli OAuth client name is hard coded, so I write this PR to eliminate this.

If it appears that  this functionality is not something you want to see in the upstream, please, let me know how our use case can be handled in a different way.


dkoper:
Hi @s-matyukevich 

Can you explain more about the UX of your enhancement? I'd like to understand that part before the team looks at your code.
Is this a refactor that users would not see, or are you introducing a `cf config` option, or a `cf oauth` option for users to set?

@sreetummidi, is a special user with long-lived token the recommended approach for CI environments?

Cheers,
Dies Koper
CF CLI PM

s-matyukevich:
Hi @dkoper 

The UX as I see it now is the following:

1. Platform administrators cretes additional UAA client and configure it to issue long lived tokens (I described approximate steps how to do this here https://docs.google.com/document/d/1r3hxsqbqtx6a7fUcNsmQBwWP49FnDBf1R5u7XIIVQXw/edit)

2. When some user requires a long lived token we will issue such token for him (for now I don't know how, maybe we will use some additional web app, or cf cli plugin, or user by himself will use UAAC to obtain such token) 

3. User modifies his cf cli config and set token there. This also can be done in many different ways either manually, or by using cf cli plugin, or by using some custom plugin for CI/CD tool, that user works with.

Best Regards,
Matyukevich Sergey.

cf/api/authentication/authentication.go
```diff
@@ -207,7 +207,8 @@ func (uaa UAARepository) getAuthToken(data url.Values) error {
 	}
 
 	path := fmt.Sprintf("%s/oauth/token", uaa.config.AuthenticationEndpoint())
```
huyvu3856: Cf

sreetummidi:
@dkoper We have plans to address this using long lived refresh tokens. We need to expose this in UAA first before making in available in CF CLI 

@s-matyukevich Can you please elaborate what this pull request entails. I was under the impression that we can plugin in an access token into the config.json without any code changes in the CLI

s-matyukevich:
@sreetummidi  yes, you are right about the possibility to add access token to the config. However then we can't use refresh token mechanism. As a result, if we delete our user or update his rights - his old access token still be valid.
This pull request eliminates this, it allows us to issue just a refresh token for the user and only refresh token in that case need to have long expiration time.
.

sreetummidi:
@s-matyukevich 
What we are trying to do on the UAA side to find a permanent fix is to allow users to generate Long Lived Refresh tokens based on the associated OAuth Client. These tokens will be revocable.

Until we have that in place, I don't see a problem with this approach.
From a UX Perspective, the end user needs to specify the OAuth Client Name and Client Secret for Automated Access. Is that correct?

s-matyukevich:
@sreetummidi  yes, you are right, the only thing that is changed from UX standpoint is  possibility for a user to specify in a config file two additional properties: CFOAuthClient and CFOAuthClientSecret.

cf/api/authentication/authentication.go
s-matyukevich: Don't understand what do you mean?

s-matyukevich:
@sreetummidi , @dkoper 

Any updates on this?

dkoper:
You seem to indicate [there is a chance](https://github.com/18F/cg-docs/issues/286#issuecomment-230576235) this modification is no longer required? Do you have an update on that?

I'm concerned I don't understand very well the impact of this change. In your reply to my question about that you wrote that the user copy & pastes their token into config.json. Two questions later you added two properties to the config file? Are you sure there is nothing more?

sreetummidi:
@dkoper If you have time today lets synch up on this. I don't view this as a breaking change.

s-matyukevich:
@sreetummidi , @dkoper

Any updates on this? If this PR can't be merged right now, can you at least estimate when it can be merged?

sreetummidi:
@s-matyukevich I had a meeting with Dies about this feature and we have decided to wait and fix the issue the right way in UAA and CLI through the use of User based API tokens.

We are currently doing this work on the UAA side and then the functionality will be exposed in the CLI.  A rough timeline is beginning/mid september.

mogul:
@sreetummidi , @dkoper: Thanks for addressing this upstream... It has always been strange to not be able to refer our users to using tokens for CF/UAA the way that most other developer-facing platforms do. Is there a particular issue in GitHub or PT that I could follow to keep up with progress on this change?

Can you speak to how this new functionality will appear for users or CD systems? In particular, we have users making use of Travis, CircleCI, and Wercker, all of which have the user providing a username+password in their docs. It'd be nice if they were able to pick up this change smoothly.

dkoper:
Reopening. @sreetummidi and I have restarted discussions about supporting long lived tokens in the cf CLI, and making the client id and secret configurable is a pre-requirement.

@s-matyukevich We're in the middle of a refactor - no need to rebase this PR, my team do the work to address it in the new architecture.

mogul:
FYI: We still don't have a good sense of what will happen here UX-wise, and figure it'll take even longer for CI services to update to using tokens rather than username/password. So in the interim we're solving this via [a broker that creates space-specific deployer accounts in a special zone with longer-lived passwords](https://github.com/18F/deployer-account-broker).

jeffpak:
Hello @s-matyukevich!

We took a look at this PR, and it's passing all of our unit and integration tests. When we tried to run the CLI manually however, the cli complains about the CFOAuthClient when logging in.  We think this is because we already had a previous cf config that did not contain CFOAuthClient and CFOAuthClientSecret. These are set to defaults when we re-enter `cf api <api-endpoint>`; if you don't do this, it complains with `The targeted API endpoint could not be reached.`

Specifically, because we aren't defaulting to "cf" when the CFOAuthClient is blank, the credentials get rejected. We believe that the CFOAuthClient() function in cf/configuration/coreconfig/config_repository.go should be configured to handle that default value. 

We'd like to have this edge case fixed before accepting this PR, since we're certain that other CLI users will experience similar problems. We were surprised that our integration tests did not catch this, so we're wondering if anything else might have slipped our notice. 

To reproduce our steps:
1. Delete the current cf config.
1. Connect to the API endpoint.
1. Delete the two new fields: CFOAuthClient and CFOAuthClientSecret.
1. Try logging in. 

For existing users of the CLI, we believe that step 3 may happen since they weren't existing fields prior. So when they try logging in again as a separate user, they're going to encounter an error because they hadn't reset their API endpoint.

Let us know if you have any questions

Regards,
@jeffpak && @mmb

jeffpak:
Hello again! Scratch what I said about fixing the PR, I think we'll just fix it ourselves as we work on putting this into the new architecture.

abbyachau:
Hi @s-matyukevich @mogul, I wanted to get in touch to let you know that we plan on deprecating this custom client workflow (allowing devs to manually write to their `config.json` to persist a long lived token) in favor of using [UAA's](https://docs.cloudfoundry.org/api/uaa/version/4.30.0/index.html#client-credentials-grant) solution of using `client credentials` for use cases such as CI/CD. 

The custom client workflow will continue to work on v6 cf CLI but we will start outputting a deprecation warning when we detect that this workflow is being used. In v7 cf CLI, we plan on removing the ability to use the custom client workflow in favor of client credentials. 

Please reach out if you have any questions, concerns or feedback. Thanks.

cc UAA PMs and stakeholders (@dbeneke @cwang-pivotal @aramprice @danjahner)

PR #926, opened 2016-08-16
trace: Redact all passwords and tokens in JSON

dcarley:
Extend the regex that redacts the values of sensitive keys in JSON responses
to include case-insensitive and substring matches of the key names
"password" and "token".

It's quite common to provide credentials to [12factor][0] applications using
environment variables. These can be unexpectedly leaked in the CLI output if
you run a command such as `CF_TRACE=true cf apps`, because the API endpoints
for space and application summaries always include `environment_json`.

This problem is mitigated by [user-provided services][1], but the docs don't
state that it's a security feature and it would be good to make the default
use case safer.

Some notes about the specific implementation, I have:

- added `(?i)` to the regex to make it a case-insensitive match
- added `[^"]*` around the property name to match substrings
- used the capture group `$1` to preserve the entire matched property name
  in the replacement
- removed the separate function calls for `access_token`, `refresh_token`
  and `oldPassword` because these are now covered by the existing `token`
  and `password` calls with the new regex
- adjusted one of the test assertions to account for the fact that the value
  of `token_type` is now also redacted. This is a side-effect of the
  changes, but I don't think it's a bad thing

[0]: http://12factor.net/config
[1]: https://docs.cloudfoundry.org/devguide/services/user-provided.html

### DIFF
--- a/cf/trace/trace.go
+++ b/cf/trace/trace.go
@@ -16,18 +16,15 @@ func Sanitize(input string) string {
 	re = regexp.MustCompile(`password=[^&]*&`)
 	sanitized = re.ReplaceAllString(sanitized, "password="+PrivateDataPlaceholder()+"&")
 
-	sanitized = sanitizeJSON("access_token", sanitized)
-	sanitized = sanitizeJSON("refresh_token", sanitized)
 	sanitized = sanitizeJSON("token", sanitized)
 	sanitized = sanitizeJSON("password", sanitized)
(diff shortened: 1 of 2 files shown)


dkoper:
Thanks again for bringing this up. The change is incorporated in the current release.
https://github.com/cloudfoundry/cli#downloads

Cheers,
Dies Koper
CF CLI PM

PR #927, opened 2016-08-16
Add the packaging metadata to build the Cloud Foundry CLI snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds for adventurous users.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: cf
+version: v6.21.1
+summary:  The official command line client for Cloud Foundry.
+description: |
+  A tool to deploy and manage Cloud Foundry applications.
+confinement: strict
+
+apps:
+  cf:
+    command: bin/cli
+    plugs: [network, network-bind]
+
+parts:
+    cli:
+      source: .
+      plugin: go
+      go-importpath: github.com/cloudfoundry/cli
+      go-packages: [github.com/cloudfoundry/cli]
+      build-packages: [g++, git]


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

If we upload the snap to the ubuntu store we will get transactional automatic updates. I've been testing the snap [1] and it seems to me good enough to publish in the edge channel, which will make it really easy to do crowd-testing before the stable release.

If you have any doubts or comments, please let me know :)

[1] https://gist.github.com/elopio/061584d5c5f4e3297104f34dec96d906

dkoper:
@elopio What does this give CLI users what publishing the cf CLI to a Debian package repository would not give?

elopio:
@dkoper this package works in more distros, not just debian and ubuntu. You can [see the list of distros here](http://snapcraft.io/docs/core/install), and there are more to come. Most of the distros already support all the security features, others still have a few things in progress. These security features mean that the app is totally isolated, it can't affect or be affected by other snaps and it can't access any data, devices or services that the user doesn't explicitly allows. The installation and updates are transactional, so you'll never end up with a broken system or app, and it can rollback to previously installed versions. Once it's published to the store it's really simple to switch between the edge, beta, candidate and stable channels depending on how early you want to start testing. And we are working on ubuntu images for IOT and prototyping boards like raspberry pi, dragonboard, intel joule. Every new enabled board will be able to use the app right away.

Those are my favorite features, quite a few because I'm enjoying a lot snapping free software projects :smile: There are some more, and some other niceties for devs and maintainers too.

dkoper:
I'm concerned about supporting every package manager out there.
Most common now are deb, rpm and pkg standard packages. Now we have [snap and Flatpak](http://arstechnica.com/information-technology/2016/06/here-comes-flatpak-a-competitor-to-ubuntus-cross-platform-linux-apps/) and possibly other competing standards to try to unify them.
I am not sure how the features of snap you mentioned address any of the issues our users have raised in using the cf CLI, or are any better than what we have now.
I think we'll wait and see which packaging system wins.

Thanks for bringing this up though, certainly something to keep an eye out for.
Cheers,
Dies Koper
CF CLI PM

elopio:
Hey @dkoper. I can take care of maintaining the snap and organizing the testing with the Ubuntu community to get new releases to the stable channel. I was already planning for it. So it would be almost* zero effort for you and after a trial period we can take a look at the store stats to see how many users are getting benefits from the snap.

Of course, I understand your concerns, and I respect if you prefer to wait before upstreaming this file.

Thanks.

* almost, because you will have me around asking questions :stuck_out_tongue_winking_eye:

dkoper:
Hi @elopio 

I assume you can maintain the snap on a (regularly sync'ed) fork and proceed your experimenting without us incorporating this file. We prefer to wait until snap becomes commonly used and highly requested.
Note that when we are ready to take your PR, we need a [CLA on file](https://github.com/cloudfoundry/cli/blob/master/CONTRIBUTING.md).

Thanks again for sharing!
Dies Koper
CF CLI PM

elopio:
Yes, I can.
I've made one branch to follow master [1] and one to follow the release tags [2]. They are set up with CD scripts so when there are changes, they are pushed to launchpad in order to get snaps build for all the architectures [3]. Then the master snaps are pushed to the edge channel, and the tags to the beta channel.

People can give it a try with:

`$ sudo snap install cf --edge`

or

`$ sudo snap install cf --beta`

Whenever you want I can make the transfer or give you access to the delivery system and the store. And I will keep the testing status updated in https://gist.github.com/elopio/061584d5c5f4e3297104f34dec96d906

About the CLA, that's done :)

Thank you!

[1] https://github.com/elopio/cli/tree/snapcraft
[2] https://github.com/elopio/cli/tree/snapcraft-beta
[3] https://code.launchpad.net/~cloudfoundry-snappy/+snap/snapcraft-edge

dustinkirkland:
Once upon a time, I was responsible for the deb packaging of the cf CLI (and the "vmc" CLI before that, if anyone remembers that!) for Ubuntu.

The pace of development and change of the cf CLI unfortunately has never aligned well with Ubuntu's 6-month release schedule and certainly not the 2-year LTS cycle.  Inevitably, the cf CLI tool would always be out of date in Ubuntu's deb package archives.

The real advantage here of a cf snap is that the CloudFoundry community could publish/upload new cf releases AT ANY TIME!  That has never before been possible with .deb's in the Ubuntu archive.  Any Ubuntu 16.04 LTS user -- at any time -- could 'sudo snap install cf', and get the exact version of cf that the CloudFoundry community has released.  The default channel would be "stable", but your Ubuntu users could just as easily snap install from "rc", "beta", or "devel" channels.

Amazingly, what Leo has provided here in this PR is the yaml packaging itself for cf :-)  That's pretty awesome.  With this little bit of yaml, the CloudFoundry community could easily integrate a snap build and upload in your BOSH CI/CD build system, bringing the cf CLI *straight* into Ubuntu 16.04 LTS whenever you want...

chipchilders:
@dkoper I think that the comment from @dustinkirkland above is certainly worth considering. We want to be sure that the CLI project can keep moving fast, and at the same time fit into the Ubuntu ecosystem in the best way possible. With a commitment from @elopio to maintain it going forward, I'd vote that you reconsider and accept (pending CLA of course). cc/ @dieucao

elopio:
of course, I'm here to help. I signed the CLA some time ago.

PR #929, opened 2016-08-16
Implement rune replacement in 'cf logs'

jasonkeene:
Here is the relevant story in our backlog:

https://www.pivotaltracker.com/story/show/128174339

To summarize, java developers sometimes output their logs with a given unicode code point instead of `\n` (typically `\u2028`) to avoid logs that span multiple lines getting split up before being sent to loggregator. They want to be able to reverse this when doing `cf logs`.  To ensure this is opt-in rather than affecting all apps, we've implemented it in the CLI as a flag.


dkoper:
Hi @jasonkeene 

This looks like a hack. The user wants to see their stacktrace in one message. I'm not sure where the stacktrace is broken up and why, but that seems to be the place to address this. Users should not have to use a flag to display logs correctly - in particular not a flag with some secret value they need to obtain from the app developer OOB.

Regards,
Dies Koper
CF CLI PM

dkoper:
Feel free to open an issue, send us an email or schedule a meeting with me if you have any questions.

jasonkeene:
@dkoper I was under the impression our PM at the time had talked with you about this. Diego splits messages up before they get to loggregator. We don't want to impose this expensive transformation on any part of our system since most users won't want this and will add latency. Pushing it to the edge and putting it under a flag seemed like the best place for this *if* we do it. That said it is a hack. I'll talk with the team.

dkoper:
Thanks. Your PM hadn't spoken to me about this.

PR #931, opened 2016-08-17
fixing links

shalako:


dkoper:
I don't see any other CF github readme's use https to the cloudfoundry.org domain. Why bother?
(I am a big proponent for using https when up/downloading binaries, or loading pages with download links)

dkoper:
Thanks Shannon!

shalako:
Issue was that without any scheme at all, clicking links in GitHub sent me to a 404 github page: http://github.com/cloudfoundry/cli/docs.cloudfoundry.org

PR #947, opened 2016-09-01
Update install instructions for binaries

Rory80Hz:
During user testing for installing the CLI we discovered that people had issues using the
curl based install. The binary is not on the path, and the instructions generally in the documentation assume it is. For example cf --version rather than ./cf --version.

By suggesting moving to somewhere in your path like a common location like /usr/local/bin the confusion can be reduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,16 +47,28 @@ Check out our [community contributed CLI plugins](https://plugins.cloudfoundry.o
 $ curl -L "https://cli.run.pivotal.io/stable?release=macosx64-binary&source=github" | tar -zx
 # ...or Linux 64-bit binary
 $ curl -L "https://cli.run.pivotal.io/stable?release=linux64-binary&source=github" | tar -zx
-# ...and to confirm your cf CLI version
-$ ./cf --version
-cf version x.y.z-...
 ```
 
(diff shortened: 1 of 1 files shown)


dkoper:
Hi @Rory80Hz 

Thanks, this is great feedback.
If your testers have access to /usr/local/bin (i.e. root access), why don't they just use the deb/rpm installers?

Rory80Hz:
We found people used different options, a number got caught out by curl (most were using OSX rather than linux) and didn't opt for the homebrew option.

They could work around it easily enough (some knew to move it) but it feels like the docs could be a little more explicit.

dkoper:
Thanks for the details.
We're close to supporting `apt-get` and `yum`; a good opportunity to rearrange the section and make the installation with package managers more prominent. I'd like to address this issue as part of that.

Rory80Hz:
Excellent, thanks @dkoper

dkoper:
Hi @Rory80Hz 

We've added support for `apt-get` and `yum` package management support and have updated the readme accordling, including your feedback. Please check it out.

dkoper:
Closing, I believe this is addressed now.

Rory80Hz:
Thanks @dkoper great improvements, thanks!

PR #959, opened 2016-09-14
improve descriptions for restart and restage

mastertinner:
During a usability testing, I found out that the difference between restage and restart is not clear to many new users.

### DIFF
--- a/cf/commands/application/restart.go
+++ b/cf/commands/application/restart.go
@@ -35,7 +35,7 @@ func (cmd *Restart) MetaData() commandregistry.CommandMetadata {
 	return commandregistry.CommandMetadata{
 		Name:        "restart",
 		ShortName:   "rs",
-		Description: T("Stop and restart the app within the existing droplet"),
+		Description: T("Restart the app within its existing container"),
 		Usage: []string{
 			T("CF_NAME restart APP_NAME"),
 		},


mastertinner:
This is a follow up issue to https://github.com/cloudfoundry/docs-cf-cli/pull/19

dkoper:
Hi @mastertinner 

We welcome improvements, but your suggestions introduce terminology that I feel is internal to CF, or at least is not widely used and understood.
A droplet to me is what floats through the air when you sneeze and makes people sick, no matter how many times you point me to a different definition.
This is why we're not using `droplet` and application `bits` in any CLI help pages. (This may change in the future as we look at exposing new V3 API functionality, but we'll try to avoid it as much as possible.)

As for  submitting a contribution that updates messages, refer to [this section](https://github.com/cloudfoundry/cli/blob/master/CONTRIBUTING.md#i18n) for instructions on how to deal with our translations.

Regards,
Dies Koper
CF CLI PM

mastertinner:
Thanks for your quick reply, @dkoper! I adjusted the descriptions to omit these words. Can you please check my updated descriptions? In case they are still not correct, do you have any idea how we could make the difference between restarting and restaging more clear to new users?

dkoper:
"Container"? Another new term...
I definitely agree there is room for improvement, haven't been able to come up with something good yet. Let's keep trying though!

dkoper:
@mastertinner How about the following?
```
   restage - Rebuild the app using the latest pushed application files and the latest environment (variables, service bindings, stack, etc.)
```

dkoper:
@SocalNick Can you review this improved description for the `cf restage` command?
```
   restage - Recreate the app's executable artefact using the latest pushed app files and the latest environment (variables, service bindings, stack, etc.)
```

SocalNick:
Sounds good, probably want to say latest buildpack as well.

mastertinner:
So shall we go with the text below?

```text
restage - Recreate the app's executable artefact using the latest pushed app files and the latest environment (buildpack, variables, service bindings, stack, etc.)
```

dkoper:
@mastertinner We have already implemented this as follows:
```
   restage - Recreate the app's executable artifact using the latest pushed app files and the latest environment (variables, service bindings, buildpack, stack, etc.)

   restart - Stop all instances of the app, then start them again. This may cause downtime.
```
How does that sound?

mastertinner:
Perfect, thanks! So can I close this PR?

dkoper:
Thanks again for bringing up the issue with these command descriptions and working with us to make them better.
They have been incorporated in the just published 6.22.2 release:
https://github.com/cloudfoundry/cli#downloads

Cheers,
Dies Koper
CF CLI PM

mastertinner:
I saw :) thanks a lot for your support, @dkoper and @SocalNick!

PR #963, opened 2016-09-20
Fix: cf update-service when current service plan is private

markstgodard:
In cases where we are updating a service instance whose current plan is private, we will need to look up the service guid. This PR is to change cf cli to use `/v2/spaces/:guid/summary` and match the existing service instance name and determine the associated service offering guid.

Basically /v2/spaces/space-guid/service_instances API was changed awhile back to not allow a space developer to see the service plan info, and hence the current code gets the service_guid via the current (inactive) plan's service_guid.

See original PR for background and context: https://github.com/cloudfoundry/cli/pull/831


dkoper:
@markstgodard @SocalNick Can you confirm the new API being called (and its fields the command will now depend on) have been around for a long time (like, since CF 212 (over 1 year ago)?

SocalNick:
@dkoper I think it goes back as far as the documentation goes: http://apidocs.cloudfoundry.org/192/spaces/get_space_summary.html

dkoper:
Closing this PR as we're not going to merge it (we're moving away from using summary endpoints due to issues with >50 entries) but leaving the story in the tracker to ensure we revisit the issue it intended to address when we refactor this command.

PR #966, opened 2016-09-28
Use noaa token refresher and retry logic for logs

iplay88keys:
[#130990227]

Signed-off-by: Andrew Poydence <apoydence@pivotal.io>

### DIFF
--- a/cf/api/logs/logsfakes/fake_noaa_consumer.go
+++ b/cf/api/logs/logsfakes/fake_noaa_consumer.go
@@ -5,17 +5,18 @@ import (
 	"sync"
 
 	"code.cloudfoundry.org/cli/cf/api/logs"
+	"github.com/cloudfoundry/noaa/consumer"
 	"github.com/cloudfoundry/sonde-go/events"
 )
 
 type FakeNoaaConsumer struct {
-	TailingLogsWithoutReconnectStub        func(string, string) (<-chan *events.LogMessage, <-chan error)
-	tailingLogsWithoutReconnectMutex       sync.RWMutex
-	tailingLogsWithoutReconnectArgsForCall []struct {
+	TailingLogsStub        func(string, string) (<-chan *events.LogMessage, <-chan error)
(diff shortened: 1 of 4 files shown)


jeffpak:
Hey! I'm assuming that this PR has to do with this [story](https://www.pivotaltracker.com/story/show/131316887) but there doesn't seem to be much more detail regarding it. We're wondering if you could give us some context regarding what this PR accomplishes, since it seems like only the tests, some fakes, \ and a couple methods are being changed on our side. I'm assuming the rest is already implemented on the loggregator side?

Cheers!
@jeffpak & @mmb, CF CLI Team

iplay88keys:
@jeffpak @mmb,

Currently, the cf cli uses `TailingLogsWithoutReconnect()` on the NOAA consumer with the token refresher logic being called in the [cli connection code](https://github.com/cloudfoundry/cli/blob/37002dd188428d84c17b80ffb31455c57616b4d5/cf/api/logs/noaa_logs_repository.go#L91-L95). This pull request switches over to using the `TailingLogs()` method which, when the token refresher is set on the NOAA consumer object, runs the refresh [logic](https://github.com/cloudfoundry/noaa/blob/c2c343ac099f2de5915968ff241e4c780f9758b8/consumer/sync.go#L118) within the NOAA consumer as well as some minimal reconnect logic.

Thanks!
Jeremy, Loggregator Team

jeffpak:
So does that mean that if the token refresher is not set on the NOAA consumer, the functionality of `TailingLogs()` will mimic `TailingLogsWithoutReconnect()`.

iplay88keys:
They both will use the token refresher if it is set. The only difference between the two is that if `TailingLogsWithoutReconnect()` is called, it does not try to reconnect. `TailingLogs()` will try 5 times to reconnect if it gets an error the first time.

jeffpak:
This commit should be fine, so we went ahead and merged it. Thanks for submitting the PR!

iplay88keys:
Thanks for merging it!

PR #974, opened 2016-10-06
TailingLogs does not exit when noaa returns RetryError

iplay88keys:
[#130990227]

Signed-off-by: Andrew Poydence <apoydence@pivotal.io>


apoydence:
Please note that this will make `cf logs` block indefinitely trying to reconnect to the Loggregator system. The code writes info to the `DebugPrinter` and therefore the user will only see information (if the CLI is not connecting to Loggregator properly) in verbose mode.

If the user SHOULD see more information than that while not in the verbose mode, then that should be added before merging this in.

nelsam:
We've updated this PR to allow a timeout to be passed to the NOAA logs repository.  We're resetting a timer with the passed in timeout when a RetryError is received on the errors channel, and stopping it in the callback passed in to noaa's `OnConnectCallback(func())` method.  We've passed the timeout in as 15 seconds in the production code, but passed 1.5 seconds (we used 1.5 seconds so as to avoid disrupting any existing `Eventually`/`Consistently` calls) in the tests.

We had to modify one test that was asserting on the pointer to the callback passed to the fake NOAA consumer; now, it just asserts that the callback is called when the function passed to the fake NOAA consumer is called.

Issue #994, opened 2016-11-03
Small language issue when confirming installation of a plugin


lbayerl:
**cf version 6.22.2**+a95e24c-2016-10-27
Kubuntu 14.04 LTS

Command:
**cf install-plugin https://example.com/some-plugin**
with a German language version of cf (don't know how I got this) asks me to confirm whether I really want to install the plugin:

"**Achtung: Plug-ins werden als Binärdateien von möglicherweise nicht vertrauenswürdigen Autoren geschrieben. Sie installieren und verwenden Plug-ins auf eigenes Risiko.**

Möchten Sie das Plug-in https://swisscom-plugin.scapp.io/linux64/swisscom-plugin installieren? (J oder N)"

**(J oder N)** is the same as the English **(Y or N)**, but in this case J won't be accepted and the error message is:

"FEHLGESCHLAGEN
Plug-in-Installation abgebrochen" (which means FAILED - plugin installation cancelled).

One has to know that you'll have to confirm with Y but with J.

dkoper:
Hi @lbayerl 

It seems this is the only confirmation prompt where we displayed the `Y or N` and German was one of the p[;few locales where the translators did not leave that intact.

We're migrating commands to a new library that does the printing of [y/n] for us. I prefer to not localise this: it gets too confusing in some locales (like where they have no single-letter abbeviations on a QWERTY keyboard, like Japanese).

In cf CLI 6.23.1, this command no longer displays the `J oder N`, consistent with other confirmation prompts. In a future release, you'll see the `[y/n]` added.
https://github.com/cloudfoundry/cli#downloads

Thanks for your report!

Regards,
Dies Koper
CF CLI PM

PR #999, opened 2016-11-11
Fix prompting users in locales other than en-US

stefanlay:
terminal.UI.Confirm retured true when the user's entry after
transformation to lower case matched either "y", "yes" or the
translation of T("yes").

This did not work for most of the translated prompts.

E.g. "Do you want to install the plugin {{.Plugin}}? (y or n)"
is translated to
"... installieren? (J oder N)"

This only works if "y" is translated and terminal.UI.Confirm also
checks against T("y").

### DIFF
--- a/cf/i18n/resources/de-de.all.json
+++ b/cf/i18n/resources/de-de.all.json
@@ -6396,6 +6396,10 @@
     "translation": "Version"
   },
   {
+    "id": "y",
+    "translation": "j"
+  },
+  {
     "id": "yes",
     "translation": "Ja"
   },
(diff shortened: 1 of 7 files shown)


lbayerl:
#994 - one of those issues, can be probably closed when this PR is merged.

dkoper:
Hi @stefanlay 

We're migrating to another library that adds a (non-localised) `[y/n]` to our confirmation prompts.
Rather than internationalising them, I care more about making them consistent among all our commands that offer prompts.
That would also ensure users of different locales won't need to relearn what button to press to confirm, and won't need to think about it when switching locales.

I believe this is the only command, and German one of the few locales, that translated the values.
We'll take them out in the next release, and you should see the new `[y/n]` prompts slowly spread to other commands as we refactor them.

Cheers,
Dies Koper
CF CLI PM

PR #1005, opened 2016-11-21
don't order the first page of api returns in "cf orgs"

pusherofbrooms:
If there is more than one page of api results, "cf orgs" seems to order the first page and not order the rest of the pages. This will cause some orgs to be missing in the list and some orgs to be duplicated. Either order all of the results, or don't order the first page. This patch opts for the latter.

### DIFF
--- a/cf/api/organizations/organizations.go
+++ b/cf/api/organizations/organizations.go
@@ -40,7 +40,6 @@ func (repo CloudControllerOrganizationRepository) ListOrgs(limit int) ([]models.
 	orgs := []models.Organization{}
 	err := repo.gateway.ListPaginatedResources(
 		repo.config.APIEndpoint(),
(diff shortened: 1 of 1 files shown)


pusherofbrooms:
As the diff is 100% removal of code, I'm not sure a CLA is necessary for this PR. I will look into it when I have the time though.

dkoper:
Hi @pusherofbrooms 

Thanks for pointing out this issue.
Sorting of orgs (and spaces) is intentional, introduced in cf CLI 6.18.1. Clearly we never tested it with >50 entries in the result.
I'll ask my team to take a look - undoing the sort is not what we want.

Cheers,
Dies Koper
CF CLI PM

XenoPhex:
Looks like https://github.com/cloudfoundry/cf-release/issues/1121 fixes the ordering issue.

PR #1025, opened 2016-12-08
Clarify timeout option for push command

PeterWippermann:


dkoper:
Hi @PeterWippermann 

Thanks for this suggestion!
We're not allowed to accept contributions without a signed CLA (and your GH profile does not list a membership to an organization covered by a CLA) and your change changes a message id breaking translations, but we're looking at improving this description soon anyway, in particular clarifying the relation between this flag and the timeout related env vars, so I'll leave it open and ensure your feedback of including the app health aspect is considered. Thanks!

PeterWippermann:
Hi @dkoper 
Thanks for your response. It's nice to not only having to talk to bots ;-)
Sending the CLA back to you took me a while and I just did so. But as far as I understand, you're unable to accept this PR atm, right? Would be ok to for to close this PR than as long as you track me intent in your internal system.

dkoper:
Closing as suggested. We'll keep the story linked above open to track your suggestion - this PR will be automatically updated (or I'll do it manually if not) when it's been addressed.

dkoper:
@PeterWippermann In case you hadn't noticed, we recently improved the `push -t` description to:
```
   -t   Time (in seconds) allowed to elapse between starting up an app and the first healthy response from the app
```

Thanks again for submitting your suggestion!

PeterWippermann:
Thanks for keeping me updated @dkoper ! I'm glad, I could help :-)
@cf-gitbot shouldn't you remove the "in progress" label as well?...

dkoper:
It has been removed, hasn't it?
I must say the `cf-gitbot` message is confusing. It seem it added and immediately removed the label again.

PeterWippermann:
Oh you're right, sorry! :-)

PR #1040, opened 2016-12-30
clean up the issue template

afeld:
This moves the instructions to the submitter to be comments, so that they don't accidentally get submitted with the issue. Also changed each of the sections to have an actual heading.


dkoper:
Much appreciated!
The untidiness our issue template had been inviting has been bothering me for some time :)

PR #1090, opened 2017-03-13
Corrected a typo in the command line, `cf config -locale` to `cf conf…

mnitta:
…ig --locale`

## What Need Does It Address?

This correction addresses document correctnes.
## Possible Drawbacks

None

Because this PR corrects README-i18n.md, not code itself.
## Description of the Change
In the first paragraph of README-i18n.md, the command line example contains a typo.
It should be cf config --locale, instead of cf config -locale.


dkoper:
Thanks Minoru!

PR #1094, opened 2017-03-14
installation instructions should be copy/pasteable

jfmyers9:
I've found it difficult to follow the bash instructions in the README for installation on various platforms due to the `$`. It would be awesome if instructions that are meant to be copy/pasted did not include this character. This seems to be a pattern on other projects.

I left the `$` on examples as they do a good job indicating a difference between the command and the output of the command.


dkoper:
Thanks Jim!

We're looking at replacing the example section with an animated GIF, so if it looks inconsistent now, it won't be for long.

Issue #1130, opened 2017-05-03
Misleading error message when `cf ssh` is used with too-large `-i` value


davwards:
## Command  

Given I've deployed `my-app` with 2 instances:

`cf ssh my-app -i 2`

## What occurred

Command fails with:

"Error opening SSH connection: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain"

## What you expected to occur

An error message letting me know that I referenced an app instance that doesn't exist

## CLI Version  

`cf version 6.26.0+9291ea37e.2017-05-02`

## CC API Endpoint Version  

`api version:    2.80.0`

dkoper:
I believe what adds to the confusion is that the help text doesn't specify what the `-i` value defaults to, which would clarify the index starts at 0, not 1. We'll look at improving that after reviewing your PR.

dkoper:
Resolved in 6.27.0. Thanks for your contribution.
https://github.com/cloudfoundry/cli#downloads

PR #1131, opened 2017-05-03
Validate -i value for ssh command is an existing app instance

davwards:
## What Need Does It Address?

Without this change, passing a -i value that is equal to or greater than
the total number of instances the application has results in a cryptic
error message about the ssl handshake, which can lead users down a false
trail.

## Possible Drawbacks

To keep the validation checks for the `-i` flag together, I moved the not-negative validation check **after** the creation of `cmd.opts` and `cmd.appReq`. No tests were upset by this, but it could be the case that due to factors I didn't catch--if populating `cmd.appReq` is expensive, for instance--this could result in slower feedback when you enter a negative value for `-i`.

Also, I didn't add any i18n translations for the new message. **Note:** When trying to run `bin/i18n-checkup`, I observed that the schema of the translation `.json` files was completely changed. On Slack, `@anand` indicated that this was probably due to the fact that the team is using a particular version of `i18n4go` that is different than the one I got with `go get`, and that I should not make any changes to the translation files and submit anyway.

It improves the messaging of a core feature, as opposed to adding new functionality.

## Description of the Change

In addition to checking that the `-i` value is non-negative, fetch the InstanceCount of the specified application and ensure the `-i` value is less than that. If not, fail with a descriptive message.

## Applicable Issues

https://github.com/cloudfoundry/cli/issues/1130

### DIFF
--- a/cf/commands/application/ssh.go
+++ b/cf/commands/application/ssh.go
@@ -67,11 +67,6 @@ func (cmd *SSH) Requirements(requirementsFactory requirements.Factory, fc flags.
 		return nil, fmt.Errorf("Incorrect usage: %d arguments of %d required", len(fc.Args()), 1)
 	}
 
-	if fc.IsSet("i") && fc.Int("i") < 0 {
-		cmd.ui.Failed(fmt.Sprintf(T("Incorrect Usage:")+" %s\n\n%s", T("Value for flag 'app-instance-index' cannot be negative"), commandregistry.Commands.CommandUsage("ssh")))
-		return nil, fmt.Errorf("Incorrect usage: app-instance-index cannot be negative")
-	}
-
 	var err error
 	cmd.opts, err = options.NewSSHOptions(fc)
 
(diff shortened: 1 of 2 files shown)


nickwei84:
@davwards When I have one running instance of the app "my-app", the following no longer works:
```
$ cf ssh my-app -i 0
FAILED
Incorrect Usage: The specified application instance does not exist

NAME:
   ssh - SSH to an application container instance
...
```

The `Requirements` method is meant for setup so that the `Execute` method can perform the core logic of the ssh command. What you have could work, if you did a `cmd.appReq.Execute()` before performing the check. This will make the API call to get the application object. Without it, calling `cmd.appReq.GetApplication()` returns an empty application model with instance count defaulting to 0, thus the scenario above.

However, the calling function that runs `Requirements` and `Execute` is already responsible for running `cmd.appReq.Execute()`. It would be much cleaner to move this check in the `Execute` method.

davwards:
I see! That suggests that the way I simulated the situation in `ssh_test` wasn't true to life; how would I do that properly?

nickwei84:
@davwards At the layer that was being unit-ish tested, you were doing the right thing. The correct way to expose this bug would be through integration tests. You can add a test for this [here](https://github.com/cloudfoundry/cli/blob/master/integration/isolated/ssh_command_test.go), and you can use [this as an example](https://github.com/cloudfoundry/cli/blob/master/integration/isolated/set_health_check_command_test.go).

davwards:
Hi all; I'm getting pulled off the Labs beach, so my available time to work on this is diminishing. Getting the bosh-light integration test environment set up is proving pretty difficult. All this to say, I don't expect to be able to deliver an integration-test-driven fix soon.

@nickwei84, I also feel like my understanding of the test setup is incomplete; it seems odd to me that the unit test I wrote could be correct, given that the implementation was completely broken. Is there not a way to simulate the life-like behavior of `cmd.appReq.Execute()` and `cmd.appReq.GetApplication()`? (In particular, the fact that `GetApplication()` returns an empty model until `Execute()` has been called.)

nickwei84:
@davwards I'd recommend getting a PWS account and test against that.
The unit tests for the `Requirements()` and `Execute()` methods intrinsically have different assumptions. The calling function in `cf/cmd/cmd.go` should tell you what these assumptions are [link to code chunk](https://github.com/cloudfoundry/cli/blob/master/cf/cmd/cmd.go#L107-L124). Following the order that things are called, stubbing out `GetApplications()` in the `Execute()` method would have been the right thing to do.

dkoper:
@davwards Please let us know if you think you won't be able to get to this any time soon and would like us to finish the job.

davwards:
@dkoper Indeed, feel free to wrap this up; I don't want it to get stuck in limbo waiting on me.

pivotal-stuart-pollock:
resolved the issues which arose in an initial review

PR #1145, opened 2017-05-19
Add create-org & delete-org to Org management help

ronakbanka:
## What Need Does It Address?

As of now when user do `cf help` , `create-space` is visible in default view, but `create-org` is not visible in default view , which sometimes confuses users about basic org management functionality. 

screen shot 2017-05-19 at 10 29 58 am

## Possible Drawbacks

Just a change in default view so there shouldn't be any drawbacks.

Improvement in default options for Org management

## Description of the Change

This PR adds two options `create-org` & `delete-org` to Org management default view , which are now in sync with space management.

screen shot 2017-05-19 at 10 30 20 am

### DIFF
--- a/command/common/help_command_test.go
+++ b/command/common/help_command_test.go
@@ -405,7 +405,7 @@ var _ = Describe("help Command", func() {
 			Expect(testUI.Out).To(Say("  spaces\\s+create-space\\s+set-space-role"))
 
 			Expect(testUI.Out).To(Say("Org management:"))
-			Expect(testUI.Out).To(Say("  orgs,o\\s+set-org-role"))
+			Expect(testUI.Out).To(Say("  orgs,o\\s+create-org,co\\s+set-org-role"))
 
 			Expect(testUI.Out).To(Say("CLI plugin management:"))
 			Expect(testUI.Out).To(Say("  install-plugin    list-plugin-repos"))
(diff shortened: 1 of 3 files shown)


dkoper:
The purpose of the `cf help` page is to provide a quick overview of the most common cf CLI commands for app developers.
`cf create-org` and `cf delete-org` do not belong there.
Also, these commands can only be executed by admins (unless a feature flag is enabled, but it's disabled by default).

I'd be interested to hear about your use case. I see `create-org` as an advanced command that an advanced user like an admin runs at least once after deploying CF. Those users would be more interested in `cf help -a`.

Regards,
Dies Koper
CF CLI PM

ronakbanka:
@dkoper Agree on admin part , but for eg. in case of bosh-lite deployment of cf , when someone is getting started , it might confuse the users about missing command line option . 

This is what i observed while working with some team members , who were new to environment .

dkoper:
@ronakbanka What instructions did they follow to do the bosh-lite deployment of cf? Maybe those instructions could be updated to run `cf help -a` to see what's available.

dkoper:
Closing. As per my above comments, `cf help` is for most common cf CLI commands for app developers. If we start making exceptions, we'll find ourselves back to where we were with an unwieldy multi-page list of commands flying over our screens.
If we can improve the OSS bosh-lite docs to prevent confusion among new users, please point me to the relevant articles we can review them together. Thanks!

PR #1149, opened 2017-05-23
Add quotes around org in the create-org command TIP

vlad-stoian:
- The tip can be copy-pasted for orgs that have a space in the name
- The tip is now consistent with the create-space tip which already
contains quotes

Ran the ginkgo tests, and also manually compiled and created an org containing spaces, copy-pasted the tip, and it works.

### DIFF
--- a/cf/commands/organization/create_org.go
+++ b/cf/commands/organization/create_org.go
@@ -137,6 +137,6 @@ func (cmd *CreateOrg) Execute(c flags.FlagContext) error {
 	}
 
 	cmd.ui.Say(T("\nTIP: Use '{{.Command}}' to target new org",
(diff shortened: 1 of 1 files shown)


dkoper:
Hi @vlad-stoian 

Thanks! That looks like a useful change.
Can you please take a look at the comment from reddbot above so we can review & merge?

vlad-stoian:
Hi @dkoper,

Sorry, didn't notice my membership of @pivotal-cf was hidden. I made it public and according to what reddbot says that should be sufficient.

Hope everything is ok now. If not, I'll try again. :D 

Have a great day.

dkoper:
Not sure why GitHub didn't mark it as merged, but our commit history includes your commit, and it's part of the 6.27.0 release. Thanks again!

Issue #1175, opened 2017-06-30
`/oauth/token` with refresh token does not set authorization header


aeijdenberg:
Recent versions of the CLI seem to fail against older versions of the API for some operations, specifically those that require getting a new access token for a refresh token.

This appears to be fixed by supplying an `Authorization` header per <https://tools.ietf.org/html/rfc6749#section-2.3.1>.

## Command  

To demonstrate:

```
cf logs --recent  APPNAME
```

## What occurred

```
An Authentication object was not found in the SecurityContext
```

## What you expected to occur

Recent logs for application to be shown

## CLI Version  

```
cf version 6.27.0+d26b32dcc.2017-06-08
```

## CC API Endpoint Version  

```
api version:    2.43.0
```

## CF Trace

```
REQUEST: [2017-06-30T11:17:34+10:00]
POST /oauth/token HTTP/1.1
Host: uaa.system.[hidden]
Accept: application/json
Connection: close
Content-Type: application/x-www-form-urlencoded
User-Agent: cf/6.27.0+d26b32dcc.2017-06-08 (go1.8.3; amd64 darwin)
[PRIVATE DATA HIDDEN]

RESPONSE: [2017-06-30T11:17:34+10:00]
HTTP/1.1 401 Unauthorized
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Cache-Control: no-store
Content-Length: 108
Content-Type: application/json
Date: Fri, 30 Jun 2017 01:17:34 GMT
Expires: 0
Pragma: no-cache
Pragma: no-cache
Server: Apache-Coyote/1.1
Strict-Transport-Security: max-age=31536000 ; includeSubDomains
Www-Authenticate: Basic realm="UAA/client", error="unauthorized", error_description="An Authentication object was not found in the SecurityContext"
X-Cf-Requestid: d29d3687-8fb5-4b8f-76aa-d3f8da0cae03
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-Xss-Protection: 1; mode=block
{
  "error": "unauthorized",
  "error_description": "An Authentication object was not found in the SecurityContext"
}


An Authentication object was not found in the SecurityContext
```

## Platform & Shell Details  

Not relevant

## Any other relevant information  

The same cli tool works fine against another one of our environments that has a newer API:

```
api version:    2.75.0
```

dkoper:
@aeijdenberg Would you mind downloading an edge release and confirm the issue is resolved?
[...]

PR #1176, opened 2017-06-30
Always add authorization header when requesting new tokens

aeijdenberg:
Appears to be required for older versions of API server. Fixes issue 1175.

## What Need Does It Address?

Fixes https://github.com/cloudfoundry/cli/issues/1175

### DIFF
--- a/api/uaa/refresh_token.go
+++ b/api/uaa/refresh_token.go
@@ -1,6 +1,7 @@
 package uaa
 
 import (
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


dkoper:
Hi @aeijdenberg 

Thanks for submitting the issue and a fix. Note that we cannot accept any code contributions without a CLA on file, and our tooling didn't find any (see comment above).
I have asked my team to investigate the issue.

aeijdenberg:
@dkoper - understood. In parallel I'm investigating how to get the CLA issue dealt with on our end as well.

api/uaa/refresh_token.go
```diff
@@ -36,6 +51,7 @@ func (client *Client) RefreshAccessToken(refreshToken string) (RefreshToken, err
 			"Content-Type": {"application/x-www-form-urlencoded"},
 		},
 		Body: body,
+		AddClientAuthorization: true,
 	})
```
mariash: Is there a reason why you did not just add `SetBasicAuth` on a request here? It would be nice to have a single way of setting the auth header.

Also, would be nice to have a test for this in `refresh_token_test.go`.

aeijdenberg: `request.SetBasicAuth()` does not actually do what [section 2.3.1 of the OAuth 2.0 spec](https://tools.ietf.org/html/rfc6749#section-2.3.1) requires. ie. the client ID and secret must be URL encoded, before being constructed into a basic auth header, which is why I added the `oauthClientAuthorizationHeader()` method.

mariash: @aeijdenberg according to golang comment here these are not meant to be url encoded https://github.com/golang/go/blob/master/src/net/http/client.go#L335-L343

mariash:
@aeijdenberg thank you for submitting the PR. We used your idea in our commit https://github.com/cloudfoundry/cli/commit/1539ecc4e1783964458a6bbb94440dc4e331ac53 We decided to use native golang method and also added a unit test.

api/uaa/refresh_token.go
aeijdenberg: @mariash - I think you're confusing the HTTP specification with the OAuth specification. You are correct that [RFC2617: HTTP Authentication: Basic and Digest Access Authentication](https://www.rfc-editor.org/rfc/rfc2617.txt) states to simply concatenate with no query escaping, and that is what the Golang HTTP library is correctly doing, since it is an HTTP library, not an OAuth library.

However the [section 2.3.1 of OAuth 2.0 spec](https://tools.ietf.org/html/rfc6749#section-2.3.1) states:

> The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password.

This means that both the client ID and secret need to be URL encoded before calculating the Basic auth header.

aeijdenberg: I will note that the UAA server would appear to have the same bug: https://github.com/cloudfoundry/uaa/blob/fe946fdc871027fcea9278e97c09cbf270831cf4/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientBasicAuthenticationFilter.java#L82

PR #1209, opened 2017-08-21
Add network commands

chinangela:
These changes add network policy commands `allow-network-access`, `list-network-access`, and `remove-network-access` to the CLI.

[#145552365], [#148703351], [#148703369]

Let us know if you have any questions/concerns.

Thanks,
Angela and @joachimvaldez


dkoper:
Thanks!
I've move this to the top of our backlog - the team will start reviewing tomorrow.
Also, I have added my expected command help & output & A/C in the stories in the container networking tracker.
I will start acceptance after this code's been merged. If you can take a look at my comments on the stories and apply any changes that may be required to this PR, that will make acceptance go faster.

api/cfnetworking/cfnetv1/policy_test.go
```diff
@@ -114,4 +111,188 @@ var _ = Describe("Policy", func() {
...
+								Start: 1234,
+								End:   1235,
+							},
+						},
+					},
+				}
+			})
+
+			It("returns the policies correctly", func() {
+				policies, err := client.ListPolicies("")
```
XenoPhex: This and line 193 would no longer have to pass empty string to `ListPolicies` when it becomes variadic.

command/common/command_list.go
```diff
@@ -36,6 +36,8 @@ type commandList struct {
 	AllowSpaceSSH                      v2.AllowSpaceSSHCommand                      `command:"allow-space-ssh" description:"Allow SSH access for the space"`
 	Api                                v2.ApiCommand                                `command:"api" description:"Set or view target api url"`
 	Apps                               v2.AppsCommand                               `command:"apps" alias:"a" description:"List all apps in the target space"`
+	AllowNetworkAccess                 v3.AllowNetworkAccessCommand                 `command:"allow-network-access" description:"Allow direct network traffic from one app to another"`
+	RemoveNetworkAccess                v3.RemoveNetworkAccessCommand                `command:"remove-network-access" description:"Remove direct network traffic from one app to another"`
```
XenoPhex: Outside of the top V3/experimental section. We try and keep the command list in alphabetical order.

command/v3/list_network_access_command.go
```diff
@@ -0,0 +1,95 @@
...
+
+	user, err := cmd.Config.CurrentUser()
+	if err != nil {
+		return err
+	}
+	cmd.UI.DisplayTextWithFlavor("Listing network traffic as {{.User}}...", map[string]interface{}{
+		"User": user.Name,
+	})
+
+	policies, warnings, err := cmd.Actor.ListNetworkAccess(cmd.Config.TargetedSpace().GUID, cmd.SourceApp)
```
XenoPhex: There should be a switch here that checks if `cmd.SourceApp` is empty, it calls `ListNetworkAccessBySpace`, otherwise it calls  `ListNetworkAccessBySpaceAndAppName`.

chinangela:
We believe now all changes on our end have been done. Let us know if there are any other issues.

Cheers,
Angela and @joachimvaldez

integration/isolated/add_network_policy_command_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"code.cloudfoundry.org/cli/integration/helpers"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = FDescribe("add-network-policy command", func() {
```
XenoPhex: Remove Fdescribe

integration/isolated/network_policies_command_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"code.cloudfoundry.org/cli/integration/helpers"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = FDescribe("network-policies command", func() {
```
XenoPhex: Remove FDescribe

chinangela:
Whoops, our bad. Focuses have been removed.

Cheers,
Angela

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,184 @@
...
+	for _, v1Policy := range v1Policies {
+		if v1Policy == policyToRemove {
+			return allWarnings, actor.NetworkingClient.RemovePolicies([]cfnetv1.Policy{policyToRemove})
+		}
+	}
+
+	return allWarnings, PolicyDoesNotExistError{}
+}
+
+func transformPolicy(applications []v3action.Application, v1Policy cfnetv1.Policy) Policy {
```
mariash: the `transformPolicy` method should be defined on the actor so that it is not floating in the package.

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,184 @@
...
+	var v1Policies []cfnetv1.Policy
+	v1Policies, err = actor.NetworkingClient.ListPolicies()
+	if err != nil {
+		return []Policy{}, allWarnings, err
+	}
+
+	var policies []Policy
+	emptyPolicy := Policy{}
+	for _, v1Policy := range v1Policies {
+		policy := transformPolicy(applications, v1Policy)
```
mariash: instead of calling `transformPolicy` for each policy, which loops over every app, it would be more efficient to create first a map of app GUID to app name and in current loop just get the app name by its GUID.

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,184 @@
...
+		if policy != emptyPolicy {
+			policies = append(policies, transformPolicy(applications, v1Policy))
```
mariash: `transformPolicy` is being called here again, instead you could use the policy variable.

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,184 @@
...
+		return []Policy{}, allWarnings, err
+	}
+
+	var policies []Policy
+	emptyPolicy := Policy{}
+	for _, v1Policy := range v1Policies {
+		if appGUID == "" || v1Policy.Source.ID == appGUID {
+			policy := transformPolicy(applications, v1Policy)
+			if policy != emptyPolicy {
+				policies = append(policies, transformPolicy(applications, v1Policy))
```
mariash: same as above: `transformPolicy` is being called here again, instead you could use the policy variable.

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	v1Policies, err := actor.NetworkingClient.ListPolicies(srcApp.GUID)
+	if err != nil {
+		return allWarnings, err
+	}
+
+	for _, v1Policy := range v1Policies {
+		if v1Policy == policyToRemove {
+			return allWarnings, actor.NetworkingClient.RemovePolicies([]cfnetv1.Policy{policyToRemove})
```
mariash: instead of getting list of policies and trying to find it in order to remove it, better approach would be to just delegate that to `RemovePolicies` method on networking client. It would make 1 http call instead of 2. And it is not going to be vulnerable to race conditions where `ListPolicies` has it and `RemovePolicies` will actually fail to remove it. `RemovePolicies`. Or the other way around. `RemovePolicies` can just throw a corresponding error when API returns 404.

api/cfnetworking/cfnetv1/policy.go
```diff
@@ -52,3 +53,52 @@ func (client Client) CreatePolicies(policies []Policy) error {
...
+	var err error
+	if len(appGUIDs) == 0 {
+		request, err = client.newHTTPRequest(requestOptions{
+			RequestName: internal.ListPolicies,
+		})
+	} else {
+		request, err = client.newHTTPRequest(requestOptions{
+			RequestName: internal.ListPolicies,
+			Query: map[string][]string{
+				"id": {strings.Join(appGUIDs, ",")},
```
mariash: if your API accepts and empty `id` then you won't need to have an if/else block. But it depends on how your API handles it...

api/cfnetworking/cfnetv1/policy_test.go
```diff
@@ -12,10 +12,7 @@ import (
 
 var _ = Describe("Policy", func() {
 	var client *Client
-
-	BeforeEach(func() {
-		client = NewTestClient()
-	})
+	BeforeEach(func() { client = NewTestClient() })
```
mariash: usually, go fmt keeps BeforeEach on separate lines. please make sure you ran go fmt. (`bin/test` should run go fmt for you)

api/cfnetworking/cfnetv1/policy_test.go
```diff
@@ -114,4 +111,227 @@ var _ = Describe("Policy", func() {
 			})
 		})
 	})
+
+	Describe("ListPolicies", func() {
+		var expectedPolicies []Policy
+		Context("when the policies are found", func() {
+			BeforeEach(func() {
+				expectedResponse := `{
```
mariash: can be just response, since it is not expected

api/cfnetworking/cfnetv1/policy_test.go
```diff
@@ -114,4 +111,227 @@ var _ = Describe("Policy", func() {
...
+				}`
+				server.AppendHandlers(
+					CombineHandlers(
+						VerifyRequest(http.MethodGet, "/policies"),
+						RespondWith(http.StatusBadRequest, response),
+					),
+				)
+			})
+
+			It("returns the error and warnings", func() {
```
mariash: warnings are not tested, or maybe there is no warnings?

api/cfnetworking/cfnetv1/policy_test.go
```diff
@@ -114,4 +111,227 @@ var _ = Describe("Policy", func() {
...
+				}`
+				server.AppendHandlers(
+					CombineHandlers(
+						VerifyRequest(http.MethodPost, "/policies/delete"),
+						RespondWith(http.StatusBadRequest, response),
+					),
+				)
+			})
+
+			It("returns the error and warnings", func() {
```
mariash: same as above: warnings are not tested, or maybe there is no warnings?

command/v3/add_network_policy_command_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	Context("when the user is logged in", func() {
+		BeforeEach(func() {
+			fakeConfig.CurrentUserReturns(configv3.User{Name: "some-user"}, nil)
+			fakeConfig.TargetedSpaceReturns(configv3.Space{Name: "some-space", GUID: "some-space-guid"})
+			fakeConfig.TargetedOrganizationReturns(configv3.Organization{Name: "some-org"})
+		})
+
+		It("outputs flavor text", func() {
+			Expect(testUI.Out).To(Say("Allowing network traffic from app %s to %s in org some-org / space some-space as some-user...", srcApp, destApp))
```
mariash: the `...` are not escaped; they would match every letter. Ideally you would escape them. We noticed this in a lot of test expectations. Please check them. (we know we also sometimes forget to escape them :))

actor/cfnetworkingaction/policy.go
jaydunk: Our API will return a successful response when the policy being deleted does not exist. Is it desired/required that the CLI return an error in this situation? If so then we should probably leave this as is, but I would lean towards not returning an error at all.

api/cfnetworking/cfnetv1/policy_test.go
jaydunk: Our client connection does not generate any warnings, we're changing the test description to reflect this.

actor/cfnetworkingaction/policy.go
mariash: @jaydunk usually CLI delete commands do not fail when object they are trying to delete does not exist, they will print a warning to user only.

api/cfnetworking/cfnetv1/policy.go
jaydunk: It looks like we will need to keep the if/else clause, the API will always return an empty policy list if we pass an empty string as a query param

actor/cfnetworkingaction/policy.go
jaydunk: Yeah it looks like that is the current behavior, we create an error in the actor which is caught in the command and converted to a warning. This was the recommendation of the CLI team.

XenoPhex: It might not be possible to throw the error/display the warning if the API does not error when deleting non-existant objects. We should let Dies know about this during acceptance.

jaydunk:
We've updated the PR to reflect the latest feedback from your code review

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,185 @@
...
+		return []Policy{}, allWarnings, err
+	}
+
+	var policies []Policy
+	emptyPolicy := Policy{}
+	for _, v1Policy := range v1Policies {
+		if v1Policy.Source.ID == appGUID {
+			policy := actor.transformPolicy(appNameByGuid, v1Policy)
+			if policy != emptyPolicy {
+				policies = append(policies, actor.transformPolicy(appNameByGuid, v1Policy))
```
mariash: instead of calling `actor.transformPolicy` you could use `policy` variable, it seems like they are doing the same thing

actor/cfnetworkingaction/policy.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, app := range applications {
+		appNameByGuid[app.GUID] = app.Name
+	}
+
+	var policies []Policy
+	emptyPolicy := Policy{}
+	for _, v1Policy := range v1Policies {
+		policy := actor.transformPolicy(appNameByGuid, v1Policy)
+		if policy != emptyPolicy {
+			policies = append(policies, actor.transformPolicy(appNameByGuid, v1Policy))
```
mariash: instead of calling `actor.transformPolicy` you could use `policy` variable, it seems like they are doing the same thing

PR #1226, opened 2017-09-13
Fix commit fd90a6273c4dfe58f479c89c15a47aecd5bbacba

ice-ship:
It should be possible to publish docker apps with docker image defined in manifest.yaml, without -o option


nickwei84:
@ice-ship Good catch. This work was ongoing and you caught it while it was being addressed in this tracker story: https://www.pivotaltracker.com/story/show/97697116.

PR #1229, opened 2017-09-18
ccv3: Add list spaces call

zmb3:
## What Need Does It Address?

The ccv3 package exposes a call for _list organizations_, but the corresponding call for listing spaces is missing.  Users of a `ccv3.Client` can list orgs, but if they want to list spaces they have to create a `ccv2.Client`.  With this change both orgs and spaces can be listed with a `ccv3.Client`.

## Possible Drawbacks

None

It's a missing API call.

## Description of the Change

Add `GetSpaces()` call.  The API is identical to that of `GetOrganizations()`.

## Alternate Designs

None

## Applicable Issues

None


nickwei84:
@zmb3 Thanks for the PR! Merged.

zmb3:
Thanks for the quick response @nickwei84!

Issue #1261, opened 2017-11-01
Further questionable words in randomly generated routes


mkretz:
## Command  

cf push example-app -d shared-domain.example.com --random-route

## What occurred

An app received the route static-app-nonarithmetic-**paedophilia**.scapp.io

## What you expected to occur

No problematic words occur in randomly generated routes

Follow-up to PR #217. I checked [`nouns.txt`](https://github.com/cloudfoundry/cli/blob/master/util/words/dict/nouns.txt) and [`adjectives.txt`](https://github.com/cloudfoundry/cli/blob/master/util/words/dict/adjectives.txt). It is just a matter of removing this word (and related ones) from these two files. Do you agree that I create a pull request?

dkoper:
That would be great, thanks!

mkretz:
Thanks! Created PR #1262.

dkoper:
Thanks!

PR #1262, opened 2017-11-01
stop words related to paedophilia appearing in random routes

mkretz:
## Description of the Change
Removed words related to paedophilia from the dictionary used for random route generation.

## Applicable Issues
#1261


nickwei84:
@mkretz Thanks for making this PR!

Issue #577, opened 2015-09-14
Consider support for HTTP proxy authentication incl windows NTLM 


gberche-orange:
Some CLI users are behind a corporate firewall requiring NTLM authentication. The current workaround for these users to allow connecting to public CloudFoundry instances such as IBM bluemix or pivotal PWS, it to install a local http proxy such as CNTLM which exposes a local http proxy without authentication support.

This has the following drawbacks:
- requires a set up task by every CL user
- local http proxy such as cntlm are sometimes unreliable and may interfere with CF CLI requests, in particular for high request rates over HTTP persistent connection, see [bug 61](http://sourceforge.net/p/cntlm/bugs/61/) as an example

More details about NTLM authentication protocol
https://msdn.microsoft.com/en-us/library/ms789031%28v=vs.110%29.aspx
https://msdn.microsoft.com/en-us/library/dd925287%28v=office.12%29.aspx

This problem is not visible to cf users using UIs since most browsers support ntlm proxy authentication
https://www.chromium.org/developers/design-documents/http-authentication
https://support.mozilla.org/en-US/questions/926378 and detailed docs about [network.automatic-ntlm-auth.allow-proxies](http://kb.mozillazine.org/Network.automatic-ntlm-auth.allow-proxies)
http://stackoverflow.com/questions/5541365/configure-ntlm-for-opera-and-safari-and-chrome

Potential fix steps:
* add support for http proxy NTLM authentication to use current windows session. See [curl -U option](http://curl.haxx.se/docs/manpage.html#-u) for inspiration. This could be enabled through a ``cf config enable_automatic_ntml_proxy_auth true`` command
* add support for general http proxy authentication by specifying a proxy login/password (e.g. by a refinement of the HTTP_PROXY environment variable in the form of ``protocol://username:password@host:port/``  through a common
http://curl.haxx.se/docs/manpage.html#--proxy-ntlm

Example NTLM go-lang lib https://github.com/ThomsonReutersEikon/go-ntlm

I realize this enhancement suggestion may be too specific to specific user base. Feel free to close the issue if you feel so.

Some background on rationale for some organizations to use NTLM authentication for outgoing internet access http proxies:
* some industry sectors / countries have regulations that require to trace external internet traffic back to originated user accounts (e.g. SOX, or EU/FR laws)
* some computers can be run by multiple user accounts, so host IP address/mac address is not authentifying user account who triggered outgoing internet traffic 
[...]

PR #1270, opened 2017-11-17
experimental: NTLM proxy authentication using NTLM_PROXY environment variable

gogolok:
Go does not support NTLM proxy authentication by default.

An attempt https://github.com/golang/go/issues/22288 to add NTLM
proxy authentication to Go's code base has not been accepted.

But there is a workaround/hack overwriting http.Transport.DialContext
to do NTLM proxy authentication (also mentioned in https://github.com/golang/go/issues/22288).

Unfortunately, there was no follow-up discussion in https://github.com/cloudfoundry/cli/issues/577
Therefore I'm just posting the current WIP here to discuss how NTLM proxy authentication might be possible.

### DIFF
--- a/api/cloudcontroller/cloud_controller_connection.go
+++ b/api/cloudcontroller/cloud_controller_connection.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccerror"
+	"code.cloudfoundry.org/cli/cf/net/dialcontext"
 )
 
 // CloudControllerConnection represents a connection to the Cloud Controller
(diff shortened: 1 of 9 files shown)


dkoper:
Hi @gogolok 

It's unfortunate that Golang is not providing NTLM support out of the box.

It would be nice if this patch just worked with `https_proxy` instead of the user having to be aware that their proxy is an NTLM proxy and needing to specify a custom environment variable for it. And the CLI needing to expose and document a new environment variable for what some deem to be a ` specific and narrow use case`.
Is that possible?

Cheers,
Dies Koper
CF CLI PM

gogolok:
Hi @dkoper,

I'll check whether it's possible to reuse http(s)_proxy and let you know.

Robert

geekwise-nicole-deltoro:
k

XenoPhex:
Closing due to inactivity as well waiting for GoLang to support NTLM via `https_proxy`.

PR #1273, opened 2017-11-27
Order logs by timestamp

bradylove:
## What Need Does It Address?

When a user invokes `cf logs <app-name>`, streaming logs will be ordered.

## Possible Drawbacks

There will be a slight (300ms) delay when streaming logs.

A user would be surprised if their application log stream did not appear in chronological order.

## Description of the Change

Rather than write a log message into a channel immediately, the code here adds a 300ms delay and orders those logs before passing them onto the channel.

### DIFF
--- a/actor/v2action/logging.go
+++ b/actor/v2action/logging.go
@@ -1,6 +1,7 @@
 package v2action
 
 import (
+	"sort"
 	"time"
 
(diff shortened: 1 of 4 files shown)


dkoper:
This was incorporated into cf CLI 6.33.1. Thanks again!

PR #1275, opened 2017-11-29
Add v3-share-service command

Samze:
[As a CLI user, I want to share my service instance with another space so I can bind it to my apps there](https://www.pivotaltracker.com/story/show/151669275)

## What Need Does It Address?

This PR adds the new `v3-share-service` command which allows space developers to share service instances into other spaces for binding. You can see a broad summary in [these slides](https://docs.google.com/presentation/d/1NMGK5dzW7WEqJ5bUjX6uXPRThb1v-oRthQFHaphpCLE/edit#slide=id.gd3f84e0d5_3_52) and a design doc [here.](https://docs.google.com/document/d/1V5m7L6TPMdR1phRWqzjzCoQIlw44FWZHjKxBsDjUaQQ/edit#heading=h.jrol36ywo0sr).

## Description of the Change

This change introduces the following command:
```
cf v3-share-service
Incorrect Usage: the required flag `-s' was not specified

NAME:
   v3-share-service - Share a service instance with another space

USAGE:
   cf v3-share-service SERVICE_INSTANCE -s OTHER_SPACE [-o OTHER_ORG]

OPTIONS:
   -o      Org of the other space (Default: targeted org)
   -s      Space to share the service instance into

SEE ALSO:
   bind-service, service, services
```

#### Environment requirements

The CF instance must have the `service_instance_sharing` feature flag enabled. See note below about minimum supported capi-release version.

#### Service requirements

The service broker must specify that it is shareable in it's `/v2/catalog` service metadata. [see here](https://docs.google.com/document/d/1F59E63d0L47pNTa1h_XyiKnw1zz2g5rbAz1P4L1qZ4E/edit).  For this first version only non-route managed services are supported. 

#### User requirements
A user who has SpaceDeveloper permissions in both source and destination space can share services.

## Notes

* The command `v3-share-service` resolves the service instance name to a guid in order to perform the POST. Currently the `/v3/service_instances` API does not support user provided services, only managed. Therefore when looking up the service instance guid and 404 is returned, there no way to distinguish between the instance not existing and it being a user provided service. We therefore catch that 404 and wrap it in a more suitable error message: `Specified instance not found or not a managed service instance. Sharing is not supported for user provided services` . Once there is a way to get user provided services in v3 this can be removed.
* This command requires [this](https://github.com/cloudfoundry/cloud_controller_ng/pull/1007) Cloud controller PR to be merged first. At the moment we have specified `3.34.0` as the minimum v3 API version for this command (see api/cloudcontroller/ccversion/minimum_version.go). However, once that PR gets merged and a capi-release cut, it is likely to change. We are submitting this now anyway to get feedback!
* Unshare PR coming soon.

Thanks, SAPI team (Sam & @deniseyu)


actor/actionerror/shared_service_instance_not_found.go
```diff
@@ -0,0 +1,9 @@
+package actionerror
+
+// SharedServiceInstanceNotFound is returned when a service instance is not found when performing a share service.
+type SharedServiceInstanceNotFound struct {
```
nickwei84: Please create a translatable error for this actionerror so that the error message can be localized. You can use this translatable error as an example: https://github.com/cloudfoundry/cli/blob/ee9d4748db79bc07352776e250383873c3e578cf/command/translatableerror/authorization_endpoint_not_found_error.go#L1-L12

Please also make sure the actionerror is converted to this new translatable error by adding it to this file: https://github.com/cloudfoundry/cli/blob/ee9d4748db79bc07352776e250383873c3e578cf/command/translatableerror/convert_to_translatable_error.go#L17

And update unit tests here: https://github.com/cloudfoundry/cli/blob/ee9d4748db79bc07352776e250383873c3e578cf/command/translatableerror/convert_to_translatable_error_test.go#L20
and here: https://github.com/cloudfoundry/cli/blob/ee9d4748db79bc07352776e250383873c3e578cf/command/translatableerror/translate_test.go#L32

actor/v3action/space_test.go
```diff
@@ -145,4 +147,78 @@ var _ = Describe("Space", func() {
...
+				})
+			})
+
+			Context("when the cloud controller returns back no spaces", func() {
+				BeforeEach(func() {
+					fakeCloudControllerClient.GetSpacesReturns(
+						nil, ccv3.Warnings{"some-space-warning"}, nil)
+				})
+
+				It("returns back the first space and warnings", func() {
```
nickwei84: the spec should say returns a SpaceNotFoundError

api/cloudcontroller/ccv3/internal/api_routes.go
```diff
@@ -48,6 +49,7 @@ const (
 	PostIsolationSegmentRelationshipOrganizationsRequest    = "PostIsolationSegmentRelationshipOrganizations"
 	PostIsolationSegmentsRequest                            = "PostIsolationSegments"
 	PostPackageRequest                                      = "PostPackageRequest"
+	PostServiceInstanceRelationshipSharedSpacesRequest      = "PostServiceInstanceRelationshipSharedSpaces"
```
nickwei84: please add a "s" to "Relationship" ("Relationships") because our naming convention follows the CC endpoint name

api/cloudcontroller/ccv3/service_instance.go
```diff
@@ -0,0 +1,39 @@
...
+	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/internal"
+)
+
+type ServiceInstance struct {
+	GUID string `json:"guid"`
+	Name string `json:"name"`
+}
+
+// GetServiceInstances lists ServiceInstances with optional filters.
+func (client *Client) GetServiceInstances(query url.Values) ([]ServiceInstance, Warnings, error) {
```
nickwei84: please add unit tests for this method

api/cloudcontroller/ccv3/relationship_list_test.go
```diff
@@ -92,4 +92,95 @@ var _ = Describe("RelationshipList", func() {
...
+		BeforeEach(func() {
+			serviceInstanceGUID = "some-service-instance-guid"
+			spaceGUIDs = []string{"some-space-guid", "some-other-space-guid"}
+		})
+
+		JustBeforeEach(func() {
+			relationshipList, warnings, executeErr = client.ShareServiceInstanceToSpaces(serviceInstanceGUID, spaceGUIDs)
+		})
+
+		Context("when the delete is successful", func() {
```
nickwei84: please change the spec to: when no errors are encountered

integration/experimental/v3_share_service_command_test.go
```diff
@@ -0,0 +1,297 @@
...
+			Context("when the org I want to share into does not exist", func() {
+				It("fails with an org not found error", func() {
+					session := helpers.CF("v3-share-service", serviceInstance, "-s", sharedToSpaceName, "-o", "missing-org")
+					Eventually(session).Should(Say("FAILED"))
+					Eventually(session.Err).Should(Say("Organization 'missing-org' not found"))
+					Eventually(session).Should(Exit(1))
+				})
+			})
+
+			Context("when the space I want to share into does not exist", func() {
```
nickwei84: this is a question and not a suggestion, is this the same scenario as when the user does not have authorization for the share-to space?

actor/v3action/space_test.go
jenspinney: Whoops! Fixed :)

api/cloudcontroller/ccv3/internal/api_routes.go
jenspinney: 👍

api/cloudcontroller/ccv3/relationship_list_test.go
jenspinney: 👍

api/cloudcontroller/ccv3/service_instance.go
jenspinney: 😱 Not sure how this happened. Added now.

actor/actionerror/shared_service_instance_not_found.go
jenspinney: Added. That's some pro-level githubbing by the way 😄 (and a super helpful walkthrough!)

integration/experimental/v3_share_service_command_test.go
jenspinney: Well, it depends if the user has no read access or if they have read-but-not-write access.

For the no-read-access case, the behavior is the same as if the space doesn't exist at all. The API call we make is `GET /v3/spaces?names=<shared-to-space-name>&organization_guids=<org-guid>`, and the result is the same (an empty list) whether it doesn't exist or you don't have permission to view.

For the read-but-not-write-access case, we fail later during the API call to actually share the service instance into that space. The error the user sees in this case is
```
You are not authorized to perform the requested action
FAILED
```

Should we try to add a test for the second case?

nickwei84: yes please add an integration test for the second case thanks!

jenspinney: Great, we've added an integration test for this now :)

PR #1281, opened 2017-12-04
Add v3-unshare-service command

jenspinney:
**NOTE**: This PR builds on top of #1275, which should be merged first. The actual changes on top of #1275 can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cli-sapi/compare/pr-service-instance-sharing-v3-share-service...cloudfoundry-incubator:pr-service-instance-sharing-v3-unshare-service).

This command won't work with the latest release of capi-release until [this PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/1016) is merged by CAPI and that commit makes it into a capi-release.

## What Need Does It Address?

We have a previous PR that allows users to share a service instance another space (we call this the "shared-to" space in the code comments and tests). This PR adds the `v3-unshare-service` command, which removes the sharing relationship previously created via `v3-share-service`.

## Description of the Change

To remove a sharing assocation between a service instance and a shared-to space, an DELETE request is made to `/v3/service_instances/:service_instance_guid/relationships/shared_spaces/:shared_to_space_guid`. Because the user issuing the unshare request may not have read access to the shared-to space and thus wouldn't normally be able to resolve the space guid from the space name, we need to first make a call to `/v2/service_instance/:service_instance_guid/shared_to` to find the space's GUID based on the space's name and org name.

Because the v3 API for this hasn't been designed yet, we're using the v2 API to do this space GUID resolution. This mixing of v3 and v2 calls can be fixed in the future once there is a v3 endpoint that would allow this GUID resolution.

Thanks,
SAPI team


command/v3/v3_unshare_service_command.go
```diff
@@ -0,0 +1,123 @@
...
+		"OrgName":             orgName,
+		"SpaceName":           cmd.SpaceName,
+		"Username":            user.Name,
+	})
+
+	sharedToSpaceGUID, warningsV2, err := cmd.ActorV2.GetSharedToSpaceGUID(cmd.RequiredArgs.ServiceInstance, cmd.Config.TargetedSpace().GUID, orgName, cmd.SpaceName)
+	cmd.UI.DisplayWarnings(warningsV2)
+
+	if err != nil {
+		if _, ok := err.(actionerror.ServiceInstanceNotSharedToSpaceError); ok {
```
nickwei84: please handle the error conversion here: https://github.com/cloudfoundry/cli/blob/5923b92489b9dcb96bbc3c8ebb413fd27a1e18bb/command/translatableerror/convert_to_translatable_error.go#L17
and add a unit test here: https://github.com/cloudfoundry/cli/blob/5923b92489b9dcb96bbc3c8ebb413fd27a1e18bb/command/translatableerror/convert_to_translatable_error_test.go#L26

command/v3/v3_unshare_service_command_test.go
```diff
@@ -0,0 +1,233 @@
...
+				})
+
+				Context("when the unsharing is successful", func() {
+					BeforeEach(func() {
+						fakeActor.UnshareServiceInstanceFromSpaceReturns(
+							v3action.Warnings{"unshare-service-warning"},
+							nil)
+					})
+
+					It("unshares the service instance with the provided space and org and displays all warnings", func() {
```
nickwei84: Please check GetSharedToSpaceGUID call count and args passed in on the first call

command/v3/v3_unshare_service_command_test.go
```diff
@@ -0,0 +1,233 @@
...
+				})
+
+				Context("when the unsharing is successful", func() {
+					BeforeEach(func() {
+						fakeActor.UnshareServiceInstanceFromSpaceReturns(
+							v3action.Warnings{"unshare-service-warning"},
+							nil)
+					})
+
+					It("unshares the service instance with the provided space and displays all warnings", func() {
```
nickwei84: Please check GetSharedToSpaceGUID call count and args passed in on the first call

actor/v2action/service_instance_test.go
```diff
@@ -306,4 +306,251 @@ var _ = Describe("Service Instance Actions", func() {
...
+				It("returns an error and warnings", func() {
+					Expect(executeErr).To(MatchError("some-shared-to-api-failure"))
+					Expect(warnings).To(Equal(Warnings{
+						"get-space-service-instances-warning",
+						"get-service-instance-shared-tos-warning"}))
+				})
+			})
+		})
+
+		Context("when the service instance does not exist", func() {
```
nickwei84: Please add a unit test for the general failure test case when looking up the service instance fails.

actor/v2action/service_instance_test.go
```diff
@@ -306,4 +306,251 @@ var _ = Describe("Service Instance Actions", func() {
...
+							Name: "some-service-instance",
+						},
+					},
+					ccv2.Warnings{"get-space-service-instances-warning"},
+					nil,
+				)
+			})
+
+			It("calls GetSpaceServiceInstance with the correct service instance name and space guid", func() {
+				spaceGUIDArg, _, queryArg := fakeCloudControllerClient.GetSpaceServiceInstancesArgsForCall(0)
```
nickwei84: Please check call count before checking args, otherwise this could potentially throw a panic.

actor/v2action/service_instance_test.go
```diff
@@ -306,4 +306,251 @@ var _ = Describe("Service Instance Actions", func() {
...
+
+				Expect(spaceGUIDArg).To(Equal("some-source-space-guid"))
+				Expect(queryArg[0].Values[0]).To(Equal("some-service-instance"))
+			})
+
+			It("calls GetServiceInstanceSharedTos with the correct service instance guid", func() {
+				serviceInstanceGUIDArg := fakeCloudControllerClient.GetServiceInstanceSharedTosArgsForCall(0)
```
nickwei84: Please check call count before checking args, otherwise this could potentially throw a panic.

integration/experimental/v3_unshare_service_command_test.go
```diff
@@ -0,0 +1,319 @@
...
+				})
+
+				Context("when I want to unshare my service instance from the space", func() {
+					It("successfully unshares the service instance when I am admin", func() {
+						session := helpers.CF("v3-unshare-service", serviceInstance, "-s", sharedToSpaceName)
+						Eventually(session).Should(Say("OK"))
+						Eventually(session).Should(Exit(0))
+					})
+
+					Context("when I have no access to the shared-to space", func() {
```
nickwei84: should this say when I only have read access?

ablease: Hey @nickwei84 

To share a service the user must have SpaceDeveloper permissions in the source space & read permission in the target space. However, the desired behaviour is that users can unshare service instance when they do not have any access to the shared-to space. 

This is so SpaceDevelopers in the source space maintain complete control of that instance. So that for example, service instances cannot be held hostage in shared-to spaces. 

Sapi, @Samze + Alex

command/v3/v3_unshare_service_command_test.go
ablease: Thanks, added a new test for this.

command/v3/v3_unshare_service_command_test.go
ablease: Thanks, added a new test for this.

actor/v2action/service_instance_test.go
ablease: Tahnks, new test added for the error case.

command/v3/v3_unshare_service_command.go
ablease: Thanks, changes this to use ConvertToTranslatableErrror instead.

command/v3/v3_unshare_service_command_test.go
nickwei84: @ablease I forgot to add the reason why, it's because if we don't check call count first and the method is not called, then we will get a hard panic in the code.

integration/experimental/v3_unshare_service_command_test.go
nickwei84: Cool thanks!

PR #1309, opened 2018-01-18
Add ability to provide a name when creating a service binding

elenasharma:
## Description of the Change

A user can now provide a name for their service binding. This name is `BindingName` until the API layer, where it is `Name` because that is what the CC API expects.

The default name, if none is provided, is an empty string.

#### CAPI story: https://www.pivotaltracker.com/story/show/152947740


command/v2/bind_service_command.go
```diff
@@ -44,7 +47,14 @@ func (cmd *BindServiceCommand) Setup(config command.Config, ui command.UI) error
...
 func (cmd BindServiceCommand) Execute(args []string) error {
-	err := cmd.SharedActor.CheckTarget(true, true)
+	var err error
+
+	err = cmd.SharedActor.CheckTarget(true, true)
+	if err != nil {
+		return err
+	}
+
+	err = command.MinimumAPIVersionCheck(cmd.Actor.CloudControllerAPIVersion(), ccversion.MinVersionProvideNameForServiceBinding, "Option '--name'")
```
XenoPhex: Note to self - check to see if we say `Option --flagname` or just `--flagname`.

jbpivotal:
@zrob 
1. For the current story, when there is a binding name, please output:
`Binding service s to app a with binding name [binding name] in org test / space test as admin...`
If no binding name, output is unchanged. 
We can update `services` in a later story.
2. OK for now.
3. Please use flag `--binding-name`.

PR #1317, opened 2018-02-01
Add test case for permissions in spaces that receive shared service instances

deniseyu:
## Description

This additional integration test addresses an [edge case](https://www.pivotaltracker.com/n/projects/892938/stories/151669277/comments/185217233) where users with read access to a service instance as a result of it being shared into their space were seeing 404s when curling the `/v2/service_instances/:guid/shared_to` endpoint as part of the `cf v3-unshare-service` command.

This endpoint now returns a 403 for permissioned users (anyone in the space where the service instance was shared to, who is either a SpaceDeveloper, SpaceAuditor, SpaceManager, or OrgManager). OrgBillingManagers and OrgAuditors, as well as users not in either the source or target space, will continue to receive 404.

This should be blocked on a new release of CAPI that contains the [change to the cloud_controller endpoint](https://github.com/cloudfoundry/cloud_controller_ng/pull/1077).

### DIFF
--- a/integration/experimental/v3_unshare_service_command_test.go
+++ b/integration/experimental/v3_unshare_service_command_test.go
@@ -389,5 +389,63 @@ var _ = PDescribe("v3-unshare-service command", func() {
 				})
 			})
 		})
+
+		Context("when there is a shared service instance in my current targeted space", func() {
+			var broker helpers.ServiceBroker
+			var user string
+			var password string
+
+			BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


jenspinney:
In reference to "This should be blocked on a new release of CAPI that contains the change to the cloud_controller endpoint," capi-release 1.49 has the necessary commit, and the latest cf-deployment (v1.12.0) has capi-release 1.49.

So, this PR now has no blockers and should be merge-able. @nickwei84

Issue #1373, opened 2018-04-30
CTRL+C makes the CLI exit before a plugin can run cleanup tasks


46bit:
## Command  

`cf conduit example-postgres` followed by a `CTRL+C`

## What occurred

The CF CLI process received the `SIGINT` and exited immediately. Our [Conduit plugin](https://github.com/alphagov/paas-cf-conduit) received the `SIGINT` and started to delete its CF resources before exiting, but the CF CLI's RPC server is no longer running.

## What you expected to occur

We expected the CF CLI to wait for our plugin to terminate before exiting.

## What impact this has

This makes it harder for plugins to clear up after themselves. Our plugin cannot count on the CF CLI's RPC server running when it cleans up after a `CTRL+C`. This means we duplicate a lot of code.

## CLI Version  

Current master.

## Platform & Shell Details  

Mac OS X 10.11 iTerm

## Any other relevant information  

We're proposing a PR for this shortly.

XenoPhex:
Merged your pull request, thanks for the fix! It'll be in the next version of the CF CLI.

XenoPhex:
:sob: 

I've had to revert this commit due to segfaults in our pipeline. While hard to trigger by hand - it appears that this change causes the plugin interface to error in an extremely unexpected way. (I'll post the segfault below). For now reverting is the safest solution.

Integration test log:
```
[1525460363] Plugin Suite - 140 specs - 16 nodes •••
------------------------------

CMD> cf api https://api.tartine.cli.fun --skip-ssl-validation 
OUT: Setting api endpoint to https://api.tartine.cli.fun...
OUT: OK
OUT: 
OUT: api endpoint:   https://api.tartine.cli.fun
OUT: api version:    2.107.0
OUT: Not logged in. Use 'cf login' to log in.

CMD> cf auth admin u4a5sqn0km8nblhmsek5 
OUT: API endpoint: https://api.tartine.cli.fun
OUT: Authenticating...
OUT: OK
OUT: Use 'cf target' to view or set your target org and space.

CMD> cf remove-plugin-repo CF-Community 
OUT: 
OUT: OK
OUT: CF-Community removed from list of repositories
OUT: 

CMD> cf install-plugin -f /tmp/gexec_artifacts797859608/g389912983/test_plugin 
OUT: Attention: Plugins are binaries written by potentially untrusted authors.
OUT: Install and use plugins at your own risk.
OUT: Installing plugin CF-CLI-Integration-Test-Plugin...
[...]

PR #1374, opened 2018-04-30
On an exit signal, wait for plugins to exit

46bit:
## What Need Does It Address?

In #1373 we reported the CF CLI exits immediately when it receives a `SIGINT` (e.g., when a user runs `CTRL+C` in their shell.) This is problematic because it means that plugins can't use the CF CLI's RPC server to perform cleanup tasks.

## Possible Drawbacks

If a `SIGINT` is directed at the CF CLI process alone, rather than the process tree, the CF CLI will no longer exit. This could be worked around by only intercepting the first `SIGINT`, or relaying signals to the plugin's child process.

The RPC server is in Core and it keeps it alive for cleanup.

## Description of the Change

See commit message.

## Alternate Designs

This was a first pass at the problem, which ignores all signals until the plugin exits. If you review we're happy to adopt a different design, such as the first signal triggering a controlled shutdown and the second force-killing.

## Applicable Issues

https://github.com/cloudfoundry/cli/issues/1373

### DIFF
--- a/integration/assets/test_plugin/test_plugin.go
+++ b/integration/assets/test_plugin/test_plugin.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
+	"time"
 
 	"code.cloudfoundry.org/cli/plugin"
 )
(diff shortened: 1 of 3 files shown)


XenoPhex:
This seems like a reasonable change - I've tested this with a few plugins and nothing unexpected [too] happened. Given the nature of the plugin architecture we don't know what plugins this might affect overall.

We'll merge it in with the expectation that if this breaks other plugins in a weird ways that it'll be reverted.

Issue #1375, opened 2018-05-08
Passwords not properly sanitized in CF_TRACE


andreasf:
## Command  

`CF_TRACE=1 cf service-keys` (and others)

## What occurred

- Passwords in query parameters are not sanitized if the parameter is the last one
- Passwords in the userinfo component of URIs (`mysql://user:pass@host/db`) are not sanitized
- Half the response is missing

## What you expected to occur

- The cli should properly sanitize passwords in query parameters, regardless of parameter order
- The cli should probably also sanitize the userinfo URI component
- CF_TRACE should show the full, sanitized response

## CLI Version  

cf version 6.36.1+e3799ad7e.2018-04-04

## CC API Endpoint Version  

api endpoint:   https://api.run.pivotal.io
api version:    2.109.0

## CF Trace

```
$ CF_TRACE=1 cf service-keys fine-service
[...]
REQUEST: [2018-05-08T14:47:16+02:00]
GET /v2/service_instances/48741228-398f-4660-9ad9-b848f21cb3b8/service_keys HTTP/1.1
Host: api.run.pivotal.io
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
Content-Type: application/json
User-Agent: go-cli 6.36.1+e3799ad7e.2018-04-04 / darwin

RESPONSE: [2018-05-08T14:47:17+02:00]
HTTP/1.1 200 OK
Connection: close
Content-Length: 2239
Content-Type: application/json;charset=utf-8
Date: Tue, 08 May 2018 12:47:17 GMT
Server: nginx
X-Content-Type-Options: nosniff
X-Ratelimit-Limit: 20000
X-Ratelimit-Remaining: 19938
X-Ratelimit-Reset: 1525786392
X-Vcap-Request-Id: e63ba650-c5e9-483a-4464-517379d8818d::53ebfd5f-3f30-4725-9bcd-c16c0a913e38

{
  "total_results": 2,
  "total_pages": 1,
  "prev_url": null,
  "next_url": null,
  "resources": [
    {
      "metadata": {
        "guid": "81ba24fa-f6eb-4689-995a-21996a54ad64",
        "url": "/v2/service_keys/81ba24fa-f6eb-4689-995a-21996a54ad64",
        "created_at": "2018-05-08T12:45:24Z",
        "updated_at": "2018-05-08T12:45:24Z"
      },
      "entity": {
        "name": "key1",
[...]

PR #1376, opened 2018-05-08
Fix query parameter regex in trace.Sanitize()

andreasf:
## Description of the Change

The previous regular expression, `&password=[^&}*&`, did not match password parameters at the end of the query string. In that case it also ate the rest of the input, until the next & character.

I replaced it with an expression that is, sadly, rather complicated, but should match all characters that can be part of a query parameter value as per the RFC.

This change does not address passwords in the userinfo URI component.

## Alternate Designs

`trace.Sanitize()` could find URIs in the input string and parse them, then cleanly replace the userinfo component and parameter values

## Applicable Issues

#1375


XenoPhex:
Thanks for the fix!!

PR #1384, opened 2018-05-22
Tags for user provided services

lurraca:
## What Need Does It Address?

[Tracker Story](https://www.pivotaltracker.com/n/projects/2105761/stories/155062901)

In this PR we are adding support for setting tags on user provided service instances. The support for this functionality at the API level has already been merged in Cloud Controller ([See PR](https://github.com/cloudfoundry/cloud_controller_ng/pull/108)).

This makes user provided service instances consistent with managed service instances.

## Description of the Change

The `create-user-provided-service` and `update-user-provided-service` now have a `-t` flag that allows the user to pass in tags as paramaters to the request that will be sent from the CLI to the API.

Service API (SAPI)
@ablease  & @lurraca


Samze:
We saw this is failing in Travis, we are investigating!

Issue #1318, opened 2018-02-01
bind-route-service fails with "411" error from Akamai as content-type set but no content-length


andrew-edgar:
## Command  

cf bind-route-service stage0.mybluemix.net mylogger --hostname ae-spring-music-refreshful-advisableness

## What occurred

The error occurs if you have (as we do) a system validating the request (Akamai) as this is sent back with a "411" which means no content length.

## What you expected to occur

No error to occur as the request would have valid information

## CLI Version  

cf version 6.33.1+c77e55743.2017-12-15
But also still exists in MASTER branch as I tested with the latest code from github

## CC API Endpoint Version  

 2.92.0

## CF Trace

```
REQUEST: [2018-02-01T14:29:19+01:00]
PUT /v2/user_provided_service_instances/3f64eaf8-f6f2-4d06-8277-cfea98d6fa17/routes/2ad90452-0a61-465b-8d0d-00074809aecf HTTP/1.1
Host: api.stage0.ng.bluemix.net
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
Connection: close
Content-Type: application/json
User-Agent: go-cli 6.33.1+c77e55743.2017-12-15 / darwin



RESPONSE: [2018-02-01T14:29:20+01:00]
HTTP/1.1 411 Length Required
Connection: close
Content-Length: 222
Cache-Control: max-age=0, no-cache, no-store
Content-Type: text/html
Date: Thu, 01 Feb 2018 13:29:20 GMT
Expires: Thu, 01 Feb 2018 13:29:20 GMT
Mime-Version: 1.0
Pragma: no-cache
Server: AkamaiGHost

<HTML><HEAD>
<TITLE>Bad Request</TITLE>
</HEAD><BODY>
<H1>Bad Request</H1>
Your browser sent a request that this server could not understand.<P>
Reference&#32;&#35;7&#46;b7921602&#46;1517491760&#46;3a3a69d
</BODY>
</HTML>

FAILED
Server error, status code: 411, error code: 0, message:
```

## Platform & Shell Details  

MACOS but failed on also Linux

## Any other relevant information  

We have Akamai between the user and the endpoint which will validate requests and enforce https:

So I looked into this and found the issue occurs because the CLI is setting content-type to be application/json but in this request the body is empty and since it is empty the ContentLength is not sent.

I found several ways to fix this.   
1) https://github.com/cloudfoundry/cli/blob/master/cf/api/route_service_binding_repository.go#L55
[...]

PR #1404, opened 2018-06-22
Do not set content type when request body is blank in old net package

jberkhahn:
This checks to see if there is body content before setting the content type to application json in the OLD net package. The new net package already handles this correctly.

This is necessary as certain load balancers reject blank requests with application json types to prevent chunked requests. On many load balancers (akamai) this is the default behavior.

Every old command that can have a blank body will benefit from this change. Specifically we encountered difficulty with the bind route service command.

## Possible Drawbacks

I'm not aware of any possible drawbacks.

n/a


abbyachau:
Related issue [#1318](https://github.com/cloudfoundry/cli/issues/1318)

abbyachau:
Hey @jberkhahn please when you folks can, test this again and let us know if the fix works. We'll pull it in once we have confirmation, thanks again.

abbyachau:
@andrew-edgar for tracking purposes

PR #1409, opened 2018-06-25
add QuoteMeta to allow any timezone to be checked

ablease:
Fixed this test that fails due to the `+` in the GMT timezone:
```
• Failure [0.000 seconds]
Request Logger Terminal Display
/Users/pivotal/go/src/code.cloudfoundry.org/cli/util/ui/request_logger_terminal_display_test.go:14
  DisplayType
  /Users/pivotal/go/src/code.cloudfoundry.org/cli/util/ui/request_logger_terminal_display_test.go:180
    displays the passed type and time in localized ISO 8601 [It]
    /Users/pivotal/go/src/code.cloudfoundry.org/cli/util/ui/request_logger_terminal_display_test.go:181

    Got stuck at:
        banana: [2018-06-25T10:55:26+01:00]


    Waiting for:
        banana: \[2018-06-25T10:55:26+01:00\]

    /Users/pivotal/go/src/code.cloudfoundry.org/cli/util/ui/request_logger_terminal_display_test.go:189
```

Anand approved!

PR #1410, opened 2018-06-26
158448671 async bind service

nmaslarski:
- always send the accepts_incomplete=true on service binding as old CAPI
    versions will ignore the additional parameter. The display changes based
    on the status of the "last_operation.state"
    - update assets/service broker to include an async binding (but a sync
    service creation)
    - cleanup - remove unused variable
    - add some clarifying words around constants

    [Finishes #158448671]

**NOTE**: This PR builds on top of #1409 , which should be merged first. The actual changes on top of #1409  can be viewed in [this diff](https://github.com/cloudfoundry-incubator/cli-sapi/compare/fix-test-timezones...cloudfoundry-incubator:158448671-async-bind-service).


abbyachau:
Hey @nmaslarski @ablease thanks let us know when we can merge these in as you work through async stories. Thanks again.

PR #1425, opened 2018-07-23
Provide a default passcode prompt if login server does not provide one (OSS CF/cfdev)

drnic:
```
$ out/cf login -a https://api.v3.pcfdev.io --skip-ssl-validation --sso
API endpoint: https://api.v3.pcfdev.io

Temporary Authentication Code ( Get one at https://login.v3.pcfdev.io/passcode )>
Authenticating...
OK
```

The `cf login --sso` flag is awesome. It feels great to be putting my login details into the UAA (where I have 1password), than to have to copy/paste them into the terminal.

But by default, the cf-deployment + cfdev projects do not include a 'passcode' prompt, even though the UAA does have a /passcode endpoint. So the CLI misbehaves with the `--sso` flag.

This PR ensures that `--sso` works for all Cloud Foundry installations.

This PR is instead of me raising an Issue that `--sso` doesn't work for cfdev.

If the UAA does not include a `passcode` prompt, the CLI will decide on a nice default (the text comes from https://uaa.run.pivotal.io)

```
$ curl -s https://uaa.run.pivotal.io/login -H 'Accept: application/json' | jq .prompts.passcode
[
  "password",
  "Temporary Authentication Code ( Get one at https://login.run.pivotal.io/passcode )"
]
```

### DIFF
--- a/cf/commands/login.go
+++ b/cf/commands/login.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"strconv"
 
 	"code.cloudfoundry.org/cli/cf/commandregistry"
@@ -175,6 +176,12 @@ func (cmd Login) authenticateSSO(c flags.FlagContext) error {
 
 	credentials := make(map[string]string)
 	passcode := prompts["passcode"]
+	if passcode.DisplayName == "" {
+		passcode = coreconfig.AuthPrompt{
+			Type:        coreconfig.AuthPromptTypePassword,
+			DisplayName: fmt.Sprintf("Temporary Authentication Code ( Get one at %s/passcode )", cmd.config.AuthenticationEndpoint()),
+		}
+	}
 
 	for i := 0; i < maxLoginTries; i++ {
 		if c.IsSet("sso-passcode") && i == 0 {
--- a/cf/commands/login_test.go
+++ b/cf/commands/login_test.go
@@ -394,10 +394,6 @@ var _ = Describe("Login Command", func() {
 						DisplayName: "Username",
 						Type:        coreconfig.AuthPromptTypeText,
 					},
-					"passcode": {
-						DisplayName: "It's a passcode, what you want it to be???",
-						Type:        coreconfig.AuthPromptTypePassword,
-					},
 					"password": {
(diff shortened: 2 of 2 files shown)


drnic:
I'm confused about why my single commit PR says 250 commits. Investigating.

drnic:
Bump

drnic:
Bump

abbyachau:
@drnic sorry we didn't see this bump on this pr - thanks for creating this pull request. This sounds like a workaround instead of raising a feature request with `cf dev` to support `--sso` - we would prefer the change to be made on the `cf dev` side instead.  Please let us know if you have any questions, thanks. 

cc @ssisil

sreetummidi:
@drnic Passcode is meant only for flows that can't be achieved via the CLI. For ex. SAML. The UAA info endpoint intentionally suppresses this if there is not SAML provider enabled.

drnic:
I use —sso all the time. I’d prefer we made this flow the default rather than the password flow. I like the uaa cli’s authcode flow especially since you don’t have to copy/ paste the code.

But this PR is to just ensure we can allow users to start using this style of flow instead of password flow.


________________________________

aeijdenberg:
I agree with @drnic - passwords are long-lived credentials, and often managed by password managers that have good integrations with browsers, and far less good integration with CLIs. It's definitely useful to be able to copy/paste from the `/passcode` URL to the terminal, regardless of whether the user authentication is managed by UAA or externally.

Mind you, what would be even better is a proper OAuth integration from the `cf` CLI, ie use the reasonably standard technique of opening a server on http://localhost:someportnumber and handing the redirect (like Concourse does). But that's probably another issue for another day.

drnic:
I'm revisiting this PR. I'm disappointed that it wasn't merged - it was a simple fix for any cf deployment (`cf dev start` was just one example) that misconfigured something. There's no reason a cf user should be penalised because of a mistake by their platform operator. We should encourage `--sso` flag.

I'm also disappointed that we are suggesting `cf login -u -p` flags are preferred to `cf login --sso` (especially combined with #1617).

Can this please be reopened and merged?

I've rebased the branch.

PR #1430, opened 2018-08-07
Feature/origin parameter auth

tack-sap:
As discussed we provide a first version of the enhanced cf auth command with the --origin parameter.

Story: https://www.pivotaltracker.com/story/show/159412764

### DIFF
--- a/api/uaa/auth.go
+++ b/api/uaa/auth.go
@@ -33,8 +33,9 @@ func (client Client) Authenticate(ID string, secret string, origin string, grant
 
 	var query url.Values
 	if origin != "" {
-		query = url.Values{}
-		query.Set("login_hint", "{\"origin\":\""+origin+"\"}")
+		query = url.Values{
+			"login_hint": {string("{\"origin\":\"" + origin + "\"}")},
+		}
 	}
 
 	request, err := client.newRequest(requestOptions{
--- a/api/uaa/auth.go
+++ b/api/uaa/auth.go
@@ -1,7 +1,6 @@
 package uaa
 
 import (
-	"errors"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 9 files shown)


tack-sap:
What is included in this PR from the story:
Scenario: LDAP origin happy flow
Scenario: UAA origin happy flow
Scenario: OIDC origin happy flow
Scenario: no origin flag happy flag
Scenario: error message for client credential + origin flags
Scenario: error message for bad credentials (returning current cli authentication failure message)

The remaining error scenarios do return an error, but with the current cli error message based on the http status code (401) (UAA does return the desired error message, but cli only uses the status code and defines an own error message) see: https://github.com/cloudfoundry/cli/blob/master/cf/api/authentication/authentication.go#L126-L131

api/uaa/auth.go
```diff
@@ -18,17 +19,23 @@ type AuthResponse struct {
...
+func (client Client) Authenticate(ID string, secret string, origin string, grantType constant.GrantType) (string, string, error) {
 	requestBody := url.Values{
 		"grant_type": {string(grantType)},
 	}
 	switch grantType {
 	case constant.GrantTypeClientCredentials:
 		requestBody.Set("client_id", ID)
 		requestBody.Set("client_secret", secret)
+		if origin != "" {
+			return "", "", errors.New("Incorrect Usage: The following arguments cannot be used together: --client-credentials, --origin")
```
tjvman: This error contains flag information. The API clients _should_ know nothing about the command line.

Move this validation into the command layer. Look at [create-buildpack's `--enable`/`--disable`](https://github.com/cloudfoundry/cli/blob/master/command/v2/create_buildpack_command.go#L66-L70) to see how we implement it. And don't forget to add command layer tests for them.

api/uaa/auth.go
```diff
@@ -18,17 +19,23 @@ type AuthResponse struct {
...
+		}
 	default:
 		requestBody.Set("username", ID)
 		requestBody.Set("password", secret)
+		if origin != "" {
+			requestBody.Set("login_hint", "{\"origin\":\""+origin+"\"}")
```
tjvman: Per the [UAA docs](http://docs.cloudfoundry.org/api/uaa/version/4.19.0/index.html#authorization), this needs to be a query parameter with a value of URL encoded JSON. Thus, when creating the request (which is below) with `newRequest`, include the `Query` struct field in the `requestOptions`.

Additionally there should be a corresponding test validation for this change (not just passing the value into the function header).

torsten-sap:
@tjvman @DennisDenuto @tack-sap 
We provided the changes you asked for. Please review again. 

There is only one thing that we were not able to figure out and this is how to verify an URL parameter for origin. We had a look into the documentation but without success. The verification should be added here:
https://github.com/cloudfoundry/cli/blob/ce80437367e398ddc98ca1c307c7ae3eaafee1c5/api/uaa/auth_test.go#L86:L90

It would be great if you could take over this change when you merge the PR. Thanks a lot in advance!

Best regards
Florian & Torsten

tjvman:
Thanks for making the changes and letting us know about the query parameter issue.

PR #1447, opened 2018-08-23
Experimental zero-downtime push command

Gerg:
## What Need Does It Address?

Users can now push their apps without downtime!

Beta users of zero downtime deployments

Frequently by a small number of users

## Possible Drawbacks

Zero downtime deploys are still experimental and may have some undiscovered issues.

This should eventually be part of regular push

## Description of the Change

[Deployments](http://v3-apidocs.cloudfoundry.org/version/3.55.0/index.html#deployments), a new experimental resource on CAPI, transitions the web process of an application to a droplet, without having to stop the application. This lets users continue to serve traffic on their application while it is rolling to the next version.

## Alternate Designs

We decided to clone and extend v3_original_push instead of adding it to one of the existing v3_pushes. We did this because:
- v3_push is under active refactoring by the CLI team
- The CLI UX for zdt push is still under discussion, but we believe it will be the default behavior
- This will help make it clear to CLI users which push they are using

## Applicable Issues

N/A

## How Urgent Is The Change?

Zach will be happy if you merge it. We have beta users waiting to use it.

## Other Relevant Parties

@zrob


Gerg:
You will need this ops file in order for deployments to work: https://github.com/cloudfoundry/capi-ci/blob/master/cf-deployment-operations/temporary/add-deployment-updater.yml

PR #1470, opened 2018-09-28
Add clarifying notes for integration tests

ericpromislow:
Specifically, the docs currently don't point out that the onus is on
the developer to ensure that they are pointing to the correct instance
of cf and have built it before running integration tests.

## What Need Does It Address?

Coworkers on CAPI new to CLI dev were tripping up on the lack of documentation when writing/running integration tests.

The change is for non-CLI-team members trying to ensure they're building correct PRs by running integration tests.

Not very often. Only when non-CLI-team members make contributions.

## Possible Drawbacks

None

It's just documentation.

## Description of the Change

Added a paragraph to the integration-test README

## Alternate Designs

N/A

## Applicable Issues
N/A

## How Urgent Is The Change?

Normal

## Other Relevant Parties

No one

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -1,7 +1,23 @@
 # CLI Integration Tests
+
+## Introduction
+
 These are high-level tests for the CLI that make assertions about the behavior of the `cf` binary.
 
-These tests require that a `cf` binary built from the latest source is available in your `PATH`.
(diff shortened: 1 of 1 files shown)


abbyachau:
Thank you @ericpromislow!

PR #1474, opened 2018-10-05
add stack flag to v3-zdt-push

tcdowney:
[#161023897](https://www.pivotaltracker.com/story/show/161023897)

Thanks!
Tim and @cwlbraa 

## What Need Does It Address?

Upgrading stacks without downtime

beta zdt users

eventually, whenever folks need to update to cflinuxfs4

(and in a summit demo)

one day zdt-push will be push.

## Description of the Change

We added a stack flag to the zdt-push command. We also needed to fix some json marshalling in the API layer so that an absence of buildpacks wouldn't prevent the stack from being set on the update request.

## How Urgent Is The Change?

planned for use in a summit demo next week, but a custom binary has already been built.

## Other Relevant Parties

@zrob @cwlbraa @dieucao


integration/experimental/v3_zdt_push_command_test.go
```diff
@@ -554,6 +555,37 @@ var _ = Describe("v3-zdt-push command", func() {
...
+					session = helpers.CustomCF(helpers.CFEnv{WorkingDirectory: appDir}, "v3-zdt-push", appName, "-s", "cflinuxfs2")
+					Eventually(session).Should(Exit(0))
+				})
+			})
+
+			It("uses the specified stack", func() {
+				Eventually(session).Should(Say("name:\\s+%s", appName))
+				Eventually(session).Should(Say("requested state:\\s+started"))
+				Eventually(session).Should(Say("routes:\\s+%s\\.%s", appName, domainName))
+				Eventually(session).Should(Say("stack:\\s+cflinuxfs2"))
```
tcdowney: This is the default stack, so this example isn't really that great. We weren't sure if you regularly run these integration tests against multi-stack environments, though.

Additionally, if there's additional tests that you'd like to see please let us know. We were quite sure of where the boundaries between integration/unit testing lie within this codebase and would appreciate some guidance.

Thanks!

PR #1479, opened 2018-10-16
Refactor marketplace command

nmaslarski:
This commit is split into a few parts.

Firstly, it introduces some utility methods on the SharedActor for
checking whether the CLI is logged in, targeting a space and targeting
an org. The CheckTarget and Require methods has been refactored to make
use of this new functionality. These new methods were abstracted over
the very similar methods on the Config, but we thought that it would be
better to provide a layer of indirection to commands, such that we could
later decide to determine these behaviours in a different manner.

Secondly, we introduced a set of integration tests for the marketplace
command. They live in the global integration suite as they require
registration of service brokers, and enabling access to those services,
both of which result in global visibility changes that could affect
other tests.

Finally, we moved the marketplace command to the new v2 style,
introducing actor and cloud controller client methods to look up all
services, and a specific service by name from both the global services
endpoint and by space.


actor/sharedaction/is_logged_in_test.go
```diff
@@ -0,0 +1,65 @@
...
+	var (
+		actor      *Actor
+		binaryName string
+		fakeConfig *sharedactionfakes.FakeConfig
+	)
+
+	BeforeEach(func() {
+		binaryName = "faceman"
+		fakeConfig = new(sharedactionfakes.FakeConfig)
+		fakeConfig.BinaryNameReturns(binaryName)
```
jenspinney: It doesn't look like binary name is needed. If I comment out this line (and others relating to binaryName), the test still passes for me.

Same goes for IsOrgTargeted and IsSpaceTargeted below.

actor/sharedaction/is_org_targeted_test.go
```diff
@@ -0,0 +1,39 @@
...
+	BeforeEach(func() {
+		binaryName = "faceman"
+		fakeConfig = new(sharedactionfakes.FakeConfig)
+		fakeConfig.BinaryNameReturns(binaryName)
+		actor = NewActor(fakeConfig)
+	})
+
+	When("the config has an org targeted", func() {
+		It("returns true", func() {
+			fakeConfig.HasTargetedOrganizationReturns(true)
```
jenspinney: I think this reads a little cleaner in a BeforeEach (i.e., the style of the IsLoggedIn test).

Same goes for the other test in the file and for the tests in IsSpaceTargeted.

actor/v2action/cloud_controller_client.go
```diff
@@ -66,6 +66,7 @@ type CloudControllerClient interface {
 	GetSpaceServiceInstances(spaceGUID string, includeUserProvidedServices bool, filters ...ccv2.Filter) ([]ccv2.ServiceInstance, ccv2.Warnings, error)
 	GetSpaceStagingSecurityGroups(spaceGUID string, filters ...ccv2.Filter) ([]ccv2.SecurityGroup, ccv2.Warnings, error)
 	GetSpaces(filters ...ccv2.Filter) ([]ccv2.Space, ccv2.Warnings, error)
+	GetSpaceServices(spaceGUID string, filters ...ccv2.Filter) ([]ccv2.Service, ccv2.Warnings, error)
```
jenspinney: Could you move this right after `GetSpaceServiceInstances` to make it consistent with the alphabetization of this interface?

integration/shared/global/marketplace_command_test.go
```diff
@@ -0,0 +1,156 @@
+package global
```
jenspinney: I know this test is modifying global state because the output of the global `cf marketplace` command would change when this test is running, but global package tests make CI and dev-loops way slower, and we try to aggressively move them into the isolated suite whenever remotely possible.

In this case, as long as all integration tests that run `cf marketplace` don't do it in a way that counts lines of output (or something similar), I believe this *should* be ok in isolated. What do you think?

jenspinney:
I added some comments, but haven't finished reviewing. I figure incomplete feedback now is probably more useful than total silence.

integration/shared/isolated/marketplace_command_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+func enableServiceAccess(broker helpers.ServiceBroker) {
+	Eventually(helpers.CF("enable-service-access", getServiceName(broker))).Should(Exit(0))
+}
+
+func enableServiceAccessForOrg(broker helpers.ServiceBroker, orgName string) {
+	Eventually(helpers.CF("enable-service-access", getServiceName(broker), "-o", orgName)).Should(Exit(0))
+}
+
+func getServiceName(broker helpers.ServiceBroker) string {
```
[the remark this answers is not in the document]
georgi-lozev: Well, that indirection is not quite hard to follow. It's more or less self explanatory from the func name and it's signature what it's supposed to do.

We don't mind either way, if you insist and it's a common style around the CLI codebase we will modified it, but otherwise we don't see much value of doing this.

CC @nmaslarski

integration/shared/global/marketplace_command_test.go
jenspinney: @georgi-lozev, Yep, makes perfect sense.

integration/shared/isolated/marketplace_command_test.go
```diff
@@ -0,0 +1,336 @@
...
+							})
+						})
+					})
+				})
+			})
+		})
+	})
+})
+
+func createBroker(domain, serviceName, planName string) helpers.ServiceBroker {
```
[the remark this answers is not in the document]
Ema93sh: Thanks @georgi-lozev. We noticed that this functionality is repeated in other parts of the codebase.
A few places that we saw are:
 * https://github.com/cloudfoundry/cli/blob/52a7bffaf9433a1e064def52984115ae0117971c/integration/shared/global/share_service_command_test.go#L129-L132 
* https://github.com/cloudfoundry/cli/blob/52a7bffaf9433a1e064def52984115ae0117971c/integration/shared/global/unshare_service_command_test.go#L126-L129

It will great if we could move this into a helper for code reusability and reduce duplication.  It's not required for this PR but it would be great in a future PR.

integration/shared/isolated/marketplace_command_test.go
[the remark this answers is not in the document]
Ema93sh: Sounds good if its add to code readability.

georgi-lozev:
Thank you very much @jenspinney, @Ema93sh.

integration/shared/isolated/marketplace_command_test.go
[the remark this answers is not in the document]
georgi-lozev: Here is the [chore](https://www.pivotaltracker.com/story/show/161672768) in our backlog to follow-up on this, thanks again.

PR #1480, opened 2018-10-16
Use CC extra structure for documentation urls

williammartin:
In past, the CC used to provide the now deprecated `documentation_url`
field in the entity structure of a service instance. At some point, to
align with the OSB API spec's concept of provider metadata, we moved
this kind of information under the `extra` structure and the previous
field started returning nulls. This broke the `cf service
<service_instance>` command, which started providing empty documentation
urls.

This introduces a backwards compatible fix for this behaviour,
supporting both the `documentation_url` under `entity` and
`documentationUrl` under `extra`. There should never be a conflict
between these two (and no CC should provide both since the change was breaking),
but to be careful, we will always favour the old behaviour.


abbyachau:
Hi @williammartin thank you. Is there a min cc version for this fix?

williammartin:
@abbyachau This is a backwards compatible change. If the new CC field is present we will use that, if the old one is present we will favour that. I think the new field has been in use for 3 or 4 years now, we could even remove the old logic (I think it was introduced as part of the refactor of this command unintentionally, whenever it was done, because the old cf/command doesn't support it), but I didn't want to take that decision right now when fixing this bug.

integration/shared/global/service_command_test.go
```diff
@@ -294,7 +294,7 @@ var _ = Describe("service command", func() {
 							Eventually(session).Should(Say("tags:\\s+database, email"))
 							Eventually(session).Should(Say("plan:\\s+%s", servicePlan))
 							Eventually(session).Should(Say("description:\\s+fake service"))
-							Eventually(session).Should(Say("documentation:"))
+							Eventually(session).Should(Say("documentation:\\s+http://documentation\\.url"))
```
jenspinney: Could you please add this to line 346 below as well (and any other place that does an assertion on the documentation URL in this file)? Sorry for not mentioning this in the first review -- only just noticed it on the second glance.

jenspinney: @henryaj @williammartin

georgi-lozev: Thank you @jenspinney for pointing this our, it totally makes sense. I've added it.

PR #1487, opened 2018-10-29
command/internal/help_common_display_v6: add entry for app delete

johncornish:
## What Need Does It Address?

I found it slightly confusing to have the cli help text not explicitly show the app `delete` command

New Pivots or people who are just starting to learn how to use the `cf` command. I did not intuitively think to type `cf delete` to remove my `cf push`ed app

Often

## Possible Drawbacks

It may not work; I was unsure where to make the change and was unable to verify that it worked after build

It's in the core help text

## Description of the Change

Show `delete,d` in the help text for `cf`, which is hopefully as simple as including it in the `CommonHelpCategoryList` definition for the `Application lifecycle` category

## Alternate Designs

None

## How Urgent Is The Change?

Not urgent

### DIFF
--- a/command/common/internal/help_common_display_v6.go
+++ b/command/common/internal/help_common_display_v6.go
@@ -17,7 +17,7 @@ var CommonHelpCategoryList = []HelpCategory{
 			{"apps", "run-task", "events"},
 			{"push", "logs", "set-env"},
 			{"start", "ssh", "create-app-manifest"},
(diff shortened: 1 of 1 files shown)


abbyachau:
Thank you @johncornish, much appreciate your help here.

jenspinney:
Thanks @johncornish!

PR #1489, opened 2018-10-31
Add shortname for unset-env command

andrebrowne:
Both

## What Need Does It Address?
Provides a consistent system of aliases for commonly used command groups that set and unset via the CLI

Any user that uses environment variables with cloud foundry

Hopefully in place of `cf unset-env APP_NAME ENV_NAME`

## Possible Drawbacks
None

No new functionality was added, just a short name alias

## Description of the Change
Adds the ability to use `cf ue` instead of `cf unset-env`

## Alternate Designs
Considered `cf use APP_NAME ENV_NAME` but "use" could conflict with future commands and from a language standpoint could be confusing as an alias for unsetting an environment variable 

## Applicable Issues
None

## How Urgent Is The Change?
Not at all

## Other Relevant Parties
N/A


abbyachau:
Hi @andrebrowne thanks; sorry about the delay. We want to make sure we have a convention for this for `unset` commands before we merge this in. Thanks for your patience on this.

integration/v6/isolated/unset_env_command_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"code.cloudfoundry.org/cli/integration/helpers"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("unset-env command", func() {
+	var (
+		orgName    string
```
Ema93sh: Looks like you dont need the `var` and the `BeforeEach` blocks as they are not used for the test.

command/v6/unset_env_command.go
```diff
@@ -2,12 +2,14 @@ package v6
...
 import (
 	"code.cloudfoundry.org/cli/command"
+	"code.cloudfoundry.org/cli/command/flag"
 	"code.cloudfoundry.org/cli/command/translatableerror"
 )
 
 type UnsetEnvCommand struct {
-	usage           interface{} `usage:"CF_NAME unset-env APP_NAME ENV_VAR_NAME"`
-	relatedCommands interface{} `related_commands:"apps, env, restart, set-staging-environment-variable-group, set-running-environment-variable-group"`
+	RequiredArgs    flag.UnsetEnvironmentArgs `positional-args:"yes"`
```
Ema93sh: The `RequiredArgs` is not needed for this pull request but we can add it once we refactor the command.

andrebrowne:
Making the recommended edits. Had to put the cf cli environment on a new machine.

andrebrowne:
@Ema93sh All requested changes completed. Please squash and merge at your convenience.

tjvman:
We ran into some trouble when trying to squash the commits together and found it simpler to create a patch from the diff and apply the patch to master and commit it as you. Also removed the integration test from global as it was duplicating the functionality of the shared tests and then fixed a minor assertion error in one of the tests.

Closed by: 6f0ffb1fb7f6c0ae035841532038038397a909b7

tjvman:
Thanks for your contribution! The changes have been pulled into master, so I'll close this pr.

PR #1496, opened 2018-11-08
Update Gomega to latest sha

AartiKriplani:
In recent Go releases, go test is internally calling go vet and the
version of Gomega currently used in the CLI fails vetting on gbytes.Say
matchers that contain `%` characters.

This is the PR that fixes it in Gomega: https://github.com/onsi/gomega/pull/300

Signed-off-by: William Martin <wmartin@pivotal.io>
Co-authored-by: William Martin <wmartin@pivotal.io>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -400,7 +400,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:9f7277d03f6ef663bef62532e3aa4aa81b027d5bcf104c05f1fe2a6885c99bbd"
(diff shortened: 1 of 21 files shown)


georgi-lozev:
Is this an issue when using Go `1.10.x` as well ?

williammartin:
No, although the vetter has been executed since `1.10.x`, only in `1.11` did it become stricter: https://golang.org/doc/go1.11#vet

jenspinney:
@AartiKriplani @williammartin Thanks! We were just looking into upgrading to Go 1.11, so this comes at the perfect time :D

PR #1500, opened 2018-11-13
Default route fix in v3-push and v3-zdt-push

ewrenn8:

v6

## What Need Does It Address?

Fixes a bug which caused `v3-push` and `v3-zdt-push` to map the default route to the app regardless of if there are already routes mapped. This behavior was not present in `cf push`.

## Description of the Change

CreateAndMapDefaultRoute in the pushaction layer checks to see if routes have already been mapped to the app before creating and mapping the default route. If there are already routes, the default create and map action does not take place.

## Applicable Issues

https://www.pivotaltracker.com/story/show/161832651

## Other Relevant Parties

@matt-royal @zrob


abbyachau:
@Gerg @mkenyon FYI

tjvman:
Merged. Just as a heads up, there was one (very minor) `go fmt` change that was missing; we made the change locally after verifying the PR.

PR #1502, opened 2018-11-16
Enable-service-access refactor

AartiKriplani:
This commit refactors the `cf enable-service-access` command to the new
style, and adds integration tests.

The changes are detailed in full here:
https://www.pivotaltracker.com/story/show/161228421

[finishes #161228421]
@ablease @nmaslarski @henryaj 

Co-authored-by: Alex Blease <ablease@pivotal.io>
Co-authored-by: Niki Maslarski <nikolay.maslarski@sap.com>
Co-authored-by: Aarti Kriplani <akriplani@pivotal.io>
Signed-off-by: Aarti Kriplani <akriplani@pivotal.io>


AartiKriplani:
@jenspinney @tjvman : Would appreciate your review on the PR, thanks!

abbyachau:
Hey @AartiKriplani many thanks for this. We'll review and get back to you soon as we can.

actor/actionerror/service_plan_not_found_error.go
```diff
@@ -0,0 +1,12 @@
+package actionerror
+
+import "fmt"
+
+type ServicePlanNotFoundError struct {
```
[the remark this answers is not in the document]
ablease: Using this example, we created a version of ServicePlanNotFoundError in the translateableerror package. Thanks Jen!

actor/v2action/service_access.go
```diff
@@ -0,0 +1,223 @@
...
+		Values:   []string{services[0].GUID},
+	})
+	allWarnings = append(allWarnings, warnings...)
+	if err != nil {
+		return allWarnings, err
+	}
+
+	for _, plan := range servicePlans {
+		if plan.Name == servicePlan {
+			warnings, err := actor.removeOrgLevelServicePlanVisibilities(plan.GUID)
```
Ema93sh: It would be great if you add the comments from the legacy code. It was difficult for us to understand why we need to delete plan visibilities without this comment.
https://github.com/cloudfoundry/cli/blob/4445fd548dea41993dc4b9f474830f1ce7aa2d91/cf/actors/services_plans.go#L180-L181

actor/actionerror/service_plan_not_found_error.go
```diff
@@ -0,0 +1,12 @@
...
+	PlanName        string
+	ServiceName     string
+}
+
+func (e ServicePlanNotFoundError) Error() string {
+	return fmt.Sprintf("The plan '%s' could not be found for service %s", e.PlanName, e.ServiceName)
```
Ema93sh: Is the inconsistent quote between the plan name and service name to match the old code? or can this be fixed?

nmaslarski: I think it's a mistake. The quotes are not present neither in the old code, nor in the translated error. I'll remove them

actor/v2action/service_access.go
nmaslarski: Fine with me.
Even with the comment I don't really get why we do it.
Looks to me more like a CC problem, not CLI. However we kept the behaviour in order to be consistent with the old command.

actor/v2action/service_access.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			return allWarnings, err
+		}
+	}
+
+	return allWarnings, nil
+}
+
+// EnablePlanForAllOrgs enables access to a specific plan of the given service in all orgs.
+func (actor Actor) EnablePlanForAllOrgs(serviceName, servicePlan string) (Warnings, error) {
```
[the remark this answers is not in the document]
nmaslarski: Done

actor/v2action/service_access.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	if len(services) == 0 {
+		return allWarnings, actionerror.ServiceNotFoundError{Name: serviceName}
+	}
+
+	servicePlans, warnings, err := actor.CloudControllerClient.GetServicePlans(ccv2.Filter{
+		Type:     constant.ServiceGUIDFilter,
+		Operator: constant.EqualOperator,
+		Values:   []string{services[0].GUID},
```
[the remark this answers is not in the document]
georgi-lozev: Hey @jenspinney that all makes sense.

It's definitely a case for new work to be more conscious about receiving multiple results when one is expected.
However, with the current behaviour already in place for quite some time, I don't see how we could make this a less of a breaking change.

Does it sound feasible to try to release the work that should fail in case of duplicate service name and the `enable-service-access` refactoring in the same CLI version bump?

This won't make the things better, but at least won't have the nondeterministic legacy behaviour as part of a official CLI release with the command refactored.

actor/v2action/service_access.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if err != nil {
+		return allWarnings, err
+	}
+
+	if len(services) == 0 {
+		return nil, actionerror.ServiceNotFoundError{Name: serviceName}
+	}
+
+	orgs, warnings, err := actor.CloudControllerClient.GetOrganizations(ccv2.Filter{
```
Ema93sh: I think it would be better if you use `actor.GetOrganizationByName(orgName)` instead of  relying on `actor.CloudControllerClient.GetOrganziations(...)`. 

* It would remove some code duplication. Don't have to check `len(orgs)`
* The actor method also throws an exception if multiple organization with the same name are found (MultipleOrganizationsFoundError). This scenario is not possible via cf create-org but there might be other ways of creating this scenario (suppose apps-manager accidentally allows this possibility).

actor/v2action/service_access.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}
+
+	return allWarnings, nil
+}
+
+// EnablePlanForOrg enables access to a specific plan of the given service in a specific org.
+func (actor Actor) EnablePlanForOrg(serviceName, servicePlanName, orgName string) (Warnings, error) {
+	var allWarnings Warnings
+	services, warnings, err := actor.CloudControllerClient.GetServices(ccv2.Filter{
```
Ema93sh: This block of code https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L158-L170
seems to be repeated in 3 other places:
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L15
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L56
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L103

Looks like this could be abstracted into `GetServiceByName` function. It could be either private (if you do not believe it would be used outside the actor layer) or public. 
Doing this would reduce code duplication and increase code reusability.

actor/v2action/service_access.go
```diff
@@ -0,0 +1,225 @@
...
+		Type:     constant.NameFilter,
+		Operator: constant.EqualOperator,
+		Values:   []string{orgName},
+	})
+
+	if len(orgs) == 0 {
+		return nil, actionerror.OrganizationNotFoundError{Name: orgName}
+	}
+
+	servicePlans, warnings, err := actor.CloudControllerClient.GetServicePlans(ccv2.Filter{
```
Ema93sh: This block of code:
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L182-L191
is duplicated in:
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L30
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L70
https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_access.go#L133

This could be abstracted as `GetServicePlansForService` and  could belong to [service_plan.go](https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_plan.go)

You could also use https://github.com/cloudfoundry/cli/blob/267b4ec10bf5034d2d09ad7f4fd48d81008b015a/actor/v2action/service_summary.go#L108 by making it public. But my personal preference is the former.

actor/v2action/service_access.go
```diff
@@ -0,0 +1,125 @@
...
+
+	for _, plan := range servicePlans {
+		if plan.Name == servicePlanName {
+			_, warnings, err := actor.CloudControllerClient.CreateServicePlanVisibility(plan.GUID, org.GUID)
+			allWarnings = append(allWarnings, warnings...)
+			return allWarnings, err
+		}
+	}
+
+	return nil, fmt.Errorf("Service plan '%s' not found", servicePlanName)
```
williammartin: This is an unfortunate error message that is different from line 52. This is legacy behaviour on the CLI which we'd love to fix. See below for examples of how to reproduce:

```
→ cf service-access
Getting service access as admin...
broker: overview-broker
   service            plan      access   orgs
   overview-broker    simple    all

→ cf enable-service-access overview-broker -p lol
Enabling access of plan lol for service overview-broker as admin...
FAILED
The plan lol could not be found for service overview-broker

→ cf enable-service-access overview-broker -p lol -o acceptance
Enabling access to plan lol of service overview-broker for org acceptance as admin...
FAILED
Service plan lol not found
```

The former message is more descriptive. Perhaps we could have a separate story to fix this up. @abbyachau

actor/v2action/service_access.go
jenspinney: "Looks to me more like a CC problem, not CLI. However we kept the behaviour in order to be consistent with the old command." <-- I agree.

actor/v2action/service_access.go
jenspinney: Thanks for the explanation @williammartin. Abby is out this week, but I think it sounds reasonable to have a small follow-up story to fix this.

api/cloudcontroller/ccv2/service_plan_test.go
```diff
@@ -208,4 +208,93 @@ var _ = Describe("Service Plan", func() {
 			})
 		})
 	})
+
+	Describe("UpdateServicePlan", func() {
+		When("the service plan exists", func() {
+			var (
+				expectedRequestBody map[string]interface{}
+				warnings            Warnings
+				err                 error
```
jenspinney: Could you rename `err` to `executeErr`? We've been bitten in the past by accidentally shadowing variables called `err` that are scoped beyond the immediate block.

api/cloudcontroller/ccv2/service_plan_test.go
```diff
@@ -208,4 +208,93 @@ var _ = Describe("Service Plan", func() {
...
+				public              bool
```
jenspinney: I know that there is no test pollution possibility in the way you have the contexts now, but I'm worried about what happens when someone adds a context inside here in the future and they don't realize they need to initialize `public` and `expectedRequestBody`. Imagine that those values aren't related to their testing goals at all. Their new test may pass when they run it, but they don't realize they've introduced a flakey test.

Maybe you could move the BeforeEach block on line 243 up to this level?

api/cloudcontroller/ccv2/service_plan_test.go
```diff
@@ -208,4 +208,93 @@ var _ = Describe("Service Plan", func() {
...
+					public = false
+				})
+
+				It("updates the service plan and return warnings", func() {
+					Expect(err).NotTo(HaveOccurred())
+					Expect(warnings).To(ConsistOf(Warnings{"this is a warning"}))
+				})
+			})
+		})
+		When("the server returns an error", func() {
```
jenspinney: Teeny tiny nitpick: Please add a newline before the `When`.

api/cloudcontroller/ccv2/service_plan_visibility_test.go
```diff
@@ -131,7 +131,145 @@ var _ = Describe("Service Plan Visibility", func() {
...
+
+				server.AppendHandlers(
+					CombineHandlers(
+						VerifyRequest(http.MethodDelete, "/v2/service_plan_visibilities/service-plan-visibility-guid"),
+						RespondWith(http.StatusNoContent, response, http.Header{"X-Cf-Warnings": {"warning-1, warning-2"}}),
+					),
+				)
+			})
+
+			It("returns the created service plan visibility", func() {
```
jenspinney: This `It` block description looks incorrect.

api/cloudcontroller/ccv2/service_plan_visibility_test.go
```diff
@@ -131,7 +131,145 @@ var _ = Describe("Service Plan Visibility", func() {
...
+				Expect(server.ReceivedRequests()).Should(HaveLen(2))
```
jenspinney: Why 2?

command/v6/enable_service_access_command_test.go
```diff
@@ -0,0 +1,246 @@
...
+			extraArgs = []string{"some-extra-arg"}
+		})
+
+		It("fails with a TooManyArgumentsError", func() {
+			Expect(executeErr).To(MatchError(translatableerror.TooManyArgumentsError{
+				ExtraArgument: "some-extra-arg",
+			}))
+		})
+	})
+	When("a cloud controller API endpoint is set", func() {
```
jenspinney: Nit (feel free to ignore): missing newline

command/v6/enable_service_access_command_test.go
```diff
@@ -0,0 +1,246 @@
...
+				Expect(executeErr).To(MatchError(actionerror.NotLoggedInError{BinaryName: binaryName}))
+			})
+		})
+
+		When("the user is logged in", func() {
+			BeforeEach(func() {
+				fakeConfig.CurrentUserReturns(
+					configv3.User{Name: "admin"},
+					nil)
+				fakeSharedActor.IsLoggedInReturns(true)
```
jenspinney: I might be missing something or misunderstanding this, but I'm having trouble following how IsLoggedIn gets called in this test. The command-layer code under test calls `CheckTarget` and the real shared actor implementation of `CheckTarget` calls `IsLoggedIn`. But since we're using a fake shared actor, I don't see how we would be calling `IsLoggedIn` at all.

command/v6/enable_service_access_command_test.go
```diff
@@ -0,0 +1,246 @@
...
+				})
+
+				It("passes on the service", func() {
+					Expect(fakeActor.EnableServiceForAllOrgsCallCount()).To(Equal(1))
+					service := fakeActor.EnableServiceForAllOrgsArgsForCall(0)
+					Expect(service).To(Equal("some-service"))
+				})
+
+				It("displays informative success message", func() {
+					Expect(testUI.Out).To(Say("Enabling access to all plans of service some-service for all orgs as admin\\.\\.\\."))
```
jenspinney: FYI, you can use [regexp.QuoteMeta](https://golang.org/pkg/regexp/#QuoteMeta) to not need to escape those `.` characters. But this is such minor escaping (and we escape dots like this in plenty of places), so I don't have a problem with this staying as is.

jenspinney: (And in this case since the characters needed to write `regexp.QuoteMeta` is more than the extra characters needed for the escaping, it feels especially unnecessary.)

integration/shared/isolated/enable_service_access_command_test.go
```diff
@@ -0,0 +1,321 @@
...
+
+	When("two services arguments are provided", func() {
+		It("displays an error, and exits 1", func() {
+			session := helpers.CF("enable-service-access", "a-service", "another-service")
+			Eventually(session).Should(Say("FAILED"))
+			Eventually(session.Err).Should(Say(`Incorrect Usage: unexpected argument "another-service"`))
+			Eventually(session).Should(Say("NAME:"))
+			Eventually(session).Should(Say("\\s+enable-service-access - Enable access to a service or service plan for one or all orgs"))
+			Eventually(session).Should(Say("USAGE:"))
+			Eventually(session).Should(Say("\\s+cf enable-service-access SERVICE \\[-p PLAN\\] \\[-o ORG\\]"))
```
jenspinney: This might be a good candidate for regexp.QuoteMeta. Not merge-blocking though.

jenspinney: Oh, just noticed you have a `\s+` at the beginning, so QuoteMeta won't fit if that's needed. Well, you could remove one level of `\` by using \` instead of `"`, but I don't feel strongly about this.

actor/v2action/service_access.go
georgi-lozev: Here is the follow-up story https://www.pivotaltracker.com/story/show/162148852

api/cloudcontroller/ccv2/service_plan_visibility_test.go
georgi-lozev: The client makes `GET` to `/v2/info` first and then the actual `DELETE`.

We already make the verification that we want as part of the server handler, so I guess this expectation is a bit redundant.

command/v6/enable_service_access_command_test.go
georgi-lozev: You're right, we don't. We'll remove it.

command/v6/enable_service_access_command_test.go
georgi-lozev: Can we add lint rule for that?

georgi-lozev:
Hey @jenspinney, @Ema93sh Thank you very much for the review and the comments.

We've incorporated most of the feedback and left comments here and there.

actor/v2action/service_access.go
mattmcneeney: @abbyachau let us know if you're happy for us to make this follow up change. We have blocked the story @georgi-lozev linked to on your response.

abbyachau: hey @mattmcneeney @williammartin sorry I'm just seeing this because of the tracker ping (not getting the github notifications for these comments). I believe `Service plan lol not found` is the convention we try to follow for these types of error but I'm fine with changing it.

PR #1507, opened 2018-11-26
Handle ServiceBrokerCatalogInvalid errors in CLI

henryaj:
Handle invalid service broker catalog 502 errors and present them as known CLI errors. (For all other HTTP 502 errors, we continue to present a V2UnexpectedError.)

This change improves error presentation when CC returns a 502 due to an invalid service broker catalog. Without this change, in the refactored code we would be returning a V2UnexpectedError, which prints to the screen as a whole lot of JSON - instead, we want to display a better-formatted error. For more context, please see @mattmcneeney's comment [here](https://www.pivotaltracker.com/story/show/160751304/comments/196480234).

https://www.pivotaltracker.com/story/show/160751304

cc @georgi-lozev @ablease

### DIFF
--- a/api/cloudcontroller/ccv2/errors_test.go
+++ b/api/cloudcontroller/ccv2/errors_test.go
@@ -16,6 +16,7 @@ var _ = Describe("Error Wrapper", func() {
 	var (
 		serverResponse     string
 		serverResponseCode int
+		executeErr error
 
 		client *Client
 	)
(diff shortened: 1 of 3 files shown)


abbyachau:
Hey @henryaj thanks for this, appreciate it. We are releasing CLI `v6.41` soon so this pr might take a few extra days to be acted upon since we are locking down master until after the release. 

(cc @mattmcneeney @williammartin)

api/cloudcontroller/ccv2/errors_test.go
```diff
@@ -16,6 +16,8 @@ var _ = Describe("Error Wrapper", func() {
 	var (
 		serverResponse     string
 		serverResponseCode int
+		requestMethod      string
```
jenspinney: It seems like you're setting this up so that the tests can serve as more accurate documentation and simulate the real world more closely. However, it feels odd to have this "real-world-ness" just for the service broker endpoint, but all other errors in this file continue to just call `GET /v2/apps` with `client.GetApplications()`. If we want to clarify that the service broker error should generally only occur when you hit the endpoint for creating service brokers, then why not do the same for the space-creation failures, buildpack not found errors, etc.? Just feels weird that only service broker creation follows this style and the rest just rely on the fact that the error wrapper is super generic and doesn't care what request was made.

api/cloudcontroller/ccv2/errors_test.go
```diff
@@ -327,22 +347,65 @@ var _ = Describe("Error Wrapper", func() {
...
+							requestMethod = http.MethodPost
+							requestPath = "/v2/service_brokers"
+							serverResponse = `{
+								"description": "Service broker catalog is invalid: \nService overview-service\n  Service dashboard client id must be unique\n",
+								"error_code": "CF-ServiceBrokerCatalogInvalid",
+								"code": 270012
+							}`
+						})
+
+						It("returns an ServiceBrokerCatalogInvalid", func() {
```
jenspinney: Super nitpick (since we have other comments anyway): ServiceBrokerCatalogInvalid -> ServiceBrokerCatalogInvalidError

jenspinney:
Looks good. Just small comments/questions. We also can't merge at the moment because of what @abbyachau said, but that lock should be lifted soon.

Thanks for the PR!
Jen & @Ema93sh

api/cloudcontroller/ccv2/errors_test.go
henryaj: Thanks. We've moved the call to `client.GetApplications()` to a JustBeforeEach, and added a comment to explain that we're stubbing the call so the actual endpoint really matter.

henryaj:
Thanks for the feedback, Jen - just going through our pipeline now. We've updated the PR description to make it clearer.

henryaj:
(Not ready to merge; this is still going through our CI)

henryaj:
Our CI integration tests are failing for unrelated reasons [https://ci.sapi.cf-app.com/teams/main/pipelines/spikecline/jobs/run-integration-tests/builds/57]; we think this is probably fine to merge.

PR #1511, opened 2018-11-28
Fix racy ghttp server in ccv2 unit tests

### LINKED ISSUE #1510: cloud controller v2 unit tests are racy
We recently had a couple of failures of a couple of tests in the ccv2 suite:

https://ci.sapi.cf-app.com/teams/main/pipelines/spikecline/jobs/cli-unit-tests/builds/64
https://ci.sapi.cf-app.com/teams/main/pipelines/spikecline/jobs/cli-unit-tests/builds/57

Example snippet below if you don't have log access:

```
• Failure in Spec Setup (BeforeEach) [0.088 seconds]
Application Instance Status [BeforeEach] GetApplicationApplicationInstanceStatuses when the client returns an error returns the error and warnings 
/tmp/build/695050a2/gopath/src/code.cloudfoundry.org/cli/api/cloudcontroller/ccv2/application_instance_status_test.go:143

  Method mismatch
  Expected
      <string>: PUT
  to equal
      <string>: GET

  /tmp/build/695050a2/gopath/src/code.cloudfoundry.org/cli/vendor/github.com/onsi/gomega/ghttp/handlers.go:35
```

A quick look through the test set up leads me to believe that this suite has a bunch of race conditions around the shared server. 

In a `SynchronizedBeforeSuite`, we set up one `ghttp.Server`: https://github.com/cloudfoundry/cli/blob/master/api/cloudcontroller/ccv2/cloudcontrollerv2_suite_test.go#L29

In a `BeforeEach`, we reset the state of the Server: https://github.com/cloudfoundry/cli/blob/master/api/cloudcontroller/ccv2/cloudcontrollerv2_suite_test.go#L40

However, this is inherently racy because the server is being used in parallel by multiple nodes, which can interleave `VerifyReqest`s with incoming http requests from different tests. 

In the case above, we verify a `GET`: https://github.com/cloudfoundry/cli/blob/master/api/cloudcontroller/ccv2/application_instance_status_test.go#L137

But there is almost certainly some other test running in parallel that is issuing a `PUT`.

This (or similar failures) were quite easily reproducible by navigating to the `ccv2` package and running:

```
ginkgo -nodes 16 -untilItFails
```

It took maybe up to 50 attempts to fail reliably.

**Suggestion**

The cost of setting up `ghttp` server is almost negligble. I noticed practically no difference in the test run time by moving the `SynchronizedBeforeSuite` to `BeforeEach` and symmetrically on the AfterEach, and running the above command didn't fail in several hundred runs.

williammartin:
Previously, there was a single server being set up in a
SynchronizedBeforeSuite block to be shared between all tests, with an
attempt to clean state via `server.Reset()`. However, when the tests are
run in parallel, it is still possible to have a sequence where server
verification assertions interleave with incoming http requests from
concurrent tests.

The cost of starting a server appears to be negligble, so this commit
ensures there is an entirely new server set up for each spec that runs.

[fixes #1510]

### DIFF
--- a/api/cloudcontroller/ccv2/cloudcontrollerv2_suite_test.go
+++ b/api/cloudcontroller/ccv2/cloudcontrollerv2_suite_test.go
@@ -23,21 +23,15 @@ func TestCloudcontrollerv2(t *testing.T) {
 
 var server *Server
 
-var _ = SynchronizedBeforeSuite(func() []byte {
-	return []byte{}
-}, func(data []byte) {
+var _ = BeforeEach(func() {
 	server = NewTLSServer()
 
 	// Suppresses ginkgo server logs
 	server.HTTPTestServer.Config.ErrorLog = log.New(&bytes.Buffer{}, "", 0)
 })
 
-var _ = SynchronizedAfterSuite(func() {
+var _ = AfterEach(func() {
 	server.Close()
-}, func() {})
-
-var _ = BeforeEach(func() {
-	server.Reset()
 })
 
 func NewTestClient(passed ...Config) *Client {


williammartin:
I did not run this through the SAPI pipeline. I created it as a reference for what I described as a suggested fix in #1510. Will leave it up to you to accept as is or figure out another solution.

abbyachau:
Hi @williammartin many thanks for this. The team is trying to release `6.41` so they've locked down `master`. We'll review and get back to you soon. Thanks.

tjvman:
Closing this PR in favor of https://github.com/cloudfoundry/cli/issues/1510 at the @williammartin's recommendation.

Issue #1506, opened 2018-11-25
CF cli not displaying route service url on "cf service <service name>"


jeffleecoffee:
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
I was attempting to debug my user-provided-service route. However, I found that I could not see which route_service_url was already set.
![result](https://user-images.githubusercontent.com/14160064/48977543-0f080080-f051-11e8-8d7a-a2b4d571baa2.png)

**Expected behavior**
the CLI should display the route_service_url

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. Run `cf cups authz -r <some https://URL>`
2. Bind a route-service `cf bind-route-service <domain> authz --hostname <host url>`
3. 'cf service authz' route_service_url should be <some https://URL>

**Provide more context**
- I am running Ubuntu 18.04.1 LTS
- cf version 6.40.1+85d04488a.2018-10-31

georgi-lozev:
Hey @jeffleecoffee thank you for the issues. We've dragged this to our Icebox and it should get prioritized soon.

willmurphyscode:
Also @jeffleecoffee if you pass `-v` you can see the `route_service_url` in the JSON that gets printed. I know reading JSON is no fun, but it might help you in the meantime to at least be able to access the information you need.

@georgi-lozev It looks like CC API already returns the information we need, but we're just not displaying it for some reason, so this story shouldn't be a heavy lift.

jeffleecoffee:
@georgi-lozev and @willmurphyscode thank you for the prompt response and the suggestion.

jeffleecoffee:
Hello @georgi-lozev and @willmurphyscode,
I've been looking forward to making a contribution for the CLI, is it cool if I work on this bug as my first commit?

nmaslarski:
Hello @jeffleecoffee,
We'd be delighted if you contribute the fix.
Please make sure you have  read the [contributors guide](https://github.com/cloudfoundry/cli/blob/master/.github/CONTRIBUTING.md).

laurzca:
Thanks for your contribution @jeffleecoffee! This is now in.

jeffleecoffee:
[...]

PR #1520, opened 2018-12-07
Fix for CF cli not displaying route service url on "cf service"

jeffleecoffee:
CLI v6

## What Need Does It Address?
This can be used to debug issues or view the service route URL of a user created service.
Example: "cf cups test_service -r https://<url>". This URL should be shown for that service's detail "cf service test_service"

This will be to address the bug issued here: https://github.com/cloudfoundry/cli/issues/1506

This will be used specifically for "cf service" on user provided services. If routes are not present then the display will show the field, but no results (similar to tags).

## Possible Drawbacks
No potential drawbacks, other than extra clutter to display the URL.

This should a basic functionality of the service information summary. Otherwise, the end user would be using "cf service <name> -v" to find an essential component of service use.

## Description of the Change
As indicated in the initial Github issue, the Route service URL property was being returned from the cloud controller to the CLI. This commit just exposes the detail for user provided service summaries.

## Alternate Designs
For this addition to the view, due to simplicity, no alternatives were considered.

## Applicable Issues
None.

## How Urgent Is The Change?
Not urgent, as there is a work around to use "cf service <name> -v"

## Other Relevant Parties
None.

### DIFF
--- a/command/v6/service_command.go
+++ b/command/v6/service_command.go
@@ -183,8 +183,10 @@ func (cmd ServiceCommand) displayUserProvidedServiceInstanceSummary(serviceInsta
 	table := [][]string{
 		{cmd.UI.TranslateText("name:"), serviceInstanceSummary.Name},
 		{cmd.UI.TranslateText("service:"), cmd.UI.TranslateText("user-provided")},
-		{cmd.UI.TranslateText("route service url:"), serviceInstanceSummary.RouteServiceURL},
 		{cmd.UI.TranslateText("tags:"), strings.Join(serviceInstanceSummary.Tags, ", ")},
+	}
+	if serviceInstanceSummary.RouteServiceURL != "" {
+		table = append(table, []string{cmd.UI.TranslateText("route service url:"), serviceInstanceSummary.RouteServiceURL})
 	}
 	cmd.UI.DisplayKeyValueTable("", table, 3)
 }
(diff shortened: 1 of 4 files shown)


AartiKriplani:
@jeffleecoffee thanks for the PR, changes look ok but would be good to have some tests. At the least on unit level `service_instance_test.go` to make sure that we're serializing the new field.

jeffleecoffee:
@AartiKriplani thank you for the recommendation and where they can be added. I've added some unit test checks to watch the field.

AartiKriplani:
@jenspinney @tjvman : LGTM, could you please review this PR and merge if it's fine ? Thanks!

command/v6/service_command.go
```diff
@@ -183,6 +183,7 @@ func (cmd ServiceCommand) displayUserProvidedServiceInstanceSummary(serviceInsta
 	table := [][]string{
 		{cmd.UI.TranslateText("name:"), serviceInstanceSummary.Name},
 		{cmd.UI.TranslateText("service:"), cmd.UI.TranslateText("user-provided")},
+		{cmd.UI.TranslateText("route service url:"), serviceInstanceSummary.RouteServiceURL},
```
AartiKriplani: @jeffleecoffee : Would be nice to have a unit test for this in `service_command_test.go` too. Thanks

jeffleecoffee: Good idea, I've made some changes to only print routes when the service has it and also added an extra unit test for service_command_test.

abbyachau:
Hey @AartiKriplani let us know if we are able to merge this pull request - we'll wait on your ok (it looks ok to us) before we do so. Thanks!

AartiKriplani:
LGTM, please go ahead and merge @abbyachau , thanks

jeffleecoffee:
Thanks both @abbyachau and @AartiKriplani

PR #1528, opened 2018-12-17
Add destination space and org to network-policies command

christianang:

v6

## What Need Does It Address?

In a previous PR we added the ability to add/remove cross space network policies: https://github.com/cloudfoundry/cli/pull/1504.

This PR provides parity for the network policies command. It will now show the cross space network policies that exist along with its destination space and destination org.

Users who add cross space network policies.

It is our most common feature request from users.

## Possible Drawbacks

This PR and the previous PR sets a precedent for cross org/space actions.

The network-policies command already exists in the core cli.

## Description of the Change

Design doc: https://docs.google.com/document/d/17FtmNkI5iJ5Dxjgg6ToIefJeCtAEyc6f1rAZv3d5p8I/edit?usp=sharing

## How Urgent Is The Change?

The previous PR to `{add/remove}-network-policies` has already been merged, so this will bring `network-policies` to parity. And we want this to have parity before releasing a new cli.

## Other Relevant Parties

@jsirlin @ameowlia @oozie @tylerschultz


jenspinney:
@christianang Thanks for this PR! Could you rebase this so that it can be merged without conflicts?

XanderStrike:
@jenspinney Rebased! Thanks for letting us know!

XanderStrike:
Amelia says meow.

actor/cfnetworkingaction/policy.go
```diff
@@ -49,73 +52,199 @@ func (actor Actor) AddNetworkPolicy(srcSpaceGUID, srcAppName, destSpaceGUID, des
...
+	}
+
 	var v1Policies []cfnetv1.Policy
-	v1Policies, err = actor.NetworkingClient.ListPolicies()
+	v1Policies, err = actor.NetworkingClient.ListPolicies(srcAppGUIDs...)
+	if err != nil {
+		return []Policy{}, allWarnings, err
+	}
+
+	v1Policies = filterPoliciesWithoutMatchingSourceGUIDs(v1Policies, srcAppGUIDs)
```
jenspinney: It seems like this is doing the same thing as passing `srcAppGUIDs` as a filter on line 67 above. Why is this filtering needed?

actor/cfnetworkingaction/policy.go
```diff
@@ -49,73 +52,199 @@ func (actor Actor) AddNetworkPolicy(srcSpaceGUID, srcAppName, destSpaceGUID, des
...
+
+	destAppGUIDs := uniqueDestGUIDs(v1Policies)
+
+	destApplications, warnings, err := actor.V3Actor.GetApplicationsByGUIDs(destAppGUIDs...)
+	allWarnings = append(allWarnings, Warnings(warnings)...)
```
jenspinney: I believe the `...` makes the conversion to `Warnings` unnecessary.

actor/cfnetworkingaction/policy.go
```diff
@@ -164,17 +293,15 @@ func (actor Actor) RemoveNetworkPolicy(srcSpaceGUID, srcAppName, destSpaceGUID,
...
-	dstName, dstOk := appNameByGuid[v1Policy.Destination.ID]
-	if srcOk && dstOk {
-		return Policy{
-			SourceName:      srcName,
-			DestinationName: dstName,
-			Protocol:        string(v1Policy.Destination.Protocol),
-			StartPort:       v1Policy.Destination.Ports.Start,
-			EndPort:         v1Policy.Destination.Ports.End,
-		}
+func (Actor) transformPolicy(appByGuid map[string]v3action.Application, spaceNamesByGUID, orgNamesBySpaceGUID map[string]string, v1Policy cfnetv1.Policy) Policy {
```
jenspinney: It took me a while to understand that the new code above (the unique-ing and resolving of GUIDs to structs) is just to pre-build a cache so that you don't need to look up the same org/space/app more than once.

I guess I don't know enough about the likely usage scenarios to know if this optimization is worth the extra code complexity. Do we expect users to have many network policies?

If it is likely that users will have many network policies and will feel the performance pain of looking up the same resource multiple times, maybe this could be structured in a more traditionally caching way. Instead of pre-fetching everything, we could inline `transformPolicy` and build up the caches as we go. Then we also wouldn't need to add the `GetApplicationsByGUIDs` and similar methods to the actor. (I think variadic functions in our actor layer isn't super common in this codebase.)

api/cloudcontroller/ccv3/space.go
```diff
@@ -11,6 +11,8 @@ type Space struct {
 	GUID string `json:"guid"`
 	// Name is the name of the space.
 	Name string `json:"name"`
+	// Relationships list the relationships to the application.
```
jenspinney: This comment should probably say "space" instead of "application".

actor/cfnetworkingaction/policy.go
christianang: This filtering is to remove policies without matching source guids. The API that we are using potentially returns policies with destination app guids that match the source app guids depending on version of the networking API you are using.

christianang: We will add a comment here to clarify that ListPolicies returns policies that have the guid in either the source or destination.

actor/cfnetworkingaction/policy.go
christianang: The reason we've added the `GetApplicationsByGUIDs` is because we need to get the applications with only the app guid, we don't know the org or space the application comes from.

The calls for spaces and orgs is necessary because we need to get the space and org names. The API only returns the application's space guid, necessitating multiple calls to get the org name. Effort was made to reduce the number of network calls. We do know that some clients do have a large number of network policies within many spaces/orgs, which would result exponentially more network calls.

Inlining transformPolicy and caching as we go would require a lot more network calls and we do have the information upfront to reduce the number of network calls.

There is potential to de-duplicate and break down large functions. We will take a pass at it.

christianang: DRY'd up the functions some. Let us know if you see other areas that could be improved.

christianang:
Made some updates based on the feedback. Let us know if there is any other areas that can be improved or not.

actor/v3action/application_test.go
```diff
@@ -224,6 +236,74 @@ var _ = Describe("Application Actions", func() {
 		})
 	})
 
+	Describe("GetApplicationsByGUIDs", func() {
```
tjvman: This `Describe` block should have a top-level `JustBeforeEach` that invokes the method under test, and the individual `It` blocks should not be invoking it. If the method under test returns an error, it should be stored in a variable named `executeErr`. See the Describe block on 33 for an example.

The top-level JustBeforeEach is a unit testing convention that we follow so that any given describe block will have a single, predictable place where the method under test is invoked. Storing a returned error as `executeErr` helps avoid the shadowing that tends to occur when the error variable is just named `err`.

actor/v3action/organization_test.go
```diff
@@ -73,6 +73,66 @@ var _ = Describe("Organization Actions", func() {
 		})
 	})
 
+	Describe("GetOrganizationsByGUIDs", func() {
```
tjvman: See comment above in `application_test.go`.

actor/v3action/space_test.go
```diff
@@ -220,4 +221,70 @@ var _ = Describe("Space", func() {
 		})
 
 	})
+
+	Describe("GetSpacesByGUIDs", func() {
```
tjvman: See comment above in `application_test.go`.

integration/shared/isolated/add_network_policy_command_test.go
```diff
@@ -119,15 +112,12 @@ var _ = Describe("add-network-policy command", func() {
 					Eventually(session).Should(Say("OK"))
 					Eventually(session).Should(Exit(0))
 
-					session = helpers.CF("curl", fmt.Sprintf("/networking/v1/external/policies?id=%s", sourceAppGUID))
+					session = helpers.CF("network-policies")
+					Eventually(session).Should(Say(`Listing network policies in org %s / space %s as %s\.\.\.`, sourceOrg, sourceSpace, username))
+					Consistently(session).ShouldNot(Say("OK"))
```
tjvman: I'm not sure I understand what this assertion is for. It doesn't seem like there's ever a scenario where this command will print `OK`, so having a standalone negative assertion feels strange to me.

integration/shared/isolated/network_policies_command_test.go
```diff
@@ -72,8 +72,47 @@ var _ = Describe("network-policies command", func() {
...
+				Eventually(session).Should(Say("OK"))
+				Eventually(session).Should(Exit(0))
+			})
+
+			It("lists the policy", func() {
+				session := helpers.CF("network-policies")
+
+				username, _ := helpers.GetCredentials()
+				Eventually(session).Should(Say(`Listing network policies in org %s / space %s as %s\.\.\.`, orgName, spaceName, username))
+				Consistently(session).ShouldNot(Say("OK"))
```
tjvman: See comment in `add_network_policy_command_test.go`.

integration/shared/isolated/remove_network_policy_command_test.go
```diff
@@ -133,15 +133,12 @@ var _ = Describe("remove-network-policy command", func() {
 					Eventually(session).Should(Say("OK"))
 					Eventually(session).Should(Exit(0))
 
-					session = helpers.CF("curl", fmt.Sprintf("/networking/v1/external/policies?id=%s", sourceAppGUID))
+					session = helpers.CF("network-policies")
+					Eventually(session).Should(Say(`Listing network policies in org %s / space %s as %s\.\.\.`, sourceOrg, sourceSpace, username))
+					Consistently(session).ShouldNot(Say("OK"))
```
tjvman: See comment in `add_network_policy_command_test.go.`

actor/v3action/space_test.go
tylerschultz: We modeled these tests to be consistent with GetSpaceByName, so I'll change that and GetSpacesByGUIDs tests to be in alignment with your preferences.

integration/shared/isolated/add_network_policy_command_test.go
tylerschultz: Agreed, will remove.

integration/shared/isolated/network_policies_command_test.go
tylerschultz: Agree, will remove.

integration/shared/isolated/remove_network_policy_command_test.go
tylerschultz: Agreed, will remove.

tylerschultz:
@tjvman I've updated the tests per your feedback. Thanks.

PR #1529, opened 2018-12-18
Add -b flag to enable-service-access command

nmaslarski:
Since the CC has recently changed to allow the same broker to be
registered multiple times, it is possible to there to be naming
collisions on service offerings.

The -b flag allows the CLI to search for a service by name within the
context of a particular broker, allowing for disambiguation.

More context here: https://www.pivotaltracker.com/story/show/162726522 and in the linked stories.

Best,
@nmaslarski 
on Behalf of SAPI team

### DIFF
--- a/actor/actionerror/duplicate_service_error.go
+++ b/actor/actionerror/duplicate_service_error.go
@@ -1,0 +1,11 @@
+package actionerror
+
+import "fmt"
+
+type DuplicateServiceError struct {
+	Name string
+}
+
+func (e DuplicateServiceError) Error() string {
+	return fmt.Sprintf("Service '%s' is provided by multiple service brokers.", e.Name)
+}
--- a/integration/shared/isolated/enable_service_access_command_test.go
+++ b/integration/shared/isolated/enable_service_access_command_test.go
@@ -1,6 +1,7 @@
 package isolated
 
 import (
+	"code.cloudfoundry.org/cli/api/cloudcontroller/ccversion"
 	"code.cloudfoundry.org/cli/integration/helpers"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -18,6 +19,7 @@ var _ = Describe("enable service access command", func() {
 				Eventually(session).Should(Say("USAGE:"))
 				Eventually(session).Should(Say("\\s+cf enable-service-access SERVICE \\[-p PLAN\\] \\[-o ORG\\]"))
 				Eventually(session).Should(Say("OPTIONS:"))
+				Eventually(session).ShouldNot(Say("\\s+\\-b\\s+Enable access to a service from a specific service broker"))
 				Eventually(session).Should(Say("\\s+\\-o\\s+Enable access for a specified organization"))
 				Eventually(session).Should(Say("\\s+\\-p\\s+Enable access to a specified service plan"))
 				Eventually(session).Should(Say("SEE ALSO:"))
@@ -36,6 +38,7 @@ var _ = Describe("enable service access command", func() {
 			Eventually(session).Should(Say("USAGE:"))
 			Eventually(session).Should(Say("\\s+cf enable-service-access SERVICE \\[-p PLAN\\] \\[-o ORG\\]"))
 			Eventually(session).Should(Say("OPTIONS:"))
+			Eventually(session).ShouldNot(Say("\\s+\\-b\\s+Enable access to a service from a specific service broker"))
 			Eventually(session).Should(Say("\\s+\\-o\\s+Enable access for a specified organization"))
 			Eventually(session).Should(Say("\\s+\\-p\\s+Enable access to a specified service plan"))
 			Eventually(session).Should(Say("SEE ALSO:"))
@@ -54,6 +57,7 @@ var _ = Describe("enable service access command", func() {
 			Eventually(session).Should(Say("USAGE:"))
 			Eventually(session).Should(Say("\\s+cf enable-service-access SERVICE \\[-p PLAN\\] \\[-o ORG\\]"))
 			Eventually(session).Should(Say("OPTIONS:"))
+			Eventually(session).ShouldNot(Say("\\s+\\-b\\s+Enable access to a service from a specific service broker"))
 			Eventually(session).Should(Say("\\s+\\-o\\s+Enable access for a specified organization"))
 			Eventually(session).Should(Say("\\s+\\-p\\s+Enable access to a specified service plan"))
 			Eventually(session).Should(Say("SEE ALSO:"))
@@ -124,12 +128,13 @@ var _ = Describe("enable service access command", func() {
 
 		Context("a service broker is registered", func() {
 			var (
-				orgName     string
-				spaceName   string
-				domain      string
-				service     string
-				servicePlan string
-				broker      helpers.ServiceBroker
+				orgName      string
+				spaceName    string
+				domain       string
+				service      string
+				servicePlan  string
+				broker       helpers.ServiceBroker
+				secondBroker helpers.ServiceBroker
 			)
 
 			BeforeEach(func() {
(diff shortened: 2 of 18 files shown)


actor/actionerror/duplicate_service_error.go
```diff
@@ -0,0 +1,11 @@
+package actionerror
+
+import "fmt"
+
+type DuplicateServiceError struct {
+	Name string
+}
+
+func (e DuplicateServiceError) Error() string {
```
jenspinney: Am I understanding correctly that a command that would have previously succeeded will now throw an error? (in the case where a user currently has two brokers providing the same service) 

I know we've talked about this before, but I want to be really explicit about this. If we decide to make a breaking change, I think we should call out why we think it's the right decision to make.

nmaslarski: Yes it is a breaking change, but we don't completely understand what `call out` means.
Do you think we should send a mail to `cf-dev`, or discuss it more with the cli team, or maybe somehow add it to the release notes?

jenspinney: I guess I mean I would've expected it to be mentioned in the PR description with a discussion of the reasoning behind the decision. But here is fine too. I think whenever we make a breaking change, we should be aware of what the impact will be to people we break and why it's worth doing anyway. I don't want any breaks to slip through without the PM (for example) being aware of it.

Any write up you provide could be integrated into our release notes.

abbyachau: Hi @laurzca @nmaslarski I'm not aware of (or may have missed) a breaking change to support this feature. Please could you point out where this was communicated to us in the feature narrative so that I can track?

laurzca: Hey friends @abbyachau @jenspinney:

1.) We don't think this is a breaking change. Before this feature, it was not possible to register two service brokers exposing services with the same name, so it is very unlikely that a user would have encountered this scenario. 
2.) In the very unlikely situation that they did register two brokers exposing services with the same service name, then `cf create-service` would actually provision a service using seemingly random behavior: users are not asked which service from _which broker_ to provision. The user would struggle to tell which one of the services they had actually provisioned. 
3.) Without this commit, users are going to end up encountering the situation in point 2 above, which defeats the purpose of allowing them to register the same service name from two different brokers. If we view this as one track of work, then we are not introducing a breaking change.

jenspinney: Hi @laurzca and other SAPIans,

I'm numbering my responses to line up with yours :)
1) I don't think it's so unlikely that there are Cloud Foundries in use now where people have registered different brokers with the same service name. This feature was released in CAPI [1.71](https://github.com/cloudfoundry/capi-release/releases/tag/1.71.0) (cf deployment [5.4.0](https://github.com/cloudfoundry/cf-deployment/releases/tag/v5.4.0)). People may have done this without even realizing it (this is especially true when space-scoped services get involved). Even if a breaking change isn't very likely, I think it's better to say "This is a breaking change, but we don't think the pain will be very bad" or "the risk of people hitting it is low" than to say it isn't a breaking change.

2) I wouldn't characterize it as "very unlikely" that people today have multiple brokers providing services with the same name (it's so easy to do, people may not realize they're doing it). But I agree this breaking change for `cf enable-service-access` is very unlikely, particulary because it's a command that isn't used very often. And yes, the existing behavior with the randomness is very strange and undesirable. But fixing that underiable behavior is a breaking change. Imagine someone has a script running in CI that for some (odd?) reason runs `cf enable-service-access`, does some provisioning and other one-off stuff, then calls `cf disable-service-access` at the end of the script. Imagine that outside of CI, someone has now created a broker with a service of the same name. The script may continue to work exactly as before if they get "lucky" with the way the CC database is ordering the service brokers. The script happily continues in CI, with no apparent difference. This PR will cause that script to now fail.

3)  "If we view this as one track of work, then we are not introducing a breaking change." I think it doesn't really matter how we view it; I think what matters for breaking change is whether people will actually feel the pain or not. If someone's day is ruined because of a breaking change, it doesn't really make them feel better if we say it was all one track of work. My point here is just that we should consider the amount of pain someone will feel if they are affected by this and weigh it against the chance of that occuring as well as the benefits of making the change.

Ultimately what I'm really driving at with all this fuss is two things: 1) It's a lot more user-friendly if we can include information about this in the CLI release notes so folks aren't totally lost if they get hit by this and 2) for the future we can often avoid introducing any kind of breaking change by considering the order that we change things in, and I think we have an opportunity to learn from this.

jenspinney: I agree that running into this breaking change is unlikely, and I think we should definltey do it. (What I was disagreeing with above was whether it's likely or not that people today have service brokers with the same name; I do think that's likely.)

For the ordering of work -- if we had first changed the CLI to return an error if more than one matching service is found (instead of just picking the first one) before the loosening was done on the CC, then I think we could have done this without introducing a breaking change.

abbyachau: Hi all, we can't introduce a breaking change to the CLI without bumping to a major version, which we don't plan on doing just to support this feature. I don't have the full picture, it might help if we mapped out the CLI user workflow but it'll be good to discuss a few different options on how to move forward.

williammartin:
I've just rebased this on master, fixed a merge conflict and force pushed the contentious behaviour that we agreed upon back for review. Since we rebased a bunch and I resolved a merge conflict, I also kicked off our CI (https://ci.sapi.cf-app.com/teams/main/pipelines/spikecline). If this is green in an hour or two everything should be good to go. Thanks!

abbyachau:
Hi @laurzca @williammartin if there is only one service with an associated broker name and a user runs this command, what occurs?

williammartin:
@abbyachau Just going to restate your question as I understand. I think you are asking, if I do:

```
cf create-service service plan my-service-name
```

And the `service` is only offered by one broker, what is the behaviour?

There should be no change to this behaviour, you should get an instance of `service`. We only require `-b` for disambiguation if there are two brokers offering a service of the same name.

williammartin:
Looks like this has another merge conflict to fix up now :)

integration/shared/isolated/enable_service_access_command_test.go
```diff
@@ -190,6 +194,65 @@ var _ = Describe("enable service access command", func() {
 				})
 			})
 
+			When("two services with the same name are registered", func() {
```
tjvman: This `When` block should only be run when the targeted API version supports registering multiple services with the same name.

williammartin:
Hi @tjvman - we added that skip version check as requested. (While we were in the code, we also renamed a minimum version constant to refer to multi-service registration, as we feel that more generically describes the feature.)

integration/shared/isolated/marketplace_command_test.go
```diff
@@ -171,9 +171,9 @@ var _ = Describe("marketplace command", func() {
 						})
 					})
 
-					When("cc api version >= 2.125.0", func() {
+					When("CC API returns broker names in response", func() {
 						BeforeEach(func() {
-							helpers.SkipIfVersionLessThan(ccversion.MinVersionServiceBrokerNameV2)
+							helpers.SkipIfVersionAtLeast(ccversion.MinVersionMultiServiceRegistrationV2)
```
Ema93sh: This test currently fails in our legacy environment. I am not sure if this change from `SkipIfVersionLessThan` to `SkipIfVersionAtLeast` was intentional or not. The reason for the test failure is that for version below `2.125.0`, the api does not support broker name (This field `broker.Name` is empty) but the current implementation relies on this being available.

williammartin: This change was just brain not working correctly when thinking about the `When` and the `Skip`. I thought it would read correctly this way but actually flipped the conditions. I've changed the conditions back and I think it should be good now?

williammartin:
Hey folks, I know you've been on holiday and you've been great at reviewing so this is not intended as me chasing, just saying, it would be great if we could get this merged in today to unblock multi-service-registration. It's definitely the highest priority PR on our side.

PR #1531, opened 2018-12-19
Add CreateBroker helper for integration tests

AartiKriplani:
As the [comment](https://github.com/cloudfoundry/cli/pull/1479#discussion_r230159207) of the `cf marketplace` review suggests we could re-use this method in a few places.

More details in [this story](https://www.pivotaltracker.com/story/show/161672768)

Aarti & @ablease ,
On behalf of SAPI-team


AartiKriplani:
Would be nice to merge this first, so we can adapt the other PRs to use this helper

PR #1532, opened 2018-12-19
Extract composite actor for service access

nmaslarski:
During refactoring of the service-access command from v1 to v2, the
function call chain under `GetServiceBrokerSummaries` was calling many
other public and private methods on the actor. This was making testing
extremely painful, as tests became integration-like, with lots of cloud
controller client behaviour being faked far away from the method under
test.

This commit attempts to extract a composite actor, that can take the
existing actor to satisfy required dependencies e.g. fetching
organizations and services with plans, and in tests can be faked
independently.

Best,
@williammartin @AartiKriplani  @nmaslarski 
on Behalf of SAPI Team


actor/v2action/composite/service_broker_summary.go
```diff
@@ -0,0 +1,113 @@
...
+func (w *ServiceBrokerSummaryCompositeActor) GetServiceBrokerSummaries(broker string, service string, organization string) ([]v2action.ServiceBrokerSummary, v2action.Warnings, error) {
+	var filters []ccv2.Filter
+	if broker != "" {
+		filters = append(filters, ccv2.Filter{
+			Type:     constant.NameFilter,
+			Operator: constant.EqualOperator,
+			Values:   []string{broker},
+		})
+	}
+	brokers, ccWarnings, brokersErr := w.CloudControllerClient.GetServiceBrokers(filters...)
```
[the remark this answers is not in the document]
henryaj: Hi @willmurphyscode - this should be resolved as of our latest commit. Let me know if you have any questions.

actor/v2action/composite/service_broker_summary.go
```diff
@@ -0,0 +1,138 @@
+package composite
+
+import (
+	"code.cloudfoundry.org/cli/actor/v2action"
+)
+
+//go:generate counterfeiter . ServiceActor
```
jenspinney: Please add an extra newline after `//go:generate` so that it doesn't show up in GoDoc.

command/v6/service_access_command.go
```diff
@@ -6,14 +6,14 @@ import (
...
 	"code.cloudfoundry.org/cli/actor/v2action"
+	"code.cloudfoundry.org/cli/actor/v2action/composite"
 	"code.cloudfoundry.org/cli/command"
 	"code.cloudfoundry.org/cli/command/translatableerror"
 	"code.cloudfoundry.org/cli/command/v6/shared"
 	"code.cloudfoundry.org/cli/util/sorting"
 )
 
 //go:generate counterfeiter . ServiceAccessActor
-
```
jenspinney: Please undo removal of the newline (for GoDoc).

actor/v2action/composite/service_broker_summary.go
jenspinney: Wow that was quick :)

williammartin: These are not the droids you're looking for

PR #1536, opened 2019-01-03
Handle scenarios of extra org level visibility objects for `enable-service-access` cmd

AartiKriplani:
1. Disable on org level first (i.e. delete the service plan visibility, if any), when enabling service for all orgs
2. Do not enable on org level ((i.e. no new service plan visibility), when the plan is already public

[finishes #162747373]

More details can be found [here].(https://www.pivotaltracker.com/n/projects/2105761/stories/162747373)

-Aarti
On behalf of SAPI team

### DIFF
--- a/actor/v2action/service_access.go
+++ b/actor/v2action/service_access.go
@@ -16,8 +16,14 @@ func (actor Actor) EnableServiceForAllOrgs(serviceName, brokerName string) (Warn
 	}
 
 	for _, plan := range servicePlans {
-		warnings, err := actor.CloudControllerClient.UpdateServicePlan(plan.GUID, true)
+		warnings, err := actor.removeOrgLevelServicePlanVisibilities(plan.GUID, "")
 		allWarnings = append(allWarnings, warnings...)
+		if err != nil {
+			return allWarnings, err
+		}
+
+		ccv2Warnings, err := actor.CloudControllerClient.UpdateServicePlan(plan.GUID, true)
+		allWarnings = append(allWarnings, ccv2Warnings...)
 		if err != nil {
 			return allWarnings, err
 		}
@@ -66,10 +72,12 @@ func (actor Actor) EnableServiceForOrg(serviceName, orgName, brokerName string) 
 	}
 
 	for _, plan := range servicePlans {
-		_, warnings, err := actor.CloudControllerClient.CreateServicePlanVisibility(plan.GUID, org.GUID)
-		allWarnings = append(allWarnings, warnings...)
-		if err != nil {
-			return allWarnings, err
+		if plan.Public != true {
+			_, warnings, err := actor.CloudControllerClient.CreateServicePlanVisibility(plan.GUID, org.GUID)
+			allWarnings = append(allWarnings, warnings...)
+			if err != nil {
+				return allWarnings, err
+			}
 		}
 	}
 
@@ -91,6 +99,9 @@ func (actor Actor) EnablePlanForOrg(serviceName, servicePlanName, orgName, broke
 
 	for _, plan := range servicePlans {
 		if plan.Name == servicePlanName {
+			if plan.Public {
+				return allWarnings, nil
+			}
 			_, warnings, err := actor.CloudControllerClient.CreateServicePlanVisibility(plan.GUID, org.GUID)
 			allWarnings = append(allWarnings, warnings...)
 			return allWarnings, err
--- a/actor/v2action/service_access_test.go
+++ b/actor/v2action/service_access_test.go
@@ -314,6 +314,27 @@ var _ = Describe("Service Access", func() {
 				Expect(enablePlanErr).NotTo(HaveOccurred())
 			})
 
+			When("the plan is already globally enabled", func() {
+				BeforeEach(func() {
+					fakeCloudControllerClient.GetServicePlansReturns(
+						[]ccv2.ServicePlan{
+							{Name: "plan-2", GUID: "service-plan-guid-2", Public: true},
+						},
+						[]string{"foo"},
+						nil)
+				})
+
+				It("should not create org visibility", func() {
+					Expect(enablePlanErr).ToNot(HaveOccurred())
+					Expect(enablePlanWarnings).To(ConsistOf([]string{"foo"}))
+
+					Expect(fakeCloudControllerClient.GetServicesCallCount()).To(Equal(1))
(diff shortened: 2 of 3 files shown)


AartiKriplani:
@abbyachau @laurzca : Any thoughts on this? Do you have historical context on why it's left upto the client to keep the service-plan-visibility in sync and not done in CC?

laurzca:
@zrob @ssisil thoughts on this one? We think it would make sense for CC to handle the business logic instead of the CLI.

laurzca:
We'd like to merge this one as-is, fixing the issue in the CLI. If we decide to handle the logic in the CC, we can do that work independently.

actor/v2action/service_access_test.go
```diff
@@ -625,6 +666,42 @@ var _ = Describe("Service Access", func() {
...
+		When("the service is already enabled for orgs individually", func() {
+			BeforeEach(func() {
+				fakeCloudControllerClient.GetServicePlanVisibilitiesReturns(
+					[]ccv2.ServicePlanVisibility{
+						{GUID: "service-visibility-guid-1"},
+					},
+					nil, nil)
+			})
+
+			It("first disables the plan in both orgs", func() {
```
tjvman: This `It` description confused me - there is no reference to orgs anywhere in the test setup, and the method under test does not retrieve any orgs. It seems like the actual behavior is that for all plans for the target service, we will remove any associated service plan visibilities (the org associated with those visibilities doesn't matter). If this was intended to cover the case where multiple orgs have access to the plan, I would expect to see multiple service plan visibilities being returned.

actor/v2action/service_access.go
```diff
@@ -16,11 +16,17 @@ func (actor Actor) EnableServiceForAllOrgs(serviceName string) (Warnings, error)
 	}
 
 	for _, plan := range servicePlans {
-		warnings, err := actor.CloudControllerClient.UpdateServicePlan(plan.GUID, true)
+		warnings, err := actor.removeOrgLevelServicePlanVisibilities(plan.GUID)
```
tjvman: From what we can see, the legacy command did not perform this removal when making a plan public. What was the impetus for adding this step?

actor/v2action/service_access_test.go
```diff
@@ -495,6 +513,29 @@ var _ = Describe("Service Access", func() {
 				Expect(orgGUID2).To(Equal("org-guid-1"))
 				Expect([]string{planGUID1, planGUID2}).To(ConsistOf([]string{"service-plan-guid-1", "service-plan-guid-2"}))
 			})
+
+			When("service is already enabled for all orgs for a plan", func() {
```
tjvman: This `When` and accompanying `It` have odd descriptions. For one, we're only dealing with a single org here; for two, the test itself shows that a plan visibility *is* being created - for the plan that isn't public. 

Please update these descriptions to more accurately reflect what is being tested.

actor/v2action/service_access_test.go
```diff
@@ -495,6 +513,29 @@ var _ = Describe("Service Access", func() {
...
+				BeforeEach(func() {
+					fakeCloudControllerClient.GetServicePlansReturns(
+						[]ccv2.ServicePlan{
+							{Name: "plan-1", GUID: "service-plan-guid-1", Public: true},
+							{Name: "plan-2", GUID: "service-plan-guid-2", Public: false},
+						},
+						nil,
```
tjvman: Please add test that any warnings returned from the API layer are properly handled.

actor/v2action/service_access_test.go
```diff
@@ -306,6 +306,24 @@ var _ = Describe("Service Access", func() {
 				Expect(enablePlanErr).NotTo(HaveOccurred())
 			})
 
+			When("the plan is already globally enabled", func() {
+				BeforeEach(func() {
+					fakeCloudControllerClient.GetServicePlansReturns(
+						[]ccv2.ServicePlan{
+							{Name: "plan-2", GUID: "service-plan-guid-2", Public: true},
+						},
+						nil, nil)
```
tjvman: Please add test coverage for warnings.

actor/v2action/service_access_test.go
```diff
@@ -625,6 +666,42 @@ var _ = Describe("Service Access", func() {
 			Expect(public).To(BeTrue())
 		})
 
+		When("the service is already enabled for orgs individually", func() {
+			BeforeEach(func() {
+				fakeCloudControllerClient.GetServicePlanVisibilitiesReturns(
+					[]ccv2.ServicePlanVisibility{
+						{GUID: "service-visibility-guid-1"},
+					},
+					nil, nil)
```
tjvman: Please add coverage for warnings to the happy path.

actor/v2action/service_access.go
henryaj: Hi Tom – in the process of doing this work, we noticed that the old servicePlanVisibility objects weren't being swept away when a whole plan was made public. It seemed to us that these org-specific visibilities should be cleaned up in this case.

If this is incorrect behaviour, let us know and we're happy to remove it.

actor/v2action/service_access.go
```diff
@@ -91,6 +99,9 @@ func (actor Actor) EnablePlanForOrg(serviceName, servicePlanName, orgName, broke
 
 	for _, plan := range servicePlans {
 		if plan.Name == servicePlanName {
+			if plan.Public == true {
```
jenspinney: Nit:
```suggestion
			if plan.Public {
```

actor/v2action/service_access.go
jenspinney: I believe the legacy command did do this step. 
https://github.com/cloudfoundry/cli/blob/b6a3695a5db06d6628428067e0a78e63041ea1ca/cf/actors/services_plans.go#L182

So I think this change seems correct and inline with the pre-refactored code.

jenspinney:
"Signed-off-by: Aarti Kriplani <akriplani@pivotal.io>
Co-authored-by: Aarti Kriplani <akriplani@pivotal.io>
Signed-off-by: Aarti Kriplani <akriplani@pivotal.io>
Co-authored-by: Aarti Kriplani <akriplani@pivotal.io>"

@AartiKriplani , we get it.

:P

actor/v2action/service_access.go
jenspinney: @tjvman can you just confirm this resolves your concern when you get a chance?

Issue #1513, opened 2018-11-28
cf services results with a single-service-id error - take 2


julian-hj:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [ ] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
This is a re-submission of the issue reported in #1366 

We see the same issue regularly in our ci pipeline when testing our EFS volume service broker.

It appears that `cf services` is requesting a list of service instances from the API, and then making a set of subsequent API calls to get more details for those instances without appropriate error handling for the case where a service instance was deleted between the two api calls 

**What happened**
`cf services` returns an error:

```
The service instance could not be found: a2560bc8-d7d9-4900-8955-3f43593dbcdf
```

**Expected behavior**
`cf services` should omit the service instance and return `OK` if the service instance got deleted while the `cf services` command was processing

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. register an asynchronous service broker that takes some time to create and delete services.
2. create a service instance
3. wait for it to exist.
4. delete the service instance
5. call `cf -v services` in a tight loop

This is a race condition, so it will only fail if the delete operation finishes while `cf services` is processing, but when we run our test suite, it fails about 25% of the time, which I think translates to something like one out of every 20 service instances.

API dumps are included below.

**Provide more context**
- platform and shell details: Ubuntu 18.04.1 LTS (running in concourse)
- version of the CLI you are running: cf version 6.40.1+85d04488a.2018-10-31
- version of the CC API Release you are on: 2.126.0

Note: As of January 2018, we no longer support API versions older than 2.69.0/3.4.0 (CF Release: 251 / CAPI Release: 1.15.0)

API dump for 2 consecutive invocations of `cf -v services`.  The first one succeeds, but the service is still in the "delete in progress" state.  The second one shows the error.

```
[2018-11-28 04:11:39.45 (UTC)]> cf -v services 
REQUEST: [2018-11-28T04:11:39Z]
GET /v2/info HTTP/1.1
Host: api.54.184.106.177.sslip.io
Accept: application/json
User-Agent: cf/6.40.1+85d04488a.2018-10-31 (go1.10.4; amd64 linux)

RESPONSE: [2018-11-28T04:11:39Z]
[...]

Issue #1353, opened 2018-03-23
cf s command very slow since 6.35


choffm:
Hello,
the `cf s` command is much slower in version 6.35.2, compared to 6.34.1.

## Command  

`cf s`

## What occurred

With version 6.34.1: `cf s  0.18s user 0.04s system 4% cpu 4.647 total`
With version 6.35.2: `cf s  0.26s user 0.10s system 1% cpu 30.341 total`

## CLI Version  

cf version 6.34.1+bbdf81482.2018-01-17
cf version 6.35.2+88a03e995.2018-03-15

## CC API Endpoint Version  

`api version:    2.101.0`

## Platform & Shell Details  

MacOS 10.13.3 (also occured on Windows 10 x64) 

## Any other relevant information  

Also occurred with CF CLI version 6.35.0, however the command resulted in an error message (see https://github.com/cloudfoundry/cli/issues/1342).

XenoPhex:
Yup! In CF 6.35.0, the CF CLI switched from using a single `/spaces/summary` API call to individual `/service`, `/service_instance`, etc. API calls. Due to the increased number of API calls, the `services` request is much slower.

Why did it switch? The `/services/summary` API call breaks when there are more than 50 services or apps. Additionally, the break returns back a successful status code, so the CLI can't tell the difference between 51+ services/apps and 0 services/apps. So it would appear that all your services are gone!

There might be a way to cache some of these requests if they happen to be looking up duplicate objects, but overall the speed increase may be negligible.

abbyachau:
Hi @choffm thanks for creating this issue. We are closing this issue but please do not hesitate to reach out if this issue becomes too problematic. Thanks.

abbyachau:
Reopening because we've received feedback that this command was actually returning the correct number of services/apps if there are more than 50. And the change we made introduced higher latency.

choffm:
Hi,

as @abbyachau  stated, the change which slowed down the `cf s` command wasn't even necessary, but it is a real pain for people who use the cli on a daily basis. In some of my spaces, it takes up to 90 seconds for the command to return the list of services.

Can you please revert the changes or provide another fix for the performance degradation?

Thanks

laurzca:
Hi @choffm thanks for your note! For now, we're planning to fix this (and  `cf m` which can also become quite slow) in V3 of the API and not V2.

laurzca:
[...]

PR #1537, opened 2019-01-07
Ignore 404 on instance fetch for cf services cmd

ablease:
This PR is a bug fix, regarding this issue #1513

CLI v6

## What Need Does It Address?

Prior to this commit, there was a race condition, that service instances
could be listed, and service bindings for that instance could be listed,
but between the two requests, the service instance could be deleted and
the cf services command would fail.

This commit results in logging the error, but continuing with the rest
of the command.

This commit only applies to both managed and user provided service instances.

Thanks, Sapiteam

Alex + @AartiKriplani


davewalter:
Hi there,

Wondering what the status of this PR is. We have started to encounter the same issue in our CI on a fairly regular basis, which is hard to work around to get CI green.

Thanks,
Persi Team

AartiKriplani:
We will be switching back to the summary endpoint as part of this [PR](https://github.com/cloudfoundry/cli/issues/1353). 
We would not have this issue, closing this PR in favor of the other.

PR #1538, opened 2019-01-08
Include broker name in `cf services` output

nmaslarski:
When executing `cf services` command you get the list of services in your org/space.
This PR adds a column showing the service broker name of these services.

More information on https://www.pivotaltracker.com/n/projects/2105761/stories/160752090

Best,
Niki
On behalf of SAPI Team

### DIFF
--- a/integration/shared/isolated/services_command_test.go
+++ b/integration/shared/isolated/services_command_test.go
@@ -1,6 +1,7 @@
 package isolated
 
 import (
+	"code.cloudfoundry.org/cli/api/cloudcontroller/ccversion"
 	"code.cloudfoundry.org/cli/integration/helpers"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


integration/shared/isolated/services_command_test.go
```diff
@@ -123,9 +123,9 @@ var _ = Describe("services command", func() {
 		It("displays all service information", func() {
 			session := helpers.CF("services")
 			Eventually(session).Should(Say("Getting services in org %s / space %s as %s...", orgName, spaceName, userName))
-			Eventually(session).Should(Say(`name\s+service\s+plan\s+bound apps\s+last operation`))
-			Eventually(session).Should(Say(`%s\s+%s\s+%s\s+%s\s+%s`, managedService1, service, servicePlan, appName1, "create succeeded"))
-			Eventually(session).Should(Say(`%s\s+%s\s+%s\s+%s, %s\s+%s`, managedService2, service, servicePlan, appName1, appName2, "create succeeded"))
+			Eventually(session).Should(Say(`name\s+service\s+plan\s+bound apps\s+last operation\s+broker`))
```
willmurphyscode: These tests will fail when run against our legacy environment. Please conditionally skip them based on the CC API version when `service_broker_name` was added to the JSON payload.

The current behavior is that when the service broker name is unavailable, the column is displayed but its value is always empty. Please also add a test that only runs on the legacy environment that asserts that the command succeeds and displays the empty column.

willmurphyscode: Please add a constant for the lowest version where the broker name will show up here: https://github.com/cloudfoundry/cli/blob/71ba30a7fb3bb3177b49dd6f82c943db55ea2686/api/cloudcontroller/ccversion/minimum_version.go#L3

ablease: Hey @willmurphyscode. We have updated this PR so that we conditionally skip these tests when the CC API version is lower than `ccversion.MinVersionServiceBrokerNameV2`.  Please see this [commit](https://github.com/cloudfoundry/cli/pull/1538/commits/7785447f585923697a128cd8b22e698cf3070c56)

Thanks!

Alex + @nmaslarski

tjvman: These tests are now working against my local legacy environment, so I'll resolve this conversation.

PR #1539, opened 2019-01-08
Include broker name in `cf service <service-instance-name>`

nmaslarski:
`cf service <service_name>` command will now also display the name of the broker providing this service

More information on https://www.pivotaltracker.com/n/projects/2105761/stories/160752011

Best,
Niki
On behalf of SAPI Team


williammartin:
Think we've fixed up the note about legacy integration tests and @willmurphyscode has approved. Can we merge?

integration/shared/global/service_command_test.go
```diff
@@ -284,33 +284,75 @@ var _ = Describe("service command", func() {
...
+								Eventually(session).Should(Say("\n\n"))
+								Eventually(session).Should(Say(`name:\s+%s`, serviceInstanceName))
+								Consistently(session).ShouldNot(Say("shared from:"))
+								Eventually(session).Should(Say(`service:\s+%s`, service))
+								Eventually(session).Should(Say(`tags:\s+database, email`))
+								Eventually(session).Should(Say(`plan:\s+%s`, servicePlan))
+								Eventually(session).Should(Say(`description:\s+fake service`))
+								Eventually(session).Should(Say(`documentation:\s+http://documentation\.url`))
+								Eventually(session).Should(Say(`dashboard:\s+http://example\.com`))
+								Eventually(session).Should(Say(`service broker:\s+`))
```
tjvman: This line will cause line 308 to fail - the `\s+` is not restricted to the current line, so the assertion on 306 will consume the wrong newlines. By the time the test reaches 308, the session.Out buffer has been advanced too far and it will not find the correct output.

tjvman:
So there was one test that was failing on legacy - the fix was trivial and legacy envs aren't so easy to spin up on the fly, so we did made the fix and merged. I left a comment on the problem lines if you're curious.

AartiKriplani:
Thank you @tjvman, we'll try to remember this for future tests.

Issue #746, opened 2016-01-29
`cf delete-orphaned-routes` can delete routes which are not orphans


glyn:
This problem could occur when two users are simultaneously deleting orphaned routes and associating (currently orphaned) routes with applications. The problem is that the CLI checks for orphaned routes _and then_ deletes them (just before which they could have been associated with applications).

I believe this cannot be fixed unless [CC issue 527](https://github.com/cloudfoundry/cloud_controller_ng/issues/527) is implemented.

dkoper:
Closing, as I'm not seeing any progress on the CC issue you raised. As you said, not much the CLI can do if the backend doesn't offer it.

glyn:
This is important but not urgent, so please could you leave this issue open to keep it on the books. :-)

I said the same about the CC issue and they have left that issue open.

glyn:
Closing until the CC has fixed their issue.

zrob:
Please see the CC resolution to this in the following story: https://www.pivotaltracker.com/story/show/112675831

dkoper:
@zrob Can you provide a link to the relevant apidoc? Looks like a new API?

zrob:
@dkoper 
http://apidocs.cloudfoundry.org/258/spaces/delete_unmapped_routes.html

dgrizzanti:
Hi All,

I chatted with @abbyachau last week on the CF slack regarding this issue to see if it was still open for contributions. She confirmed that it was so I started taking a look at the implementation. I had a few questions hoping someone from the community can help me out with.

Is the desire here to maintain the existing `cf delete-orphaned-routes` command and update its logic to use the `delete_unmapped_routes` api or replace it with a new `cf delete-unmapped-routes` command? The latter seems more straightforward and gets around the compatibility issues with v6/v7 CLI code, but I wanted to make sure that was ok with those who would be reviewing.

Thanks!

abbyachau:
Hi @dgrizzanti it looks like we wanted to update the logic to use the `delete_unmapped_routes` api - not replace it with a different command. I've tagged this with `dev-review` in case the engs has any additional comments.

abbyachau:
Hi @glyn thanks again for creating this issue. @dgrizzanti provided a [pull request](https://github.com/cloudfoundry/cli/pull/1543) to us which we believe have addressed the issue you've raised. We plan to release this fix soon. Closing this issue as fixed but please do not hesitate to get in touch if you have any questions/feedback.

PR #1543, opened 2019-01-11
Update `delete-orphaned-routes` command to use unmapped routes API

dgrizzanti:
This PR addresses #746 by updating the logic of `delete-orphaned-routes` to use http://apidocs.cloudfoundry.org/258/spaces/delete_unmapped_routes.html instead of listing/deleting as separate steps.

This is my first go at making CLI changes, so feel free to criticize where necessary 🙂

I tried to make sure I covered all of the places that delete orphaned was called that used the list/delete pattern to be replaced with the new API. In the cases of tests, since functionality from the core command was removed (i.e. there is no more listing of the routes and then saying which ones are being deleted, much of that code was removed).


abbyachau:
Thank you @dgrizzanti we'll review and get back to you soon as we can. Appreciate the pull request!

command/v6/delete_orphaned_routes_command.go
```diff
@@ -1,18 +1,16 @@
...
 	"code.cloudfoundry.org/cli/command/v6/shared"
 )
 
-//go:generate counterfeiter . DeleteOrphanedRoutesActor
+//go:generate counterfeiter . DeleteUnmappedRoutesActor
 
-type DeleteOrphanedRoutesActor interface {
-	GetOrphanedRoutesBySpace(spaceGUID string) ([]v2action.Route, v2action.Warnings, error)
-	DeleteRoute(routeGUID string) (v2action.Warnings, error)
+type DeleteUnmappedRoutesActor interface {
```
tjvman: Our naming convention for individual commands' interfaces is `NameOfCommandActor`, so this should be changed back to DeleteOrphanedRoutesActor. 

(The discrepancy between the CLI using "orphaned" and the API using "unmapped" is unfortunate)

command/v6/delete_orphaned_routes_command_test.go
```diff
@@ -149,136 +147,45 @@ var _ = Describe("deleted-orphaned-routes Command", func() {
...
-									Domain: v2action.Domain{
-										Name: "bosh-lite.com",
-									},
-								},
-							}
-
-							fakeActor.GetOrphanedRoutesBySpaceReturns(routes, nil, nil)
 						})
 
 						It("displays getting routes message", func() {
```
tjvman: Please update this test description to reflect changes to the output (it's no longer printing a message about getting routes)

integration/shared/isolated/delete_orphaned_routes_command_test.go
```diff
@@ -57,9 +57,6 @@ var _ = Describe("delete-orphaned-routes command", func() {
 			It("deletes all the orphaned routes", func() {
 				Eventually(helpers.CF("delete-orphaned-routes", "-f")).Should(SatisfyAll(
 					Exit(0),
-					Say("Getting routes as"),
```
tjvman: I'm not seeing any test coverage for the `Deleting routes as <user>` message - please make sure that message is asserted on wherever it's printed.

integration/shared/isolated/delete_orphaned_routes_command_test.go
```diff
@@ -90,10 +87,6 @@ var _ = Describe("delete-orphaned-routes command", func() {
 			It("deletes only the orphaned routes", func() {
 				Eventually(helpers.CF("delete-orphaned-routes", "-f")).Should(SatisfyAll(
 					Exit(0),
-					Say("Getting routes as"),
-					Say(fmt.Sprintf("Deleting route orphan-1.%s/path-1...", domainName)),
-					Say(fmt.Sprintf("Deleting route orphan-2.%s/path-2...", domainName)),
-					Not(Say(fmt.Sprintf("Deleting route bound-1.%s/path-3...", domainName))),
```
tjvman: Please add a call to `cf routes` and assert that only the bound route is still present - otherwise, this test doesn't really cover what its description says it's covering.

dgrizzanti:
@tjvman Thanks for reviewing the PR! I think I've incorporated all of your comments. Originally I had issues running the integration tests but I have a CFDev up and running now so was able to debug and get my updates in properly this time. Let me know if you see anything else amiss.

@abbyachau Is there someone from the team (or maybe the API side) that could answer the question that @glyn brought up?

Thanks all!

abbyachau:
@tcdowney @ssisil please could you review @glyn's comment above, thank you.

tcdowney:
@glyn there isn't any explicit locking going on, but the implementation (which occurred in [CAPI Story #112675831](https://www.pivotaltracker.com/story/show/112675831)) should be atomic. It performs the select and delete in a single SQL query that excludes routes that are in use by either a route mapping or route service. See:

https://github.com/cloudfoundry/cloud_controller_ng/blob/d06178907cfe9f15bed1809142e46a9021170440/app/actions/routing/route_delete.rb#L22

Additionally, there is a foreign key constraint on the `route_mappings.route_guid` column which will prevent the deletion from occurring if somehow it gets to that point (I don't think it will based on the query above).

I think using this endpoint should eliminate the race condition, and there will be logs (prefixed with `'cc.action.route_delete'`) noting any routes it tried, but failed to delete because of foreign key constraints.

PR #1546, opened 2019-01-17
Make `cf services` much faster

henryaj:
`cf services` was much slower after the refactoring due to making individual API calls to `/v2/services`, `/v2/service_plans` etc. See #1353. We've switched back to the old behaviour where we hit a single summary endpoint instead.

See https://www.pivotaltracker.com/story/show/163052467 for context.

cc @ablease @AartiKriplani @nmaslarski


api/cloudcontroller/ccv2/space_summary.go
```diff
@@ -0,0 +1,72 @@
...
+	ServiceBrokerName string
+}
+
+type SpaceSummaryServicePlan struct {
+	GUID    string              `json:"guid"`
+	Name    string              `json:"name"`
+	Service SpaceSummaryService `json:"service"`
+}
+
+type SpaceSummaryServiceInstance struct {
```
tjvman: Please alphabetize these struct fields.

api/cloudcontroller/ccv2/space_summary_test.go
```diff
@@ -0,0 +1,127 @@
...
+						{
+							 "name": "service-instance-name",
+							 "last_operation": {
+									"type": "create",
+									"state": "succeeded",
+									"description": "a description",
+									"updated_at": "some time",
+									"created_at": "some time"
+							 },
+							 "shared_from": null,
```
tjvman: It looks like these two fields (`shared_from`, `shared_to`) are declared and not asserted on in these tests. Do we need these fields then?

api/cloudcontroller/ccv2/space_summary_test.go
```diff
@@ -0,0 +1,127 @@
...
+				 ]
+			}`
+				server.AppendHandlers(
+					CombineHandlers(
+						VerifyRequest(http.MethodGet, "/v2/spaces/some-guid/summary"),
+						RespondWith(http.StatusOK, response, http.Header{"X-Cf-Warnings": {"warning-1"}}),
+					))
+			})
+
+			It("returns the result and all warnings", func() {
```
tjvman: Please add tests for the returned warnings

api/cloudcontroller/ccv2/space_summary_test.go
```diff
@@ -0,0 +1,127 @@
...
+				spaceSummary, _, err := client.GetSpaceSummary("some-guid")
```
tjvman: Please extract the invocation of the method under test into a JustBeforeEach block at the top level of the `Describe("GetSpaceSummary")...` block and use a name other than `err` for the variable that stores the returned error (to avoid shadowing).

command/v6/services_command.go
```diff
@@ -105,3 +109,21 @@ func (cmd ServicesCommand) Execute(args []string) error {
...
+		return sorting.LessIgnoreCase(instanceSummaries[i].Name, instanceSummaries[j].Name)
+	})
+
+	for _, instance := range instanceSummaries {
+		sortBoundApps(instance)
+	}
+}
+
+func sortBoundApps(serviceInstance v2action.ServiceInstanceSummary) {
+	sort.Slice(
```
tjvman: Any particular reason why we use `sort.Slice` here but `sort.SliceStable` above?

command/v6/services_command_test.go
```diff
@@ -171,7 +171,7 @@ var _ = Describe("services Command", func() {
 					)
 				})
 
-				It("displays all the services in the org & space & warnings", func() {
+				It("displays all the services and apps in alphanumeric sorted order together with the org & space & warnings", func() {
```
tjvman: There should be some kind of test ensuring that the sorting described here is actually occurring.

actor/v2action/service_instance_summary.go
```diff
@@ -67,31 +67,39 @@ func (actor Actor) GetServiceInstanceSummaryByNameAndSpace(name string, spaceGUI
...
+	var serviceInstanceSummaries []ServiceInstanceSummary
+	for _, serviceInstance := range spaceSummary.ServiceInstances {
+		instanceSummary := ServiceInstanceSummary{}
+
+		instanceSummary.Name = serviceInstance.Name
+		instanceSummary.ServicePlan.Name = serviceInstance.ServicePlan.Name
+		instanceSummary.Service.Label = serviceInstance.ServicePlan.Service.Label
+		instanceSummary.LastOperation = serviceInstance.LastOperation
+		instanceSummary.Service.ServiceBrokerName = serviceGUIDToBrokerName[serviceInstance.ServicePlan.Service.GUID]
+		instanceSummary.ServiceInstance.Type = discoverServiceInstanceType(serviceInstance)
```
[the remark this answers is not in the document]
williammartin: We think this trade off is about clarity vs. visual noise. We prefer to keep the method, to keep the visual noise down on the set of assignments, but don't feel too strongly. Also happy to change to a better name if there is one?

actor/v2action/service_instance_summary.go
```diff
@@ -67,31 +67,39 @@ func (actor Actor) GetServiceInstanceSummaryByNameAndSpace(name string, spaceGUI
...
-		serviceInstanceSummary, summaryInfoWarnings, summaryInfoErr := actor.getSummaryInfoCompositeForInstance(
-			spaceGUID,
-			ServiceInstance(instance),
-			false)
-		allWarnings = append(allWarnings, summaryInfoWarnings...)
-		if summaryInfoErr != nil {
-			return nil, allWarnings, summaryInfoErr
-		}
-		summaryInstances = append(summaryInstances, serviceInstanceSummary)
+	services, warnings, err := actor.CloudControllerClient.GetServices()
```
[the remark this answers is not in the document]
williammartin: Done

api/cloudcontroller/ccv2/space_summary.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	spaceSummary.Name = ccSpaceSummary.Name
+	spaceSummary.Applications = ccSpaceSummary.Applications
+	spaceSummary.ServiceInstances = ccSpaceSummary.ServiceInstances
+
+	return nil
+}
+
+func (client *Client) GetSpaceSummary(spaceGUID string) (SpaceSummary, Warnings, error) {
```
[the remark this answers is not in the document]
williammartin: Done

api/cloudcontroller/ccv2/space_summary.go
```diff
@@ -0,0 +1,72 @@
...
+	Service SpaceSummaryService `json:"service"`
+}
+
+type SpaceSummaryServiceInstance struct {
+	Name          string                  `json:"name"`
+	ServicePlan   SpaceSummaryServicePlan `json:"service_plan"`
+	LastOperation LastOperation           `json:"last_operation"`
+}
+
+type SpaceSummary struct {
```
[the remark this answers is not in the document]
williammartin: Done

actor/v2action/service_instance_summary.go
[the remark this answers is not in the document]
georgi-lozev: Yes, there is a reason for using `/v2/services` instead of `/v2/spaces/:guid/services`.

`/v2/spaces/:guid/services` returns only the services which are currently visible to your space. In the case where someone has disabled the service access to that service offering, but you still have a service instance in your space `/v2/spaces/:guid/services` won't return that service while `/v2/services` will. This results in missing broker name in case the service offering is disabled for your org. Hope the following snippet makes it more clear.

```
→ cf enable-service-access overview-service
Enabling access to all plans of service overview-service for all orgs as admin...
OK

→ cf curl /v2/services
{
   "total_results": 1,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [
    ...
   ]
}

→ cf curl /v2/spaces/1aff19b7-8185-49ab-a4bd-e79dc50f943f/services
{
   "total_results": 1,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [
   ...
   ]
}

→ cf disable-service-access overview-service
Disabling access to all plans of service overview-service for all orgs as admin...
OK

→ cf curl /v2/spaces/1aff19b7-8185-49ab-a4bd-e79dc50f943f/services
{
   "total_results": 0,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [] // NOTE the empty resources
}

→ cf curl /v2/services
{
   "total_results": 1,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [
    ...
   ]
}
```

One option for this workflow to be simplified further and the call to `/v2/services` omitted is if we have the broker name as part of the summary response.
https://www.pivotaltracker.com/story/show/163455079
Another is to change the `/v2/space/:guid/services` behaviour, but I'm not sure if it won't affect something else.

I think for now we should stick with the call to `/v2/services`, since it won't require CC changes and improve it in future.

command/v6/services_command.go
williammartin: I don't think so. We've changed both to unstable, because it doesn't really matter and having `SliceStable` implies we need stability.

command/v6/services_command_test.go
williammartin: Agree. We've moved around some service instances, and apps.

api/cloudcontroller/ccv2/space_summary_test.go
williammartin: Done.

api/cloudcontroller/ccv2/space_summary_test.go
williammartin: Not needed. Removed.

actor/v2action/service_instance_summary.go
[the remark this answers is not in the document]
willmurphyscode: I think the verb in the function name is a bit odd - `serviceInstanceType(serviceInstance)` might be less surprising to read, but also this function does what it says and seems clear and simple, so I'm fine with it.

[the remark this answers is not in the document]
williammartin: I could fix it up now if you'd prefer that, no strong feelings here.

PR #1547, opened 2019-01-18
Update OIDC ops file for cf-deployment v7

henryaj:
https://www.pivotaltracker.com/story/show/163294251


tjvman:
Looks fine, but we'll need to wait on merging this until our environments are actually upgraded to cf-deployment v7. I'll block the tracker story for this PR on the upgrade chore.

PR #1550, opened 2019-01-23
Reduce superfluous api requests during update-service command

### LINKED ISSUE #1549: cf update-service very slow on landscapes with many orgs and/or service plans

- [ ] I am reporting a bug that others will be able to reproduce.

I am trying to run `cf update-service <instance> -p <plan>`. It takes very long to complete.

**What happened**
When enabling `CF_TRACE`, I saw that the cf cli gets a complete list of `service_plan_visibilities` - on this specific landscape there are currently **290275** "total_results" spread over **5806** "total_pages", i.e. the cf cli issues >5000 requests.

**Expected behavior**
The `cf update-service` should be *fast* and not result in superfluous api requests.

**To Reproduce**
This is only possible when you have a landscape with many orgs and/or service plans...

**Provide more context**
- I've run the same command with different cf cli versions, from `6.42.0+0cba12168.2019-01-10` back to `6.34.1+bbdf81482.2018-01-17`.
- CC API v2 version: 2.128.0
- The [buildPlanToOrgVisibilityMap](https://github.com/cloudfoundry/cli/blob/7f0caee632f42cda0eabe0c0858b489582276f58/cf/actors/planbuilder/plan_builder.go#L147) function could use [servicePlanVisibilityRepo.Search(<org_guid>)](https://github.com/cloudfoundry/cli/blob/7f0caee632f42cda0eabe0c0858b489582276f58/cf/api/service_plan_visibility.go#L59) instead of servicePlanVisibilityRepo.List(). I've already modified cf cli locally with the above suggestion and would be happy to file a PR.

Note: As of January 2019, we no longer support API versions older than CF Release v284/CF Deployment v1.7.0 (CAPI Release: 1.46.0 (APIs 2.100.0 and 3.35.0).

philippthun:
Instead of getting all service_plan_visibilities, query for a specifc org when possible.

Fixes #1549

### DIFF
--- a/cf/actors/planbuilder/plan_builder.go
+++ b/cf/actors/planbuilder/plan_builder.go
@@ -145,25 +145,19 @@ func (builder Builder) containsGUID(guidSlice []string, guid string) bool {
 }
 
 func (builder Builder) buildPlanToOrgVisibilityMap(orgName string) (map[string][]string, error) {
-	// Since this map doesn't ever change, we memoize it for performance
-	orgLookup := make(map[string]string)
-
 	org, err := builder.orgRepo.FindByName(orgName)
 	if err != nil {
 		return nil, err
 	}
-	orgLookup[org.GUID] = org.Name
 
-	visibilities, err := builder.servicePlanVisibilityRepo.List()
+	visibilities, err := builder.servicePlanVisibilityRepo.Search(map[string]string{"organization_guid": org.GUID})
 	if err != nil {
 		return nil, err
 	}
 
 	visMap := make(map[string][]string)
 	for _, vis := range visibilities {
-		if _, exists := orgLookup[vis.OrganizationGUID]; exists {
-			visMap[vis.ServicePlanGUID] = append(visMap[vis.ServicePlanGUID], orgLookup[vis.OrganizationGUID])
-		}
+		visMap[vis.ServicePlanGUID] = []string{org.Name}
 	}
 
 	return visMap, nil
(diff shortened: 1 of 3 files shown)


abbyachau:
Hi @philippthun thanks for this pull request. Could you detail the issue this pull request is meant to improve? Latency for example when calling `update-service`? If so, could you give us more details. Thanks!

philippthun:
Hi @abbyachau,

As Georgi mentioned, I've described the problems in issue #1549. Please let me know if something needs to be added to this PR.

Best regards,
Philipp

AartiKriplani:
@philippthun : thank you for submitting the PR, the change looks good but it would be nice to have some integration tests for the scenario. 
Since the fix is in legacy code, and it's a *small* change, we don't expect a full backfill of integration tests (but still expect integration tests for the change).
Thanks!

georgi-lozev:
Hey @philippthun. In case you need any support how this test should look like or need some guidelines how to run the CLI integration tests don't hesitate reach out to us.

philippthun:
@AartiKriplani and @georgi-lozev - I've added an integration test for the `update-service` command. Please let me know if you want me to test additional aspects.

philippthun:
FYI - In the meantime we are using a patched version of cf cli containing this fix in our automation pipeline. The overall runtime as well as the number of cf api calls reduced drastically as expected.

AartiKriplani:
This is released as part of CLI v6.43.0, you don't have to use a patched version now.Thank you for the PR.

PR #1551, opened 2019-01-25
Fix  enable-service-access command: creating unnecessary plan visibilities 

nmaslarski:
`cf enable-service-acess` command should not create a plan visibility if the plan is already public.

This PR:
1. Delete the service plan visibility, if any, when enabling service for all orgs
2. Do not enable on org level when the plan is already public

For more information [see here](https://www.pivotaltracker.com/story/show/162747373)

Best,
Niki
on Behalf of SAPI Team


AartiKriplani:
Closed because duplicate of https://github.com/cloudfoundry/cli/pull/1536

PR #1553, opened 2019-01-25
Finish refactor of service-access command

williammartin:
This PR finishes the work started last year on to refactor the `service-access` command. 

It's a fairly hefty set of changes, and the logic is quite tricky. We've tried our best to simplify difficult cases but ultimately, the combinatorial nature of the `-o, -b, -e` flags _is_ difficult. There are a few changes that I wouldn't necessarily count as "refactoring", but are needed to fix frankly confusing (and difficult to replicate) behaviours in the legacy command.

For example, given a set of services registered as follows:

```
→ cf service-access
Getting service access as admin...
broker: overview-broker
   service                         plan      access    orgs
   overview-service                simple    limited   acceptance
   overview-service                complex   limited   acceptance
   overview-service-volume-mount   simple    none
   overview-service-volume-mount   complex   none

broker: app-overview
   service            plan      access    orgs
   overview-service   simple    all
   overview-service   complex   limited   system
```

There are behavioural differences:

OLD:

```
→ cf service-access -b overview-broker -e overview-service -o system
Getting service access for broker overview-broker and service overview-service and organization system as admin...
broker: overview-broker
   service            plan   access   orgs
   overview-service
```

NEW:

```
→ cf service-access -b overview-broker -e overview-service -o system
Getting service access for broker overview-broker and service overview-service and organization system as admin...
```

DIFF:

The old CLI shows a broker, and a service even though there are no services or plans accessible for that broker from the `system` org.

---

Similarly, there are assumptions in the legacy command that services are uniquely named, which isn't true as of the `multi-service-registration` track. 

---

Finally, there is some difference in the output depending on whether the filter flags result in any matches:

OLD:

```
→ cf service-access -b app-overview -e overview-service-volume-mount -o acceptance
Getting service access for broker app-overview and service overview-service-volume-mount and organization acceptance as admin...
broker: app-overview
   service   plan   access   orgs
```

NEW:

```
→ cf service-access -b app-overview -e overview-service-volume-mount -o acceptance
Getting service access for broker app-overview and service overview-service-volume-mount and organization acceptance as admin...
```

DIFF:

The old CLI provides a listing for the requested broker but shows that no services were found with the provided name that were accessible from the provided org. I think this is kind of confusing behaviour, but I could understand why it might be desirable depending on the intent of the command.

I think the Old CLI behaviour makes sense if the intent is "Show me the services available with this name, under this org for this broker", and the CLI shows a broker with nothing available.

I think the New CLI behaviour makes sense if the intent is "Filter me the brokers that satisfy all these given requirements".

---

Ultimately, this command is pretty complex, with a bunch of edges so I'm sure you'll find plenty in here that makes a review valuable. We may have to make some UX changes on this PR if you disagree with the above ^, which is totally fine. If you think it makes sense to pair on a review, let me know and I'm sure I can figure out some time to make that happen.

Thanks!


williammartin:
It should be noted that I accidentally created this PR before the story was accepted, but I suspect few large changes, should you start reviewing it. I'll update the PR when acceptance is done on our side though.

tjvman:
@abbyachau Would like to get your thoughts on the UX stuff described in Will's initial comment.

actor/v2action/service.go
```diff
@@ -69,15 +70,32 @@ func (actor Actor) getServiceByNameForSpace(serviceName, spaceGUID string) (Serv
...
 }
 
+// ServicesWithPlans provides a mapping of Services to ServicesPlans
 type ServicesWithPlans map[Service][]ServicePlan
 
-func (actor Actor) GetServicesWithPlansForBroker(brokerGUID string) (ServicesWithPlans, Warnings, error) {
+// GetServicesWithPlansForBroker returns a map of Services to ServicePlans for a particular broker.
+// A particular service with associated plans from a broker can be fetched by additionally providing
+// a service name.
+func (actor Actor) GetServicesWithPlansForBroker(brokerGUID, serviceName string) (ServicesWithPlans, Warnings, error) {
```
tjvman: Given that this method now fetches plans for a given broker and (optionally) service, its current name doesn't seem accurate.

I'm also wondering if it's doing too many things at this point.

command/v6/service_access_command_test.go
```diff
@@ -138,9 +140,98 @@ var _ = Describe("service-access Command", func() {
 					nil)
 			})
 
-			It("displays flavor text", func() {
-				Expect(testUI.Out).To(Say("Getting service access as some-user..."))
-				Expect(fakeConfig.CurrentUserCallCount()).To(Equal(1))
+			Describe("flavour text", func() {
```
tjvman: I'd suggest [`DescribeTable`](https://godoc.org/github.com/onsi/ginkgo/extensions/table#DescribeTable) as an option for making these tests a bit cleaner.

command/v6/service_access_command_test.go
```diff
@@ -173,30 +264,52 @@ var _ = Describe("service-access Command", func() {
...
 
-				BeforeEach(func() {
-					brokers = generateBrokers(2)
-					fakeActor.GetServiceBrokerSummariesReturns(generateBrokers(2), v2action.Warnings{"warning"}, nil)
+					BeforeEach(func() {
+						brokers = generateBrokers(2)
+						fakeActor.GetServiceBrokerSummariesReturns(generateBrokers(2), v2action.Warnings{"warning"}, nil)
+					})
+
+					It("displays each service broker, service, plan and access with org", func() {
```
tjvman: Please add an assertion for the warning that's getting returned from the actor.

command/v6/service_access_command_test.go
```diff
@@ -173,30 +264,52 @@ var _ = Describe("service-access Command", func() {
...
-				var brokers []v2action.ServiceBrokerSummary
+			Describe("tabular output", func() {
+				When("there are multiple brokers", func() {
+					var brokers []v2action.ServiceBrokerSummary
 
-				BeforeEach(func() {
-					brokers = generateBrokers(2)
-					fakeActor.GetServiceBrokerSummariesReturns(generateBrokers(2), v2action.Warnings{"warning"}, nil)
+					BeforeEach(func() {
+						brokers = generateBrokers(2)
```
tjvman: Why is generateBrokers called on this line and then again on the line below? Isn't the output going to be identical?

command/v6/service_access_command_test.go
```diff
@@ -173,30 +264,52 @@ var _ = Describe("service-access Command", func() {
...
+				When("the summaries returned are unordered", func() {
+					BeforeEach(func() {
+						unorderedBrokers := []v2action.ServiceBrokerSummary{createBroker(2), createBroker(1)}
+						fakeActor.GetServiceBrokerSummariesReturns(unorderedBrokers, v2action.Warnings{"warning"}, nil)
+					})
+
+					It("sorts brokers and services before displaying", func() {
+						Expect(executeErr).ToNot(HaveOccurred())
+
+						// Unfortuantely we cannot use the brokers struct here as the command does an in place sort
```
tjvman: I'm unclear as to how the ordering in this case would end up being different than in the test above. Can you clarify?

command/v6/service_access_command_test.go
```diff
@@ -173,30 +264,52 @@ var _ = Describe("service-access Command", func() {
...
-					Expect(testUI.Out).To(Say(`broker:\s+%s`, brokers[1].Name))
-					Expect(testUI.Out).To(Say(tableHeaders))
-					Expect(testUI.Out).To(Say(rowMatcher(brokers, 1, 1, 1, "all")))
-					Expect(testUI.Out).To(Say(rowMatcher(brokers, 1, 1, 0, "none")))
-					Expect(testUI.Out).To(Say(rowMatcher(brokers, 1, 0, 1, "all")))
-					Expect(testUI.Out).To(Say(rowMatcher(brokers, 1, 0, 0, "limited")))
+				When("the summaries returned are unordered", func() {
+					BeforeEach(func() {
+						unorderedBrokers := []v2action.ServiceBrokerSummary{createBroker(2), createBroker(1)}
+						fakeActor.GetServiceBrokerSummariesReturns(unorderedBrokers, v2action.Warnings{"warning"}, nil)
```
tjvman: Please assert on these warnings.

integration/shared/isolated/service_access_command_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"code.cloudfoundry.org/cli/integration/helpers"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = FDescribe("service-access command", func() {
```
tjvman: Looks like a focus slipped through.

actor/v2action/organization.go
```diff
@@ -124,3 +124,22 @@ func (actor Actor) GetOrganizations() ([]Organization, Warnings, error) {
 	}
 	return returnedOrgs, Warnings(warnings), err
 }
+
+// OrganizationExistsWithName returns true if there is an Organization with the
+// provided name, otherwise false.
+func (actor Actor) OrganizationExistsWithName(orgName string) (bool, Warnings, error) {
```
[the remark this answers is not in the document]
williammartin: This is an artifact of refactoring. It used to be that the `GetOrganizationByName()` method was called directly from `GetServiceBrokerSummaries()` which didn't convey any intent at all, then we introduced this `OrganizationExistsWithName()` method, and later pulled it into a private method to reduce the visual noise.

Now I think it's fair to say that `ensureOrgExists` captures the intent. Would be happy to remove the new method.

actor/v2action/composite/service_broker_summary.go
```diff
@@ -128,15 +191,45 @@ func (c *ServiceBrokerSummaryCompositeActor) fetchServiceSummary(service v2actio
 			}
 		}
 
-		servicePlanSummaries = append(servicePlanSummaries,
-			v2action.ServicePlanSummary{
-				ServicePlan: v2action.ServicePlan(plan),
-				VisibleTo:   visibleTo,
-			})
+		if organizationName == "" || (plan.Public || isVisibleTo(organizationName, visibleTo)) {
```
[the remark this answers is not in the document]
williammartin: Agree, we've fixed this though I don't _love_ the new implementation either as it makes the reader of the `fetch` method think "wait but there was no organization passed, why do I have to check any visibility", and then go into the `isPlanVisibleToOrg` method to understand that it wasn't important at all.

actor/v2action/composite/service_broker_summary_test.go
```diff
@@ -409,5 +410,200 @@ var _ = Describe("Service Broker Summary Actions", func() {
...
+								{GUID: "service-plan-guid-1", Name: "service-plan-1", Public: false},
+							},
+						}, nil, nil)
+
+						fakeVisibilityActor.GetServicePlanVisibilitiesReturns([]v2action.ServicePlanVisibility{
+							{OrganizationGUID: "my-org-guid"},
+						},
+							nil, nil)
+
+						fakeOrgActor.GetOrganizationReturnsOnCall(0,
```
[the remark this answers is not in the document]
williammartin: Done.

actor/v2action/service.go
williammartin: So with regards a new pattern, what we were thinking was something like the following:

```
...
type FilterOpt func() ccv2.Filter

func FilterByServiceName(serviceName string) FilterOpt {
	return func() ccv2.Filter {
		return ccv2.Filter{
			Type:     constant.LabelFilter,
			Operator: constant.EqualOperator,
			Values:   []string{serviceName},
		}
	}
}

func FilterByBrokerGUID(brokerGUID string) FilterOpt {
	return func() ccv2.Filter {
		return ccv2.Filter{
			Type:     constant.ServiceBrokerGUIDFilter,
			Operator: constant.EqualOperator,
			Values:   []string{brokerGUID},
		}
	}
}

func (actor Actor) GetServicesWithPlans(opts ...FilterOpt) (ServicesWithPlans, Warnings, error) {
	filters := []ccv2.Filter{}
	for _, o := range opts {
		filters = append(filters, o())
	}
...
```

This would allow callers to do any of the following:

```
GetServicesWithPlans() // would get all services with all plans for all brokers
GetServicesWithPlans(FilterByBrokerGUID(brokerGUID)) // would get all services with all plans for one broker
GetServicesWithPlans(FilterByBrokerGUID(brokerGUID), FilterByServiceName(serviceName)) // would get all plans for one service for one broker
```
and also just filtering by serviceName but no one needs that.

For a while I (Will) have been thinking that this pattern might be nice to see in favour of all the `GetFooWithBar` functions. Interested in your thoughts.

williammartin: We could additionally have some type definitions or conversion piece to avoid the leakage of `ccv2` to callers of the `actor` layer. Alternatively we could just introduce something like the ccv2 filtering at the actor layer, but I think this proposed pattern might be better for decoupling the intent from the practicalities.

integration/shared/isolated/service_access_command_test.go
ablease: Whoops, our bad.

command/v6/service_access_command_test.go
ablease: Good catch. Thanks.

command/v6/service_access_command_test.go
ablease: Will do, Thanks.

command/v6/service_access_command_test.go
ablease: Yeah thats a good point, we will reduce these Describe down to a DescribeTable and see how it looks.

command/v6/service_access_command_test.go
williammartin: The ordering would be different because we were passing in the brokers rather than generating a duplicate of the slice, and then it gets sorted in place. However, we realised that the only extra thing this was testing was the sorting of the brokers, because the `generateBroker` function was already producing an out of order structure, so we just deleted this test and in the one above also checked that sorting of brokers works.

command/v6/service_access_command_test.go
williammartin: Done.

actor/v2action/service.go
```diff
@@ -105,3 +123,22 @@ func (actor Actor) GetServicesWithPlansForBroker(brokerGUID string) (ServicesWit
 
 	return servicesWithPlans, allWarnings, nil
 }
+
+// ServiceExistsWithName returns true if there is an Organization with the
+// provided name, otherwise false.
+func (actor Actor) ServiceExistsWithName(serviceName string) (bool, Warnings, error) {
```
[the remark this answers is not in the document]
williammartin: Mostly for clarity of intent really, as discussed for `OrganizationExistsWithName`. Can change if you prefer.

williammartin:
Re: "We spent most of our reviewing time trying to wrap our heads around GetServiceBrokerSummaries. The logic itself seems sound but it definitely took some doing to understand it. We found the comment's use of "parent" and "children" more confusing than helpful - those terms didn't appear anywhere else, so we had to figure out what they meant."

Having taken the time to understand it (thanks!), is there some better phrasing we could put here? I wasn't sure how to phrase the relationship between the concepts. Maybe we could start with an explanation that we're referring to the Broker, Service, Plan structure as a tree, and that would immediately clarify?

actor/v2action/service.go
a-shan: We really like this style!

Our only concern is whether we can lean on the compiler / types to tell a dev which filters can be paired with which functions. For example, `GetServicesWithPlans(FilterBySpaceGUID(spaceGUID))` would be a run-time problem rather than a compile-time problem.

williammartin: Hey folks, we decided to back away from this pattern (although I love it), because really, when the `FilterOpt` functions are just returning filters, we might as well pass filters directly instead. This pattern is really useful when you want to modify certain behaviours i.e. in this OSBAPI client I wrote: https://github.com/williammartin/osbapi/blob/master/client.go#L10-L21

We could consider providing the functions with the array of filters and letting it be internally modified.

integration/shared/isolated/disable_service_access_command_test.go
```diff
@@ -156,12 +156,12 @@ var _ = Describe("disable service access command", func() {
 						session = helpers.CF("service-access", "-e", service)
 						Eventually(session).Should(Exit(0))
 						Eventually(session).Should(Say("broker:\\s+%s", broker.Name))
-						Eventually(session).Should(Say("%s\\s+%s\\s+none",
+						Eventually(session).Should(Say("%s\\s+.*\\s+all",
```
jenspinney: Two spaces will match the regex `\\s+.*\\s+`. It seems a little too loose. We're wondering why the expectation needed to be loosened in the first place. If it does need to be loosened, we should probably change the `.*` to a `.+`.

Ema93sh:
> Maybe we could start with an explanation that we're referring to the Broker, Service, Plan structure as a tree

We agree that that would help

abbyachau:
hey @williammartin sorry I'm just seeing this. Had a chat with @jenspinney and we agree that we should move forward with the "new" behavior, i.e `"Filter me the brokers that satisfy all these given requirements"`. I've added this to rule to our new [product style guide](https://github.com/cloudfoundry/cli/wiki/CLI-Product-Specific-Style-Guide#providing-filters-to-a-list-command) to capture this decision.

williammartin:
No problem @abbyachau, Thanks for the style guide update. We have a story in our backlog (https://www.pivotaltracker.com/story/show/163453704) to provide a useful message in this case. It seems informative to the user that their set of filters didn't match any results.

actor/v2action/composite/service_broker_summary.go
```diff
@@ -36,13 +40,44 @@ type ServiceBrokerSummaryCompositeActor struct {
...
+	if organizationName != "" {
+		orgExistsWarnings, orgExistsErr := c.ensureOrgExists(organizationName)
+		warnings = append(warnings, orgExistsWarnings...)
+		if orgExistsErr != nil {
+			return nil, warnings, orgExistsErr
+		}
+	}
+
+	if serviceName != "" {
+		serviceExistsWarnings, serviceExistsErr := c.ensureServiceExists(serviceName)
```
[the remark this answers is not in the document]
AartiKriplani: It wasn't. but on more thought about the scenario, it would be when both broker and service filters are provided, broker exists and the service exists but not for the broker provided. With the current implementation, it would print an empty list. Taking this one step further, similar scenarios also exist for org+broker, org+service and org+broker+service filter combinations. And handling all of them is going to complicate this already complex implementation. 
@abbyachau @laurzca Wondering if an empty list is acceptable in these scenarios or do we want to explicitly fail?

PR #1561, opened 2019-02-04
Fix flaky marketplace integration test

AartiKriplani:
The issue was happening because of AfterEach() cleanup failing sometimes,
if other tests that deploy a service broker fail during their clean up step.
We have made the test generic to check that the marketplace command succeeds,
when the user is not logged in, whether or not services are empty or not.

More details can be found [here](https://www.pivotaltracker.com/story/show/163624111).

### DIFF
--- a/integration/shared/global/marketplace_command_test.go
+++ b/integration/shared/global/marketplace_command_test.go
@@ -12,18 +12,15 @@ import (
 var _ = Describe("marketplace command", func() {
 	When("an API endpoint is set", func() {
 		When("not logged in", func() {
-			When("there are no accessible services", func() {
-				BeforeEach(func() {
-					helpers.LogoutCF()
-				})
+			BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


williammartin:
I don't understand this commit. Two thoughts:
 - Now it seems we have no integration coverage of "No service offerings found"
 - Why do we think an `AfterEach` cleanup was failing in our CI failures (found on the tracker story), they don't exhibit the same `AfterEach` failure as the `integration-windows` build?

AartiKriplani:
Looking at the tests, it felt like it was about testing that not logged in users can still access the `cf m` command. Testing empty list of service offerings seemed like a less important aspect to us, given the effort to keep to maintain this test. But if we think it's an important scenario, we can add it back and try to find the cause of the debris. Closing it for now, after the internal discussion, will re-open with another solution for the flake maybe.

PR #1567, opened 2019-02-08
Add -b flag to purge-service-offering command

henryaj:
Add support for the -b (specify a broker) flag to `cf purge-service-offering`.

cc @nmaslarski


henryaj:
We had to revert this as there was some code missing (a bad squash on my part – sorry!). We'll create a new PR for this with all the work included.

PR #1573, opened 2019-02-13
Expose '-b' flag in help messages

henryaj:
This PR unmarks the '-b' (broker name) flag as experimental, and adds help text on its use to the relevant commands.

cc @waterlink @AartiKriplani @nmaslarski


henryaj:
⚠️ Not to be merged before #1572.

PR #1588, opened 2019-03-10
Update Readme to Include Windows Installer

bottkars:
Added installer using Powershell Gallery

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ sudo wget -O /etc/yum.repos.d/cloudfoundry-cli.repo https://packages.cloudfoundr
 # ...then, install the cf CLI (which will also download and add the public key to your system)
 sudo yum install cf-cli
 ```
+**Windows10** and **WindowsServer** using Powershell Gallery:
+
+```
+# First we istall the Install Script
+Install-Script install-cf-cli -Scope CurrentUser
+# Then we install the CLI. Use Tab to select Versions. The Installer uses Dynamic Variables to Browse <CLIRelease>
+Install-cf-cli.ps1 -CLIRelease '6.43.0' -DownloadDir "$HOME\Downloads" 
+```
 
 ### Installers and compressed binaries
 


bwasmith:
Hey @bottkars , thank you for submitting this update to the README.

@a-shan and I ran your commands as listed and we had a few questions. 

First, we ran the first command, and everything was fine. Then when trying to run the second command, we saw this error: 
windows-permission-error
(I mistyped "Documents" instead of "Downloads", but that should be unrelated to this error)

To get around this, we used `Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass` which [we found online here](http://tritoneco.com/2014/02/21/fix-for-powershell-script-not-digitally-signed/). Is this something we expect users to do as well? Our knowledge of Powershell is limited, so there is a good chance we are missing something. 

Second, just a small thing perhaps, but after successfully running the second command, we had to start  a new shell to get the cf-cli in our PATH. On UNIX systems, it would make sense that we might have to re-source a config. Is something similar going on here?

bottkars:
the Execution Policy Setting can be managed by GPO or locally, so it is dependant.

a dev workstation might allow unsgned script´s, or set it to unrestricted.

I updated the script do dynamically reload the path in the actual session,
so if you run Install-Script install-cf-cli -MinimumVersion 1.8 -Force, you get the latest Version 
that will reload the path ( uninstall cf cli, remove path from env,open a new PowerShell session and test )
image

Remove from System Env.
image
install after path cleared and show new path after install:
image

willmurphyscode:
Hi @bottkars ,

Thanks for the pull request.

It looks like you PR adds to the README a snippet that downloads a script from the Powershell Gallery, and that this script then knows how to install the CF CLI. In effect, it recommends that users use an installer that this team does not control (or even, until now, know about).

It's our policy that any officially recommended installer be cryptographically signed and maintained by the CF CLI team. If you feel we are missing a Powershell installer and would like us to consider maintaining a Powershell Gallery package, please open an issue.

Thank you.

PR #1593, opened 2019-03-15
Update delete-orphaned-routes description to clarify deletion scope

AP-Hunt:
It's unclear from the help text description how wide a `cf delete-orphaned-routes` command will
look to find orphaned routes, e.g. as when run as an admin or as a SpaceDeveloper who
has access to multiple spaces.

This cosmetic update makes it clear that it's only routes which exist in the current space,
not mapped to an app, which will be deleted.

## How Urgent Is The Change?

It is not.


AP-Hunt:
Hi Will,

Not a problem. I'll get that sorted out.

Issue #1617, opened 2019-04-19
Improve UX for `cf login -sso`


prolane:
I've been in touch with you guys a while ago on Slack about this feature request. I'd like to contribute a bit of code. It seemed at least some of you liked my idea. I was asked to open a Github issue, so here I am.

**What's the user value of this feature request?**
Currently authenticating with `cf login -sso` requires you to: 
- manually open a browser
- copy paste UAA url from the cf cli to the browser.
- After authenticating in UAA, copy the shown 'temporary authorization code' from the browser.
- Paste the 'temporary authorization code' on the stdin of cf cli

I'd like to change this into a user experience similar to the `fly` cli of concourse:
- Browser is opened automatically with the right URL
- Authenticate in UAA
- CF cli picks up the authorization code automatically, so it means you're done.

**Who is the functionality for?**
User who authenticate to CF using Single Sign On

**How often will this functionality be used by the user?**
Every time an user needs to re-authenticate

**Who else is affected by the change?**
No. Just like with the `fly` cli, pasting the 'temporary authentication code' on stdin will also still work.

**Is your feature request related to a problem? Please describe.**
Not a problem. Just trying to improve the user experience. Our users find the Single Sign On cumbersome with the copy-pasting currently.

**Describe the solution you'd like**
See the above value proposition

**Describe alternatives you've considered**
I don't think there are any?

**Additional context**
I was thinking to change the `cf/commands/login.go`
Here is the work in progress: https://github.com/SpringerPE/cli/blob/master/cf/commands/login.go

I'm expecting I'll need a change in UAA for this to work, as UAA needs to do a redirect to `localhost` with the temporary authorization code as query string parameter. Similar to how the oauth redirects work.

abbyachau:
Hey @prolane thank you for reaching out! Appreciate it. I'm copying in a few UAA PMs here for their feedback @aramprice @cwang-pivotal on this as well.

prolane:
@abbyachau Thanks for your positive reply.
Yes, on Slack it was mentioned the team is rewriting the `cf login` command. I guess it does indeed make sense to make this enhancement when the rewrite is done. Are there some rough timelines?

abbyachau:
cc @dbeneke (UAA PM)

hey @prolane here is the epic for the `cf login` rewrite effort: https://www.pivotaltracker.com/epic/show/4186886; and the [release marker](https://www.pivotaltracker.com/story/show/164872756) for it.

drnic:
[...]

PR #1624, opened 2019-04-25
if cf misconfigured, still allow user to 'cf login --sso'

drnic:
If a target CF has not been configured correctly, this PR still allows a user to `cf login --sso`. This is a "good thing". We should encourage authentication via browsers and password vaults.

## Why Is This PR Valuable?
​
We should encourage all users to move away from providing passwords via CLIs in terminals. An oauth sequence into the browser allows the user to use their 1password integration.

Currently some users cannot use `cf login --sso` because their platform operator misconfigured their UAA prompts. This PR fixes this issue for those users.

## What benefits will be realized by the code change? What users would want this change? What user need is this change addressing? 

I am doing a video on `cf login` tips & tricks. I'll be promoting `cf login --sso` as my preferred method for ppl to login. Currently it's broken for some users. 

## Explain why this functionality should be in the cf CLI, as opposed to a plugin.

Its a fix to `cf login`.
​
## Applicable Issues

Initial PR does not have i18n. I'll do this in a follow up PR.

## How Urgent Is The Change?

I'd like to have this feature fixed for my video.​

## Other Relevant Parties

UAA

### DIFF
--- a/cf/commands/login.go
+++ b/cf/commands/login.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"strconv"
 
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccversion"
@@ -175,6 +176,13 @@ func (cmd Login) authenticateSSO(c flags.FlagContext) error {
 
 	credentials := make(map[string]string)
 	passcode := prompts["passcode"]
+
+	if passcode.DisplayName == "" {
+		passcode = coreconfig.AuthPrompt{
+			Type:        coreconfig.AuthPromptTypePassword,
+			DisplayName: fmt.Sprintf("Temporary Authentication Code ( Get one at %s/passcode )", cmd.config.AuthenticationEndpoint()),
+		}
+	}
 
 	for i := 0; i < maxLoginTries; i++ {
 		if c.IsSet("sso-passcode") && i == 0 {


abbyachau:
Hi @drnic could you please provide more information about this pull request:

## Description of the Change

## Why Is This PR Valuable?

​

## What benefits will be realized by the code change? What users would want this change? What user need is this change addressing? -->


## Explain why this functionality should be in the cf CLI, as opposed to a plugin. -->

​

## Applicable Issues



## How Urgent Is The Change?

​
## Other Relevant Parties



​

drnic:
Ah, sorry for deleting all the text. The start of it said "you must have signed the CLA" and I assumed I could ignore it.

drnic:
@abbyachau FYI - your prompts for me above are (very?) different from https://github.com/cloudfoundry/cli/blob/master/.github/PULL_REQUEST_TEMPLATE.md

drnic:
@abbyachau I've updated the description with your questions and my answers. Thanks + sorry.

abbyachau:
Ah thanks @drnic I've updated the pull request template unsure why so much of it was obfuscated.

abbyachau:
Thanks for updating @drnic; we are currently rewriting the `cf login` command; for reference here is our [epic](https://www.pivotaltracker.com/epic/show/4186886).  

I'm struggling to understand how I would accept this story. Would you mind providing acceptance criteria so that I understand what the previous workflow looks like and what your proposed changes are. Something similar to:
**Previous workflow**
Given a user with a misconfigured CF
When I run `cf l --sso`
I expect the command to error with the following...

**Current workflow**
...

drnic:
Acceptance steps:

1. Run `cf dev start` (as a simple way to bring up a CF environment that does not support `cf login --sso`).
2. Run `cf login --sso` and the user should be prompted with the URL for the passcode; and then be allowed to complete the login sequence.

abbyachau:
Thank you. No problem, it was just an example - appreciate the acceptance steps.

abbyachau:
Hi @drnic thank you for this pull request. We were planning to review/move forward with efforts detailed in https://github.com/cloudfoundry/cli/issues/1617 and as such will not be pulling in this feature. As you know, we hope to finish rewriting `cf login` and after we complete that work, we look forward to working with the OP for #1617. Please let us know if you have any questions or feedback, thanks. I've also updated our pull request template - thank you for letting us know about the confusion.

drnic:
I would appreciate this PR being pulled into current cf cli please. I don’t feel like the explanation above included an explanation why this isn’t a valuable bug fix for current users. :(

abbyachau:
After some digging, this request is not to enhance `cf l --sso` but to allow users to always see the prompt for `Temporary Authentication Code `.  @drnic please correct me if I'm wrong. 

I'm tagging UAA PMs (@cwang-pivotal, @dbeneke) and I'll try to summarize what's going on here to get your feedback.

Currently, even if users do not have an identity provider set up, they can use `cf l --sso` which will return the following:
```
± |master U:1 ?:6 ✗| → ./cf l --sso
API endpoint: https://api.arsicault.cli.fun

Temporary Authentication Code ( Get one at https://login.arsicault.cli.fun/passcode ):
```

And when they go to that url and enter their credentials, they can retrieve the passcode to log into their cf instance.

It appears that this is unintended behavior, [according](https://cloudfoundry.slack.com/archives/C03FXANBV/p1553550203098400) to the UAA: `Passcode prompt is sent back by UAA as part of the info call only if a SAML/OIDC provider is set up`. 

However this is a workflow the CF CLI supports, and we cannot remove support for in the V6 CLI. Furthermore, we believe this is a workflow that some users might use for password managers like LastPass - although we have not validated this assumption. 

The question here is whether this is a workflow that the UAA team supports as this pull request makes that workflow more explicit, and more of a 'real' feature.

drnic:
I believe this issue has appeared on someone’s SO issue https://stackoverflow.com/questions/60210238/cloudfoundry-cli-login-not-working-credentials-were-rejected-please-try-again/60217563#60217563

Could we revisit this?

PR #1626, opened 2019-04-29
Create cf services command performance test

williammartin:
This PR introduces a simple performance test for the `cf services` command. It is in a separate suite than the rest of the integration tests such that it needs to be run directly rather than as part of an existing suite.

Right now it has no expectations to cause the tests to fail, and is only used for data gathering.


FlorentFlament:
+1

bwasmith:
I think @willmurphyscode already mentioned this, but there was a collision for `ip` in the Makefile. So we removed this one and left `integration-push`'s

PR #1628, opened 2019-04-30
if cf misconfigured, provide prompt for 'cf login --sso' (with some locale translations)

drnic:
If a target CF has not been configured correctly, this PR still allows a user to `cf login --sso`. This is a "good thing". We should encourage authentication via browsers and password vaults.

## Why Is This PR Valuable?
​
We should encourage all users to move away from providing passwords via CLIs in terminals. An oauth sequence into the browser allows the user to use their 1password integration.

Currently some users cannot use `cf login --sso` because their platform operator misconfigured their UAA prompts. This PR fixes this issue for those users.

For example, `cf dev start` (which is based on cf-deployment) does not provide any prompt:

```
cf login -a https://api.dev.cfdev.sh --skip-ssl-validation --sso
API endpoint: https://api.dev.cfdev.sh

>
```

## What benefits will be realized by the code change? What users would want this change? What user need is this change addressing? 

I am doing a video on `cf login` tips & tricks. I'll be promoting `cf login --sso` as my preferred method for ppl to login. Currently it's broken for some users on cfdev (which uses cf-deployment).

https://github.com/cloudfoundry-incubator/cfdev/issues/93

## Explain why this functionality should be in the cf CLI, as opposed to a plugin.

It is a fix to `cf login`.
​
## Applicable Issues

This PR does not have translations for all locales; but I did collect some from non-English speakers I know. They are attributed in the commit.

## How Urgent Is The Change?

I'd like to have this feature fixed for my video.​

## Other Relevant Parties

UAA


abbyachau:
Related https://github.com/cloudfoundry/cli/pull/1624

abbyachau:
Hi @drnic I'm reiterating here that we are currently trying to [begin work](https://github.com/cloudfoundry/cli/issues/1617#issuecomment-486341531) again on the `cf login`. We hope to reach out once that work is underway. Thanks.

heyjcollins:
Hello @drnic - I've recently rolled onto the CLI team as PM and I'm curious if this PR still stands given the current login implementation?

JenGoldstrich:
We believe this PR did not actually change the cf CLI since no one falls back on the legacy login command anymore.  Created a slack thread in the CloudFoundry slack in #cli to ask drnic if we should fix this in v6/v7, if he says yes we can look into creating as seperate issue https://cloudfoundry.slack.com/archives/C032824SM/p1590601934161600

cc: @heyjcollins 

Jenna and @jpalmerpivotal

PR #1631, opened 2019-05-01
Add ability to set labels on spaces

tcdowney:

**CLI Version:** v7

## Description of the Change
This adds the ability to add/update labels on a `space` via `cf set-label`.

**CAPI Story:** [#164994212](https://www.pivotaltracker.com/story/show/164994212)

Thank you!
Tim && @monamohebbi


api/cloudcontroller/ccv3/space.go
```diff
@@ -40,3 +45,34 @@ func (client *Client) GetSpaces(query ...Query) ([]Space, Warnings, error) {
 
 	return fullSpacesList, warnings, err
 }
+
+func (client *Client) UpdateSpace(space Space) (Space, Warnings, error) {
+	spaceGUID := space.GUID
+	space.GUID = ""
+	space.Relationships = nil
```
tcdowney: Having to unset these feels weird to us and we saw this also came up in https://github.com/cloudfoundry/cli/commit/781f526de8c1f8b6a065ec18fd20fe0965125790:

> Some weirdness because we need to retrieve the guid field from v3 on the
organization entity, but we are not permitted to send it in a patch
request.

I wonder since these are patch requests if we could actually get away with just having a `MetadataPatchRequest` (or something) struct that all the endpoints could use instead of us having to delete things from each resource piecemeal. Wanted to get some of your thoughts on that before attempting it, though.

tcdowney: Another thing I was thinking was what if we had a generic `ccv3/metadata.go` client. The only thing that's really going to be different between any of these resources is the endpoint that's called and we could probably just switch on `resourceName`.

They should all behave nearly the same otherwise -- especially if we limit the request struct to being purely scoped to metadata like I mentioned above. Thoughts?

willmurphyscode: I like that idea. It would clean up a lot of this duplication.

tcdowney: Logistically what do you think the best way to tackle that refactor would be? Would you be ok with me capturing this in a chore story that CAPI could prioritize working on next and PR separately or is this something y'all would like to cross-team pair on?

I feel like the refactor would make the most sense with these three `set-label` resources merged and in place as a foundation for it .

api/cloudcontroller/ccv3/internal/api_routes.go
```diff
@@ -53,6 +53,7 @@ const (
 	PatchOrganizationRequest                                    = "PatchOrganization"
 	PatchOrganizationRelationshipDefaultIsolationSegmentRequest = "PatchOrganizationRelationshipDefaultIsolationSegment"
 	PatchProcessRequest                                         = "PatchProcess"
+	PatchSpaceRequest                                           = "PatchSpace"
```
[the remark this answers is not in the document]
tcdowney: Addressed in https://github.com/cloudfoundry/cli/pull/1631/commits/425b1317d927660157acbc63a17ee77d2a670428

api/cloudcontroller/ccv3/space_test.go
```diff
@@ -159,4 +160,74 @@ var _ = Describe("Spaces", func() {
...
+							"k2": types.NewNullString("v2"),
+						},
+					},
+				}
+			})
+
+			It("should include the labels in the JSON", func() {
+				Expect(executeErr).ToNot(HaveOccurred())
+				Expect(server.ReceivedRequests()).To(HaveLen(3))
+				Expect(len(warnings)).To(Equal(0))
```
willmurphyscode: I don't understand why `len(warnings)` should be 0. It looks like line 203 has a warning in it.

command/v7/set_label_command_test.go
```diff
@@ -326,6 +326,156 @@ var _ = Describe("set-label command", func() {
...
+							Expect(spaceName).To(Equal(resourceName), "failed to pass space name")
+							Expect(orgGUID).To(Equal("some-org-guid"))
+							Expect(labels).To(BeEquivalentTo(map[string]types.NullString{
+								"FOO": types.NewNullString("BAR"),
+								"ENV": types.NewNullString("FAKE"),
+							}))
+						})
+
+						It("displays a message", func() {
+							Expect(executeErr).ToNot(HaveOccurred())
```
willmurphyscode: Would you mind making this spacing consistent? We generally don't have blank lines between assertions.

command/v7/set_label_command_test.go
```diff
@@ -326,6 +326,156 @@ var _ = Describe("set-label command", func() {
...
+				ResourceName: resourceName,
+			}
+		})
+
+		JustBeforeEach(func() {
+			executeErr = cmd.Execute(nil)
+		})
+
+		It("doesn't error", func() {
+			Expect(executeErr).ToNot(HaveOccurred())
```
willmurphyscode: I think this test is redundant with line 402 (`Expect(executeErr).ToNot(HaveOccurred())`). We expect no errors only in the happy path case, so having a test that there are no errors when nothing is stubbed to return an error seems redundant.

api/cloudcontroller/ccv3/space_test.go
andymoe: Good catch. Specifying 0 warnings was a coding error, and the implementation had hardwired `nil` for the returned warnings.

api/cloudcontroller/ccv3/space.go
tcdowney: Follow-up PR: https://github.com/cloudfoundry/cli/pull/1652

PR #1639, opened 2019-05-07
Remove spinlock in PollStart()

seif-at-sap:
Remove spin lock in PollStart(). This significantly reduces the CPU time when waiting for apps to start.

v6

## Description of the Change

The `PollStart()` method selects on channels that might be closed (`appState, apiWarnings`, and `apiErrs`) during the PollStart() invocation. Since a select on a closed channel returns the channel type's zero value, the closed channels are always 'enabled'. In combination with the check for closed channels via booleans, this is basically a spin lock.
This PR disables the channels once they are closed by setting them to `nil`. This is semantically equivalent but much less heavy on the CPU.
I ran `cf push` with an app that would never report as ready with a profiler enabled and interrupted after about 20s (sig int / ctrl c).
As you can see in the attached profile, most of the time is spent in locking of `runtime/selectgo`
[master-a3840f49e3c0d1d3c1e390d7c222a0c564c3738a.pdf](https://github.com/cloudfoundry/cli/files/3153659/master-a3840f49e3c0d1d3c1e390d7c222a0c564c3738a.pdf)
With the patch applied, the CPU is almost completely idle:
[fix-261312ab4e642fb2c5dbcf7a7e51bf7b8af80853.pdf](https://github.com/cloudfoundry/cli/files/3153647/fix-261312ab4e642fb2c5dbcf7a7e51bf7b8af80853.pdf)

## Why Is This PR Valuable?

cf push should not make my CPU burn :)

Core functionality.

## Applicable Issues

None.

## How Urgent Is The Change?

Not urgent.

## Other Relevant Parties

No one.

### DIFF
--- a/command/v6/shared/poll_start.go
+++ b/command/v6/shared/poll_start.go
@@ -7,22 +7,23 @@ import (
 	"code.cloudfoundry.org/cli/command/translatableerror"
 )
 
-func PollStart(ui command.UI, config command.Config, messages <-chan *v2action.LogMessage, logErrs <-chan error, appState <-chan v2action.ApplicationStateChange, apiWarnings <-chan string, apiErrs <-chan error) error {
-	var breakAppState, breakWarnings, breakAPIErrs bool
-	for {
+func PollStart(ui command.UI, config command.Config, messages <-chan *v2action.LogMessage, logErrs <-chan error, appState <-chan v2action.ApplicationStateChange, apiWarnings <-chan string, apiErrs <-chan error) (apiError error) {
+	for appState != nil || apiWarnings != nil || apiErrs != nil {
 		select {
 		case message, ok := <-messages:
 			if !ok {
-				break
+				messages = nil
+				continue
 			}
 
 			if message.Staging() {
 				ui.DisplayLogMessage(message, false)
 			}
+
 		case state, ok := <-appState:
 			if !ok {
-				breakAppState = true
-				break
+				appState = nil
+				continue
 			}
 
 			switch state {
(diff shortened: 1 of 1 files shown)


abbyachau:
Hi @seif-at-sap thanks for creating this pull request. For changes to `push` or any other significant changes to core commands, we would appreciate a [heads up](https://github.com/cloudfoundry/cli/blob/master/.github/CONTRIBUTING.md#prerequisites) before a pull request is created - this helps us determine whether this is a feature/fix we want to support, and it ensures you are not working on pull requests which may not get merged. 

I have a few initial comments:
- what is the user pain you are experiencing?
- how would I be able to reproduce it in acceptance testing? if the pain point is that the CPU usage spikes when you `cf push`, under which conditions does this occur - please detail by stating real world examples and how a user would experience this pain point

Thanks, and let me know if you have any questions. We'll need more information before we can proceed. Thanks.

seif-at-sap:
> what is the user pain you are experiencing?

100% CPU from the cf binary. 

> how would I be able to reproduce it in acceptance testing? if the pain point is that the CPU usage spikes when you cf push, under which conditions does this occur - please detail by stating real world examples and how a user would experience this pain point

Try to push an app that never becomes ready, e.g.:
```
---
applications:
- name: never-ready
  no-route: true
  health-check-type: http
  docker:
    image: alpine:3.9
  command: sleep 100
```

I initially noticed this when I accidentally configured the wrong health check type for an application I was developing.

abbyachau:
Hi @seif-at-sap thanks for the quick response. We'll do a review of this pull request and will give you feedback as soon as we are able to. Thanks.

giner:
Same issue here. When an application is slow to stage or slow to start `cf` is using 100% of CPU.

Issue #1605, opened 2019-04-03
`cf copy-source` space searches are case-sensitive


cwlbraa:
**Describe the bug and the command you saw an issue with**
When specifying org AND space, if a lowercase `-s` parameter is used to try to copy-source in a space that is normally represented in all caps, the command will fail to find the space as follows:

```
Could not find space dev in organization system
```

**Expected behavior**
It should find the allcaps-persisted space.

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
```
cf create-space DEV
cf target -o system -s dev
cf push source-app
cf push target-app
cf copy-source source-app target-app -v -o system -s dev
```
The final command will return:
```
Could not find space dev in organization system
```
The bug looks like it's here:
https://github.com/cloudfoundry/cli/blob/f0b289bf9eb5e56fd8dccd2012f2dfc8b6c2ae06/cf/commands/application/copy_source.go#L149-L156

tjvman:
Was able to reproduce by following the provided steps. Marking as bug.

hemchander23:
Hi,

Thought I'd contribute to this issue. 

I started with analyzing the copy_source.go file as @cwlbraa mentioned. The quickest solution would be to use `strings.EqualFold(s.Name,targetSpace)`. However, `buildCopyString `would output the given space name (which may or may not be the *actual* space name)

https://github.com/hemchander23/cli/blob/da98b6335d2c1e87f1e82d939cec99a8bb19233c/cf/commands/application/copy_source.go#L191

Although this step would obtain the end result, I thought I would make it much more better.

I made use of `cmd.orgRepo.FindByName` and `cmd.spaceRepo.FindByNameInOrg` to obtain the actual org and space names which will be passed in the buildCopyString and also in subsequent operations. 

Let me know if this is good ? 

https://github.com/hemchander23/cli/blob/da98b6335d2c1e87f1e82d939cec99a8bb19233c/cf/commands/application/copy_source.go#L108-L124

In my humble opinion, I feel 
https://github.com/hemchander23/cli/blob/da98b6335d2c1e87f1e82d939cec99a8bb19233c/cf/commands/application/copy_source.go#L164-L189
could be redundant ?

Will be making necessary changes accordingly along with tests if you can kindly confirm if the approach is right.

```
[...]

PR #1650, opened 2019-05-20
Fix case sensitivity in 'copy_source' command while specifying space name.

hemchander23:
Fix for https://github.com/cloudfoundry/cli/issues/1605

## Description of the Change

The copy_source command currently is case-sensitive. As described in the linked Github bug, the CLI is sensitive to the case of the space name. Also, upon further notice, the CLI seems to output the targetSpace and targetOrg as such based on the case entered by the user instead of showing the actual names of the space/org. 

The current change addresses the case-sensitive space name issue and also prints the actual space and org name during execution.

## Why Is This PR Valuable?

Although this PR is tiny, it makes sure all users that make use of copy-source command  get to see the right space and org details regardless of the case in which they typed in(if the space/org exists) & get to see the error if not.

## Applicable Issues

Fixes https://github.com/cloudfoundry/cli/issues/1605

https://www.pivotaltracker.com/story/show/165097075

### DIFF
--- a/cf/commands/application/copy_source.go
+++ b/cf/commands/application/copy_source.go
@@ -3,7 +3,6 @@ package application
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	"code.cloudfoundry.org/cli/cf/api/applications"
 	"code.cloudfoundry.org/cli/cf/api/authentication"
@@ -108,29 +107,19 @@ func (cmd *CopySource) Execute(c flags.FlagContext) error {
 	var targetOrgName, targetSpaceName, spaceGUID, copyStr string
 	if targetOrg != "" && targetSpace != "" {
 		var org models.Organization
+		var space models.Space
+
 		org, err := cmd.orgRepo.FindByName(targetOrg)
 		if err != nil {
 			return err
 		}
-		//Get the matching space in the org by name
-		var space models.SpaceFields
-		var foundSpace bool
-		for _, s := range org.Spaces {
-			if strings.EqualFold(s.Name, targetSpace) {
-				space = s
-				foundSpace = true
-			}
-		}
-		//If no matching space is found
-		if !foundSpace {
-			return errors.New(T("Could not find space {{.Space}} in organization {{.Org}}",
-				map[string]interface{}{
-					"Space": terminal.EntityNameColor(targetSpace),
-					"Org":   terminal.EntityNameColor(targetOrg),
-				},
-			))
-		}
+
+		space, err = cmd.spaceRepo.FindByNameInOrg(targetSpace,org.GUID)
 		spaceGUID = space.GUID
+		if err != nil {
+			return err
+		}
+
 		targetOrgName = org.Name
 		targetSpaceName = space.Name
 	} else if targetSpace != "" {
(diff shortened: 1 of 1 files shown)


hemchander23:
One of the tests might fail due to https://github.com/hemchander23/cli/blob/dev/cf/commands/application/copy_source.go#L134

since the actual org-name may / may not be equal to the one that the user provided (in terms of case)

the test passes if `targetOrgName = targetOrg` which defeats the purpose of getting the actual org name & from my perspective the org/space that is shown has to correspond to the actual ones. 

Pl. correct me if i am wrong.

abbyachau:
Hi @hemchander23 thanks for creating this pull request! Appreciate the help for this issue. We will try to get back to you as soon as we can with feedback on your pr. Thanks again.

hemchander23:
Thanks @abbyachau ! Will be glad to hear the feedback. Will incorporate the changes accordingly.  :)

abbyachau:
Hi @hemchander23 thanks again for this. Regarding the following:
```
the CLI seems to output the targetSpace and targetOrg as such based on the case entered by the user instead of showing the actual names of the space/org.
```

Please could you expand on this, and perhaps provide an example of the issue and the desired change. Thanks.

hemchander23:
Hi @abbyachau 

When I execute the command, it displays the org and space name as such while the actual names could be different (just case-wise)

the PR ensures that the names displayed post-execution in the logs show the names as seen while running a `cf spaces `or` cf orgs `command

*Before the change*

BeforeChange

BeforeChangeorgsspaces

_after the change_

AfterChange

AfterChangeorgspaces

hemchander23:
Hi @abbyachau Lemme know if any additional changes need to be done. Would be happy to help ! :)

staylor14:
Hi @hemchander23 !

Thanks so much for this pull request, and apologies for the delay in responding!

First, the basis of your solution (using EqualFold in our string comparison) looks great.  We'd prefer if you kept the logic for finding the Space in `findSpaceGUID()`.  Although no other code calls into this method, we believe it improves the readability of the already fraught `Execute()` :)

Next, it would be great if you could run the unit tests prior to pushing your change.  It looks like something broke (see the Travis CI build in this PR).

Finally, we wouldn't be able to merge this change without added unit tests that exercise the expected behavior around capitalization.  It it were us, we would probably add a "happy path" test [here](https://github.com/cloudfoundry/cli/blob/9f7b538b5f1a82fa530349f8e75b0e34ae8250b2/cf/commands/application/copy_source_test.go#L238) that would confirm that we can find a space without matching capitalization. This test would fail given the current implementation, and would pass given your fix.

Finally FINALLY (heh) it would be doing us a favor if you could squash your commits so that the merge will result in a "single" change.  And, if we're honest, it'll make Brendan feel better :)

Thanks again!
Steve && @bwasmith

bwasmith:
Hi @hemchander23 ,

We are going to move forward on a fix for this issue. 

We will use your suggested change. But, given the lack of revisions to original PR and the state of our unit tests, we cannot merge the changes as written. 

We plan to make these changes ourselves and close this PR. Please respond ASAP if you plan on addressing our requested changes soon. 

Again, apologies for our initial delay in response. We hope that you continue to make contributions to the cf-cli!

Best,
Brendan

acrmp:
Hi @hemchander23,

Thanks very much for your PR! We made a similar change which is now on `master` and should be
in the next release of the CLI.

Thanks for contributing to the CF CLI!

PR #1652, opened 2019-05-20
Refactor set-label commands to use a shared api layer

tcdowney:

This modifies the **v7** CLI.

## Description of the Change

The APIs for updating metadata on API resources in Cloud Controller
follow a consistent interface. This means that apart from the endpoint
URL, the structure of the PATCH requests to these endpoints is uniform
so we should be able to share much of the API layer across resources
when adding/updating labels (and later annotations).

See the following API docs:
https://v3-apidocs.cloudfoundry.org/version/3.71.0/index.html#metadata

## Why Is This PR Valuable?

This should simplify the addition of new supported resources for `cf set-label`.

It affects core cf cli functionality.

## Applicable Issues

This refactor comes from the following discussion in PR #1631:
https://github.com/cloudfoundry/cli/pull/1631#discussion_r280255971

CAPI Story:
[#165803403](https://www.pivotaltracker.com/story/show/165803403)

## How Urgent Is The Change?

Not urgent, but if you like this approach I believe it will simplify future label stories.

## Other Relevant Parties

I solo-ed on this, but want to tag some CAPIs that may have worked on `cf set-label` previously / be interested:
@monamohebbi @ericpromislow @cwlbraa

### DIFF
--- a/actor/v7action/label_test.go
+++ b/actor/v7action/label_test.go
@@ -11,7 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("UpdateApplicationLabelsByApplicationName", func() {
+var _ = Describe("Labels", func() {
 	var (
 		actor                     *Actor
 		fakeCloudControllerClient *v7actionfakes.FakeCloudControllerClient
@@ -21,6 +21,7 @@ var _ = Describe("UpdateApplicationLabelsByApplicationName", func() {
 		executeErr                error
 		resourceName              string
 		spaceGUID                 string
+		orgGUID                   string
 		labels                    map[string]types.NullString
 	)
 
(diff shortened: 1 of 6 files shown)


actor/v7action/label_test.go
```diff
@@ -172,4 +177,76 @@ var _ = Describe("UpdateApplicationLabelsByApplicationName", func() {
 			})
 		})
 	})
+
+	Context("UpdateSpaceLabelsBySpaceName", func() {
```
tcdowney: Backfilled this test.

api/cloudcontroller/ccv3/metadata.go
```diff
@@ -1,10 +1,67 @@
...
+func (client *Client) UpdateResourceMetadata(resource string, resourceGUID string, metadata Metadata) (ResourceMetadata, Warnings, error) {
+	metadataBtyes, err := json.Marshal(ResourceMetadata{Metadata: &metadata})
+	if err != nil {
+		return ResourceMetadata{}, nil, err
+	}
+
+	var request *cloudcontroller.Request
+
+	switch resource {
+	case "app":
```
tcdowney: All of these tests just switch on endpoint and the URI Params. I wasn't sure what the most idiomatic way to do this was in Go so would appreciate any suggestions here. 🙏

PR #1653, opened 2019-05-21
manifests include sidecars

drnic:

Both; though my dev/test is with `cf v3-apply-changes` and `cf v3-push`.

## Description of the Change

`cf v3-apply-changes` recently added support for `sidecars:` in manifests. 

Unfortunately, the `utils/manifests` library does not parse nor print them yet. I need this library so I can modify the test app's manifest, edit its app name, and save it to a temporary file. The rest of the buildpack testing process is managed by libbuildpack.

## Why Is This PR Valuable?

It is required for buildpack testing (via libbuildpack) so we can mutate the manifest.yml to edit the appname, etc. Currently when the manifest is printed it does not include sidecars from the original manifest.

## How Urgent Is The Change?

Relatively urgent to go with the sidecars epic.

## Other Relevant Parties

* Buildpacks team
* Anyone involved in Sidecars

### DIFF
--- a/util/manifest/application.go
+++ b/util/manifest/application.go
@@ -36,6 +36,7 @@ type Application struct {
 	Routes          []string
 	RoutePath       string
 	Services        []string
+	Sidecars        []interface{}
 	StackName       string
 
 	DeprecatedDomain     interface{}
@@ -59,6 +60,7 @@ func (app Application) MarshalYAML() (interface{}, error) {
 		NoRoute:                 app.NoRoute,
 		Path:                    app.Path,
 		Services:                app.Services,
+		Sidecars:                app.Sidecars,
 		StackName:               app.StackName,
 		Timeout:                 app.HealthCheckTimeout,
 	}
@@ -129,6 +131,7 @@ func (app *Application) UnmarshalYAML(unmarshaller func(interface{}) error) erro
 	app.Path = m.Path
 	app.RandomRoute = m.RandomRoute
 	app.Services = m.Services
+	app.Sidecars = m.Sidecars
 	app.StackName = m.StackName
 	app.HealthCheckTimeout = m.Timeout
 	app.EnvironmentVariables = m.EnvironmentVariables
--- a/util/manifest/raw_manifest_application.go
+++ b/util/manifest/raw_manifest_application.go
@@ -23,6 +23,7 @@ type rawManifestApplication struct {
 	RandomRoute             bool               `yaml:"random-route,omitempty"`
 	Routes                  []rawManifestRoute `yaml:"routes,omitempty"`
 	Services                []string           `yaml:"services,omitempty"`
+	Sidecars                []interface{}      `yaml:"sidecars,omitempty"`
 	StackName               string             `yaml:"stack,omitempty"`
 	Timeout                 uint64             `yaml:"timeout,omitempty"`
 }


tcdowney:
@drnic 
I'm not sure if you're trying to generate a manifest that includes sidecars, but if you are you can use the server-side generation endpoint.

http://v3-apidocs.cloudfoundry.org/version/3.71.0/index.html#generate-an-app-manifest

drnic:
@tcdowney see https://github.com/cloudfoundry/libbuildpack/pull/30/files#diff-74f5873222a036aea0e8b7675cfeb7f0R416 for how I'm using this PR - its to allow mutating manifest.yml to inject app name, whilst not losing my sidecar bits and pieces

abbyachau:
Hi @drnic the side car manifest support is done server side now (CAPI owns the manifest in V7 CLI) - see http://v3-apidocs.cloudfoundry.org/version/3.70.0/index.html#app-manifest. There are no plans to support sidecars in the v6 CLI. I'm leaving this pr open for further comments/feedback, but will be closing it as `won't-do`. Thanks.

drnic:
If you’re confused about the PR please ask for clarification! This is a PR for a libarary within the CLI.


________________________________

drnic:
I'm very confused about contributing even the tiniest patches to CF
projects sometimes.

drnic:
Why is file `util/manifest/application.go` considered v6 only??? What is going on?

XenoPhex:
Hey there @drnic,

Over the past year or so, the CF CLI (and CCNG) has been transitioning to the CC V3 API; in order to convert the CLI over iteratively, the CLI and VAT teams have had to mess the internal architecture quite a bit in order to support this transition. If you read the [opening section of the Contributing Documentation](https://github.com/cloudfoundry/cli/blob/master/.github/CONTRIBUTING.md#cli-v6x--v7-beta) _some_ of this is laid out. While it doesn't explain enough for this particular change to not be confusing, it would have at least provided you some context around the current state of the code.

As for not merging this change, there's several things that you should be made aware of:
- Starting with CLI V7 / CC API V3, the ownership of the "app manifest" has transitioned from the CLI to the CC API. This is to allow the CC API team to update/add features to the manifest and have them instantly be usable by the consumer upon API upgrade. (As opposed to the V6 CLI manifest, where an API feature would be added and it would require the CLI team to manually implement said feature in a given version of the CLI.)
- This results in the V3-Prefixed commands/V7 CLI knowing a limited amount about "app manifest" so that the CC team can change the manifest without fear of breaking the CLI. If you're wondering what that looks like, [here's what the CLI now reads](https://github.com/cloudfoundry/cli/blob/v6.45.0/util/manifestparser/application.go). As you can see, it's a much smaller set of information than the V6 version of the `manifest` package.

As a side note, one thing I would recommend against is using CLI packages external to the CF CLI. Outside of the `plugins` and [eventually] the `api` packages, all other packages in the CLI are _not_ designed to be used outside of the CLI. Generally when we get pull requests that involve using them outside of the CLI, we reject them outright unless it also supports an internal use case. In this case, `sidecars` are already supported by the V7 cli via the API, so no CLI code changes are necessary.

abbyachau:
Hi @drnic I'm going to close this issue but please do not hesitate to reach out if you have feedback. We hope to improve the contribution process and we would be open to a conversation about how best we can help our contributor community.

Issue #1694, opened 2019-06-20
`cf uups <svc>` without params deletes values


spgreenberg:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [X] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [X] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [X] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
If you run `cf uups <some-service>` without any flags (like -p), existing key/values are overwritten. This is in contrast to the message displayed by the CLI "No flags specified. No changes were made."

**What happened**
The existing key/values are removed from the UPSI.

**Expected behavior**
The existing key/values should not be changed.

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:

1. Be sure you have an app (any app) available to bind a service instance to.

2. Create a UPSI with key/values: `cf cups upsi -p '{"some-key":"some-value"}'`
3. Bind the UPSI to an app: `cf bs app upsi`
4. Observe the VCAP_SERVICES envar to ensure key/values in the credentials block are correct. `cf env app`

```
...
{
 "VCAP_SERVICES": {
  "user-provided": [
   {
    "binding_name": null,
    "credentials": {
     "some-key": "some-value"
    },
    "instance_name": "upsi",
    "label": "user-provided",
    "name": "upsi",
    "syslog_drain_url": "",
    "tags": [],
    "volume_mounts": []
   }
  ]
 }
}
...
```

5. Run update UPSI without parameters and observe the "No changes" message: `cf uups upsi`

```
Updating user provided service upsi in org blah / space blah as blah...
OK
TIP: Use 'cf restage' for any bound apps to ensure your env variable changes take effect
No flags specified. No changes were made.
```

6. Observe the key/values in the credentials block are now missing: `cf env app`

```
...
{
 "VCAP_SERVICES": {
  "user-provided": [
   {
    "binding_name": null,
    "credentials": {},
    "instance_name": "upsi",
    "label": "user-provided",
    "name": "upsi",
    "syslog_drain_url": "",
    "tags": [],
    "volume_mounts": []
   }
  ]
 }
}
...
```

**Provide more context**

MacOS
cf version 6.45.0+5f9ff16f9.2019-06-03
Using Pivotal Web Services

CC: @mxplusb

[...]

PR #1724, opened 2019-07-11
update-user-provided-service should not unset values

blgm:
- Issue https://github.com/cloudfoundry/cli/issues/1694 describes an
incorrect behaviour where update-user-provided-service can unset existing
credentials associated with a service even when the -p flags is not
specified. We should fix that so that values are not changed unless the
associated flag is specified. Unsetting a value should be done by
explicitly setting in to an empty string.
- In the process we refactored the update-user-provided-service command
to be consistent with the v6 commands

Co-authored-by: Niki Maslarski <nikolay.maslarski@sap.com>


abbyachau:
Hey @blgm thanks for this. Did you folks fix the bug on v6 after you refactored the command? or is the bug fix only on v7? Sorry i'm sure this information is listed somewhere. thanks.

blgm:
Hi @abbyachau, we fixed the bug and refactored the command to v6 at the same time.

PR #1729, opened 2019-07-16
 Add ability to use `cf set-label` to add labels to buildpacks

tcdowney:

This modifies **v7**!

## Description of the Change

We're adding the ability to use `cf set-label` to add labels to buildpacks!
See: https://www.pivotaltracker.com/story/show/166986079

We also included an additional commit since the stories so far have been calling for things like:
> The required arg resource type (space) is case insensitive
> The required arg buildpack is case insensitive.

And this wasn't actually occurring.

## Other Relevant Parties

@ssisil @Gerg @abbyachau @selzoc @sannidhi

### DIFF
--- a/api/cloudcontroller/ccv3/metadata.go
+++ b/api/cloudcontroller/ccv3/metadata.go
@@ -20,7 +20,7 @@ type ResourceMetadata struct {
 }
 
 func (client *Client) UpdateResourceMetadata(resource string, resourceGUID string, metadata Metadata) (ResourceMetadata, Warnings, error) {
-	metadataBtyes, err := json.Marshal(ResourceMetadata{Metadata: &metadata})
+	metadataBytes, err := json.Marshal(ResourceMetadata{Metadata: &metadata})
 	if err != nil {
 		return ResourceMetadata{}, nil, err
 	}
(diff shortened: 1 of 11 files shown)


acrmp:
@abbyachau @ericpromislow If there are two buildpacks but one does not have a stack assigned we will update the buildpack with the null stack. I think this is the expected behaviour?

```
$ mkdir /tmp/example-buildpack
$ touch /tmp/example-buildpack/required-file
$ cf create-buildpack example-buildpack /tmp/example-buildpack 1
$ cf update-buildpack example-buildpack --assign-stack cflinuxfs3
$ cf create-buildpack example-buildpack /tmp/example-buildpack 1
$ cf set-label buildpack example-buildpack some-key=some-value
```

acrmp:
@abbyachau @ericpromislow This [comment on the original story](https://www.pivotaltracker.com/story/show/166986079/comments/204634225) suggests we were intending to single quote the name of the buildpack when it was not found?

```
$ cf set-label buildpack missing-buildpack foo=bar
Setting label(s) for buildpack missing-buildpack as admin...
Buildpack missing-buildpack not found
FAILED
```

ericpromislow:
The buildpack name isn't single-quoted because `https://github.com/cloudfoundry/cli/blob/e03c9693b66f403a7bbc7b16fe931d3b239a22f8/command/translatableerror/buildpack_not_found_error.go#L8` doesn't quote them. Most of the other `Error` methods in `command/translatableerror` single-quote names of resources.

Maybe best to create a separate story to quote bad buildpack names?

abbyachau:
Hi there, I can create a separate story to quote the error for `update-buildpack`, thanks.

Currently the `update-buildpack` command, if there are two buildpacks, one with a stack, and one with a nil stack - if the `--stack` flag is not passed, the command will update the one with the `nil` stack. It looks like you folks updated the `set-label` command for `buildpacks` to do the same, which is great.

acrmp:
@ericpromislow @tcdowney Thanks for this PR. It looks great!

We couldn't see a test for the [case described above](https://github.com/cloudfoundry/cli/pull/1729#issuecomment-514830013) when there are multiple buildpacks and one has a null stack. Would you be able to add a test for this case?

api/cloudcontroller/ccv3/metadata.go
```diff
@@ -34,6 +34,12 @@ func (client *Client) UpdateResourceMetadata(resource string, resourceGUID strin
 			Body:        bytes.NewReader(metadataBtyes),
 			URIParams:   map[string]string{"app_guid": resourceGUID},
 		})
+	case "buildpack":
+		request, err = client.newHTTPRequest(requestOptions{
+			RequestName: internal.PatchBuildpackRequest,
+			Body:        bytes.NewReader(metadataBtyes),
```
acrmp: Can we correct the spelling of `metadataBytes` while we're making this change.

ericpromislow: I've got something on now, but will do. Thanks for catching that

ericpromislow: I'll have to work on that test on Friday (PDT). Want to get all the combinations right.

ericpromislow:
Integration tests have been backfilled to handle multiple stack-scoped and unscoped buildpacks.

Please keep this PR open while we also add https://www.pivotaltracker.com/story/show/166986428/comments/204879378 for both set-label and get-labels

PR #1732, opened 2019-07-18
Show buildpack labels

sannidhi:
## WARNING: 

This PR builds on top of https://github.com/cloudfoundry/cli/pull/1729 which adds the set-label functionality for a buildpack. Please make sure it is merged first.

Pertinent commits:

[this commit](https://github.com/cloudfoundry/cli/pull/1732/commits/a41b8a3f9f3b1fca2f2a595536ad1f66dce6df7b)

This branch was based off of master, not set-label-buildpack-166986079, so the pertinent diff command is this:

Diff for this PR on its actual base branch:

`git diff master origin/show-buildpack-labels`

v7

## Description of the Change

App operator should be able to use a new CLI command `cf labels` to view labels for a buildpack 

Tracker Story: https://www.pivotaltracker.com/story/show/166986428

### DIFF
--- a/command/v7/labels_command.go
+++ b/command/v7/labels_command.go
@@ -30,11 +30,13 @@ type LabelsActor interface {
 	GetApplicationLabels(appName string, spaceGUID string) (map[string]types.NullString, v7action.Warnings, error)
 	GetOrganizationLabels(orgName string) (map[string]types.NullString, v7action.Warnings, error)
 	GetSpaceLabels(spaceName string, orgGUID string) (map[string]types.NullString, v7action.Warnings, error)
+	GetBuildpackLabels(buildpackName string, buildpackStack string) (map[string]types.NullString, v7action.Warnings, error)
 }
 
 type LabelsCommand struct {
 	RequiredArgs flag.LabelsArgs `positional-args:"yes"`
-	usage        interface{}     `usage:"CF_NAME labels RESOURCE RESOURCE_NAME\n\nEXAMPLES:\n   cf labels app dora \n\nRESOURCES:\n   app\n   space\n   org\n\nSEE ALSO:\n   set-label, unset-label"`
+	StackName    string          `long:"stack" short:"s" description:"required when more than one buildpack has the same name"`
+	usage        interface{}     `usage:"CF_NAME labels RESOURCE RESOURCE_NAME\n\nEXAMPLES:\n   cf labels app dora \n\nRESOURCES:\n   app\n   buildpack\n   org\n   space\n\nSEE ALSO:\n   set-label, unset-label"`
 	UI           command.UI
 	Config       command.Config
 	SharedActor  command.SharedActor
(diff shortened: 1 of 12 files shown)


command/v7/labels_command.go
```diff
@@ -130,6 +138,22 @@ func (cmd LabelsCommand) fetchSpaceLabels(username string) (map[string]types.Nul
 	return cmd.Actor.GetSpaceLabels(cmd.RequiredArgs.ResourceName, cmd.Config.TargetedOrganization().GUID)
 }
 
+func (cmd LabelsCommand) fetchBuildpackLabels(username string) (map[string]types.NullString, v7action.Warnings, error) {
+	err := cmd.SharedActor.CheckTarget(false, false)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cmd.UI.DisplayTextWithFlavor("Getting labels for buildpack {{.BuildpackName}} as {{.Username}}...", map[string]interface{}{
```
acrmp: It looks like there was a request on the story (after this PR was made) to include the stack name in the message shown to the user for consistency with other commands:
https://www.pivotaltracker.com/story/show/166986428/comments/204879378

command/v7/set_label_command_test.go
```diff
@@ -30,6 +32,21 @@ var _ = Describe("set-label command", func() {
 		executeErr error
 	)
 
+	checkBuildpackArgFunc := func() {
```
acrmp: I can see we're trying to remove duplication in the test here. I don't have a strong opinion either way.

Perhaps we could rename this function to make it clearer that we are expecting an error?

acrmp:
@sannidhi @ericpromislow Thanks for this PR. It looks great.

Relevant to the conversation on #1729 It seems like we introduced quoting of the buildpack and stack names but then subsequently backed that out.

command/v7/labels_command.go
ericpromislow: The PR is still open; we can do that

command/v7/set_label_command_test.go
ericpromislow: Sure. I see that the upcoming annotation stories roll all five resource types into one story. I'm not big on lots of DRYing in tests, but it's likely to happen. Let me know if you prefer everything unrolled.

ericpromislow:
> @sannidhi @ericpromislow Thanks for this PR. It looks great.
> 
> Relevant to the conversation on #1729 It seems like we introduced quoting of the buildpack and stack names but then subsequently backed that out.

Correct; that was our mistake. We didn't run all the integration tests, but when I ran global/v7 I got failures and backed the change out (in a later PR unfortunately)

command/v7/set_label_command_test.go
ericpromislow: Rename was done, and the same function was added to `command/v7/labels_command_test.go`

PR #1799, opened 2019-09-20
delete-space returns 0 when space not found

XenoPhex:

Made the change in both V6 and V7 of delete space so that they're both aligned.

## Description of the Change

Aligns the `delete-space` "does not exists" behavior with the rest of the `delete-*` commands

## Why Is This PR Valuable?

Makes it easier to script the `delete-space` command as the space not existing does not cause an "error." Meaning I can have my script continue cleanup/teardown/etc. if the space was already cleaned up.

This addition aligns the `delete-space` command with the rest of core.

## Applicable Issues

N/A

## How Urgent Is The Change?

Medium - It's causing some flakieness in the rewrite of smoke-tests I'm working on. Not urgent, but not _not_ urgent.

## Other Relevant Parties

This'll affect everyone, but in a good way. There's a good chance a lot of people will find this helpful, so it's worth calling out in a the release notes as a "long running issue that's been recently fixed."

## Other notes
- Integration tests were added for both V6 and V7
- I was unable to run all the integration tests for V6 due to my test stack having issues, but all the V6 `delete-space` tests did pass, not sure about the rest of the suite though.
- I was unable to integration test the V7 suite as I didn't have a new enough cloud controller and when I tried to upgrade, I caused the issues mentioned above. The V7 tests were identical to the V6 tests however, so I made identical changes and they should pass given a working CC.


abbyachau:
Thank you @XenoPhex, we'll review soon. Appreciate the contribution.

reid47:
Hey @XenoPhex, thanks for the PR. We reviewed it and we would rather not make this change to V6 code, since it'll be a breaking change. Could you please remove the V6 changes? For V7, we're happy to merge it and call out the breaking change in the release notes.

XenoPhex:
Looks like the work was done in https://github.com/cloudfoundry/cli/commit/9fc648305237930fd02f39a60158f1f44fdb254b, closing the PR.

Thanks y'all!

Issue #1913, opened 2020-03-18
cf update-service remove tags when no tags are provided


RageZBla:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.
- [x] ~~If this is an issue for the v7 beta release, I've read through the [official docs](https://docs.cloudfoundry.org/cf-cli/v7.html) and the [release notes](https://github.com/cloudfoundry/cli/releases).~~

**Describe the bug and the command you saw an issue with**
Updating a service parameters using `cf update-service` would remove all the tags on a given service. 

**What happened**
Tags are removed after the service is updated. 

**Expected behavior**
Tags should be kept.

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. check given service has tags
``` 
$ cf service rabbit_test_tag
Showing info of service rabbit_test_tag in org org / space dev as admin...

name: rabbit_test_tag
service: p.rabbitmq
tags: tag initial
plan: solo-3.7
description: RabbitMQ service to provide dedicated instances of this high-performance multi-protocol messaging broker
documentation: https://docs.pivotal.io/rabbitmq-cf/
dashboard: <redacted>
service broker: rabbitmq-odb

Showing status of last operation from service rabbit_test_tag...

status: update succeeded
message:
started: 2020-03-17T14:38:02Z
updated: 2020-03-17T14:38:02Z

There are no bound apps for this service.

Upgrades are not supported by this broker.
```
2. Update service parameters `cf update-service rabbit_test_tag -c '{"tls": true}'`
3. Review the service instance, tags have been removed

```
$ cf service rabbit_test_tag
Showing info of service rabbit_test_tag in org org / space dev as admin...

name: rabbit_test_tag
service: p.rabbitmq
tags:   <==== no more tags
plan: solo-3.7
description: RabbitMQ service to provide dedicated instances of this high-performance multi-protocol messaging broker
documentation: https://docs.pivotal.io/rabbitmq-cf/
dashboard: <redacted>
service broker: rabbitmq-odb

Showing status of last operation from service rabbit_test_tag...

status: update in progress
message: Instance update in progress
started: 2020-03-17T14:38:58Z
[...]

PR #1919, opened 2020-03-28
Correct an issue with the update-service command.

RageZBla:

V6

## Description of the Change

The update-service command was removing the tags on service when no tags were provided. Correct this issue and allow the users to keep the tags without providing them on every invocation.

## Why Is This PR Valuable?

This corrects a bug. 

This is a bug in the core

## Applicable Issues

#1913 

## How Urgent Is The Change?

Not urgent

## Other Relevant Parties

N/A

### DIFF
--- a/cf/api/apifakes/fake_service_repository.go
+++ b/cf/api/apifakes/fake_service_repository.go
@@ -227,13 +227,13 @@ type FakeServiceRepository struct {
 	renameServiceReturnsOnCall map[int]struct {
 		result1 error
 	}
-	UpdateServiceInstanceStub        func(string, string, map[string]interface{}, []string) error
+	UpdateServiceInstanceStub        func(string, string, map[string]interface{}, *[]string) error
 	updateServiceInstanceMutex       sync.RWMutex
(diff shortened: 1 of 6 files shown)


RageZBla:
#1913

JenGoldstrich:
Hey @RageZBla! Thanks for filing this PR, we have this ready to be discussed at our next IPM and should be able to prioritize merging this in soon.

Thank you so much!
Jenna

JenGoldstrich:
Hey @RageZBla I am closing this PR, it actually updates code in the cf/ subdirectory, which is code that is unaccessible in v6 currently and scheduled to be removed, we needed to update the v6/update_service_command.go file, the team has prioritzed a fix and I picked this up today and you can track it here https://www.pivotaltracker.com/story/show/171854111 

Thanks!
Jenna

JenGoldstrich:
Re-opening this because I was wrong, there appears to be weird logic that falls back on /cf/ commands when we fail to process tags in the v6 update-service-command, @staylor14 and I are gonna test our integration tests on this and merge if they pass

PR #1922, opened 2020-04-01
Add cf pull

0401-7r011:
> Okay

### DIFF
--- a/actor/v7action/cloud_controller_client.go
+++ b/actor/v7action/cloud_controller_client.go
@@ -2,6 +2,7 @@ package v7action
 
 import (
 	"io"
+	"os"
 
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3"
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant"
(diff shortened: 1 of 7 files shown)


JenGoldstrich:
Hello @0401-7r011 !

Thank you for submitting this pull request, a few things I'm curious about.

You mentioned that this change would be applied to v6 and v7, but in your changelog I am only seeing the v7 command file.

I do not see any test changes for your submission, did you write a test for the pull command?

I am a little skeptical about this use case though, the cf CLI is not a source control management software, why would someone need to retrieve their code from CF after a push, with good CI/CD pipeline practices I would rely on git and my pipeline to track commits and retrieve code at specific pushes, could we get more details about this use case?

Thanks again!
Jenna and the cf CLI team

waterlink:
@JenGoldstrich I think this was a 1st of April’s (Fool’s Day) joke, and not a real feature request. The person’s account name even has `04.01` in it :D

staylor14:
Hilarious April Fool's joke that illustrates (again) how harmful April Fools' day is.  "Fake news" jokes sure haven't backfired, and now I'm forced to read, attempt to understand, and then reject "joke" assistance from our community.

What a pleasure to now have to think, "Wait, is this a joke?" for every PR that I think may not have a ton of technical merit.

What fun!

PR #1951, opened 2020-06-09
refactor: use resources pattern for service instances

blgm:
[#172178942]

### DIFF
--- a/resources/service_instance_resource_test.go
+++ b/resources/service_instance_resource_test.go
@@ -1,0 +1,42 @@
+package resources_test
+
+import (
+	"encoding/json"
+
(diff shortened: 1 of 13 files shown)


resources/service_instance_resource_test.go
```diff
@@ -0,0 +1,42 @@
...
+							"guid": "fake-space-guid"
+						}
+					}
+				}
+            }`))
+		})
+	})
+
+	Describe("UnmarshalJSON", func() {
+		It("unmarshals the guid", func() {
```
pivotal-marcela-campo, line 35: shouldn't we test it unmarshalls guid and name?

PR #1976, opened 2020-07-15
V8 create service validation of arguments #173399121

pivotal-marcela-campo:
https://www.pivotaltracker.com/story/show/173399121

Only the [last commit](https://github.com/cloudfoundry/cli/pull/1976/commits/1d2fd81b0d60dc087b03e52b14d9d89eae5a5b6b) is pending review. The previous commits where reviewed [here](https://github.com/cloudfoundry/cli/pull/1973) but cannot be merged until the command is finished.


command/v7/create_service_command_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"code.cloudfoundry.org/cli/util/ui"
+	"errors"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+)
+
+var _ = Describe("create-service Command", func() {
+	var (
+		cmd             *v7.CreateServiceCommand
```
blgm, line 20: The command has a value receiver, but in the tests we are using a pointer. It would be better to be consistent, and I think other commands generally go with a value approach.

command/translatableerror/service_plan_not_found_error.go
```diff
@@ -1,20 +1,26 @@
 package translatableerror
```
blgm, line 1: Some translatable errors have unit tests. This isn't a strict requirement, but can be a good way to test all the possibilities.

blgm: Also, is this error actually used?

pivotal-marcela-campo: Thanks! forgot to update them. 

It is I am afraid the one actually used for printing the message. The command parser converts to translatable error before printing to stderr. I could remove the mapping and the error altogether and then it would just use the one in actionerror package. But... this is used by v2 as well. Do you think we should remove it now that we are using branches instead of flags?

pivotal-marcela-campo:
Merged by https://github.com/cloudfoundry/cli/pull/1997

PR #1978, opened 2020-07-15
v8(services): display parameters in service command

blgm:
[#171954640](https://www.pivotaltracker.com/story/show/171954640)

### DIFF
--- a/actor/v7action/service_instance_test.go
+++ b/actor/v7action/service_instance_test.go
@@ -159,62 +159,60 @@ var _ = Describe("Service Instance Actions", func() {
 				)
 			})
 
-			It("returns a service instance with details and warnings", func() {
-				By("returning warnings and no errors", func() {
-					Expect(executionError).NotTo(HaveOccurred())
-					Expect(warnings).To(ConsistOf(
-						"some-service-instance-warning",
-						"some-parameters-warning",
-					))
-				})
+			It("returns warnings and no errors", func() {
+				Expect(executionError).NotTo(HaveOccurred())
+				Expect(warnings).To(ConsistOf(
+					"some-service-instance-warning",
+					"some-parameters-warning",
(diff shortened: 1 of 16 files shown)


command/v7/service_command_test.go
```diff
@@ -368,12 +374,83 @@ var _ = Describe("service command", func() {
 			})
 		})
 
+		When("the was a problem retrieving the parameters", func() {
```
pivotal-marcela-campo, line 377: "there was a problem..."

integration/v7/isolated/service_command_test.go
```diff
@@ -12,7 +13,7 @@ import (
 	. "github.com/onsi/gomega/gexec"
 )
 
-var _ = Describe("service command", func() {
+var _ = FDescribe("service command", func() {
```
pivotal-marcela-campo, line 16: Remove focus

actor/v7action/service_instance_test.go
```diff
@@ -133,48 +134,70 @@ var _ = Describe("Service Instance Actions", func() {
...
+							ParametersMissingReason: "",
 						},
-						ServiceOffering: resources.ServiceOffering{
-							Name:             serviceOfferingName,
-							Description:      serviceOfferingDescription,
-							DocumentationURL: serviceOfferingDocumentation,
+					))
+				})
+
+				By("making the correct call to get the service instance", func() {
```
pivotal-marcela-campo, line 176: Just an opinion regarding the use of By. I find it more readable to have separate It blocks. I see this worth it in an integration test where set up and tear down are so costly. But cli unit test run so fast that I'd rather have the separate It. 
Also, passing a function to By makes it so verbose, maybe just leaving the function out (as it is optional) would reduce the clutter?

PR #1980, opened 2020-07-16
v8(services): rename service command

blgm:
[#173416148](https://www.pivotaltracker.com/story/show/173416148)
[#173416145](https://www.pivotaltracker.com/story/show/173416145)
[#173416373](https://www.pivotaltracker.com/story/show/173416373)

### DIFF
--- a/command/v7/rename_service_command.go
+++ b/command/v7/rename_service_command.go
@@ -1,0 +1,63 @@
+package v7
+
+import (
+	"code.cloudfoundry.org/cli/actor/actionerror"
+	"code.cloudfoundry.org/cli/command/flag"
+	"code.cloudfoundry.org/cli/command/translatableerror"
+)
+
+type RenameServiceCommand struct {
+	BaseCommand
+	RequiredArgs    flag.RenameServiceArgs `positional-args:"yes"`
+	relatedCommands interface{}            `related_commands:"services, update-service"`
+}
+
+func (cmd RenameServiceCommand) Execute(args []string) error {
+	if err := cmd.SharedActor.CheckTarget(true, true); err != nil {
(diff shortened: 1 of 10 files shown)


command/v7/rename_service_command.go
```diff
@@ -0,0 +1,63 @@
...
+	case nil:
+	case actionerror.ServiceInstanceNotFoundError:
+		cmd.UI.DisplayText("TIP: Use 'cf services' to view all services in this org and space.")
+		return translatableerror.ServiceInstanceNotFoundError{Name: e.Name}
+	default:
+		return err
+	}
+
+	cmd.UI.DisplayOK()
+	return nil
```
butzopower, line 41:
```suggestion
	switch e := err.(type) {
	case nil:
		cmd.UI.DisplayOK()
		return nil
	case actionerror.ServiceInstanceNotFoundError:
		cmd.UI.DisplayText("TIP: Use 'cf services' to view all services in this org and space.")
		return translatableerror.ServiceInstanceNotFoundError{Name: e.Name}
	default:
		return err
	}
```

Not a change suggestion, but wondering if one of these is more idiomatic Go than the other?

blgm: I did think of doing it that way, but thought the code reviewer might not like it. But as the the code reviewer has actively suggested it, I'll make that simplification!

blgm: I did consider doing it that way. The pro is that the switch statement handles all the possibilities. The con is that the switch statement _looks like_ error handling, when it's actually doing success handling too. But actually the `case nil` could be confusing, so maybe it makes sense to actually do something in this case.

PR #1991, opened 2020-07-22
V8 create service command 173400287

pivotal-marcela-campo:
https://www.pivotaltracker.com/story/show/173399121

This implements happy path and v3 api errors when creating the instance

Only the [last commit](https://github.com/cloudfoundry/cli/commit/a1bff8c888daf97b77ee6dc4628c2859318fdc62) is pending review. The previous commits where reviewed [here](https://github.com/cloudfoundry/cli/pull/1976) and [here](https://github.com/cloudfoundry/cli/pull/1973) but cannot be merged until the command is finished.


resources/service_instance_resource.go
```diff
@@ -21,6 +21,8 @@ type ServiceInstance struct {
 	Name string `jsonry:"name,omitempty"`
 	// SpaceGUID is the space that this service instance relates to
 	SpaceGUID string `jsonry:"relationships.space.data.guid,omitempty"`
+	// PlanGUID is the plan this instance is created from
+	PlanGUID string `jsonry:"relationships.service_plan.data.guid,omitempty"`
```
blgm, line 25: Ideally this would be added to the resources unit tests

api/cloudcontroller/ccv3/job.go
```diff
@@ -114,6 +124,10 @@ func (client *Client) PollJob(jobURL JobURL) (Warnings, error) {
 			return allWarnings, nil
 		}
 
+		if job.IsAt(state) {
```
blgm, line 127: Could also be:
```go
if job.State == state {
```
My personal opinion is that having a method makes the solution more complex.

actor/v7action/cloud_controller_client.go
```diff
@@ -129,6 +129,7 @@ type CloudControllerClient interface {
 	MakeRequestSendReceiveRaw(Method string, URL string, headers http.Header, requestBody []byte) ([]byte, *http.Response, error)
 	MapRoute(routeGUID string, appGUID string) (ccv3.Warnings, error)
 	PollJob(jobURL ccv3.JobURL) (ccv3.Warnings, error)
+	PollJobForStatus(jobURL ccv3.JobURL, state constant.JobState) (ccv3.Warnings, error)
```
blgm, line 132: Given a job has a "state" and not a "status", it might be clearer to call this `PollForJobState()`, but it's quite common to use the words interchangeably so this is not a big one.

actor/actionerror/service_plan_not_found_error_test.go
```diff
@@ -16,7 +16,24 @@ var _ = Describe("ServicePlanNotFoundError", func() {
 			OfferingName: offeringName,
 			PlanName:     planName,
 		}
-		Expect(err.Error()).To(Equal(fmt.Sprintf("The plan %s could not be found for service %s.", planName, offeringName)))
+		Expect(err.Error()).To(Equal(fmt.Sprintf("The plan %s could not be found for service offering %s.", planName, offeringName)))
+	})
+
+	It("returns a message with the service offering name, the plan name and broker name", func() {
+		offeringName := "some-service-offering"
```
blgm, line 23: Or even better:
```go
const (
  offeringName = "some-service-offering"
...
```

actor/v7action/service_instance_test.go
```diff
@@ -898,4 +899,138 @@ var _ = Describe("Service Instance Actions", func() {
...
+			fakeCloudControllerClient.PollJobForStatusReturns(ccv3.Warnings{"job-warning"}, nil)
+		})
+
+		JustBeforeEach(func() {
+			warnings, err = actor.CreateManagedServiceInstance(fakeServiceOfferingName, fakeServicePlanName, fakeServiceInstanceName, fakeServiceBrokerName, fakeSpaceGUID)
+
+		})
+		It("gets the service plan", func() {
+			Expect(fakeCloudControllerClient.GetServicePlansCallCount()).To(Equal(1))
+			query := fakeCloudControllerClient.GetServicePlansArgsForCall(0)
```
blgm, line 934: This tests the query `Values` and not the `Key`.

actor/v7action/service_plan.go
```diff
@@ -22,9 +22,13 @@ func (actor Actor) GetServicePlanByNameOfferingAndBroker(servicePlanName, servic
 
 	switch len(servicePlans) {
 	case 0:
-		return resources.ServicePlan{}, Warnings(warnings), actionerror.ServicePlanNotFoundError{PlanName: servicePlanName}
+		return resources.ServicePlan{}, Warnings(warnings), actionerror.ServicePlanNotFoundError{
```
blgm, line 25: Nice improvement

command/v7/actor.go
```diff
@@ -38,6 +38,7 @@ type Actor interface {
 	CreateDockerPackageByApplication(appGUID string, dockerImageCredentials v7action.DockerImageCredentials) (v7action.Package, v7action.Warnings, error)
 	CreateDockerPackageByApplicationNameAndSpace(appName string, spaceGUID string, dockerImageCredentials v7action.DockerImageCredentials) (v7action.Package, v7action.Warnings, error)
 	CreateIsolationSegmentByName(isolationSegment v7action.IsolationSegment) (v7action.Warnings, error)
+	CreateManagedServiceInstance(serviceOfferingName, servicePlanName, serviceInstanceName, serviceBrokerName, spaceGUID string) (v7action.Warnings, error)
```
blgm, line 41: That's quite a complex signature. Some other resources have opted for passing an object such as `CreateApplicationInSpace` and `CreateIsolationSegmentByName`.

integration/v7/isolated/create_service_command_test.go
```diff
@@ -0,0 +1,180 @@
+package isolated
```
blgm, line 1: It would be nice to see a "happy case" test

resources/service_instance_resource.go
pivotal-marcela-campo: Didn't notice, I'll add it

actor/v7action/cloud_controller_client.go
pivotal-marcela-campo: That is a good point. Thanks for spotting it

command/v7/actor.go
pivotal-marcela-campo: Yes, it became pretty annoying specially now that I am adding also tags and parameters, so doing that refactor within the tags story.

api/cloudcontroller/ccv3/job.go
pivotal-marcela-campo: hmm, it is a fair point. I'd rather use a method though as the other conditions do it that way and I find it more readable when everything is at the same level of abstraction.

actor/v7action/service_instance_test.go
pivotal-marcela-campo: Indeed. I relaxed it because it is using another public method which has tests, so I am just testing that I passed in the right values. This is not mocked though, as they are all methods in the same actor struct, so I'll just add the missing bits

PR #1995, opened 2020-07-23
v8(services): service command displays upgrade info

blgm:
[#171843319](https://www.pivotaltracker.com/story/show/171843319)

### DIFF
--- a/actor/v7action/service_instance_details_test.go
+++ b/actor/v7action/service_instance_details_test.go
@@ -31,6 +31,7 @@ var _ = Describe("Service Instance Details Action", func() {
 			serviceInstanceGUID          = "some-service-instance-guid"
 			spaceGUID                    = "some-source-space-guid"
 			servicePlanName              = "fake-service-plan-name"
+			servicePlanGUID              = "fake-service-plan-guid"
 			serviceOfferingName          = "fake-service-offering-name"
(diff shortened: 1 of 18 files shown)


actor/v7action/service_instance_details_test.go
```diff
@@ -151,7 +152,11 @@ var _ = Describe("Service Instance Details Action", func() {
 				Expect(fakeCloudControllerClient.GetServiceInstanceParametersArgsForCall(0)).To(Equal(serviceInstanceGUID))
 			})
 
-			When("the service instance is managed", func() {
+			It("does not get the service plan", func() {
+				Expect(fakeCloudControllerClient.GetServicePlanByGUIDCallCount()).To(Equal(0))
+			})
```
butzopower, line 157: I know this is to guard to make sure we only call the service plan endpoint in situations where upgrade info will be available, but does it make sense to move this closer to perhaps: `"upgrade is not available and the service does not have maintenance info"`?  I'm just weary that we might refactor how / what method gets called on the service, and this sort of floating `It` will remain technically valid, but others might not know why it's there or that it relates to upgrades.

blgm: That makes sense. I'll update that in another PR.

PR #1996, opened 2020-07-24
v8(services): update managed service tags

blgm:
[#173416527](https://www.pivotaltracker.com/story/show/173416527)


actor/v7action/service_instance.go
```diff
@@ -50,6 +51,49 @@ func (actor Actor) UpdateUserProvidedServiceInstance(serviceInstanceName, spaceG
...
+		jobURL   ccv3.JobURL
+	)
+
+	warnings, err := railway.Sequentially(
+		func() (warnings ccv3.Warnings, err error) {
+			original, _, warnings, err = actor.CloudControllerClient.GetServiceInstanceByNameAndSpace(serviceInstanceName, spaceGUID)
+			return
+		},
+		func() (warnings ccv3.Warnings, err error) {
+			if original.Type != resources.ManagedServiceInstance {
```
kirederik, line 66: Is there a reason for this check to be in its own function rather than as part of the previous one?

actor/v7action/service_instance_test.go
```diff
@@ -368,6 +363,239 @@ var _ = Describe("Service Instance Actions", func() {
...
+			It("makes the right calls and returns all warnings", func() {
+				warnings, err := actor.UpdateManagedServiceInstance(
+					serviceInstanceName,
+					spaceGUID,
+					resources.ServiceInstance{Tags: types.NewOptionalStringSlice("foo", "bar")},
+				)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(warnings).To(ConsistOf("warning from get", "warning from update"))
+
+				Expect(fakeCloudControllerClient.GetServiceInstanceByNameAndSpaceCallCount()).To(Equal(1))
```
kirederik, line 400: I'm a big fan of `By` to divide tests with multiple related but different assertions like this one. I know it's a matter of preference, but it's something to consider 👀

actor/v7action/service_instance.go
blgm: It could be merged into the previous step, but I think having it here helps with separation of concerns. Also since you’d only want to do this if the previous step passed, it helps with consistency of error handling.

actor/v7action/service_instance_test.go
blgm: I’ll look into doing that for the next story

PR #1998, opened 2020-07-27
v8(services): service instance parameters can be updated

blgm:
[#173416526](https://www.pivotaltracker.com/story/show/173416526)

### DIFF
--- a/actor/v7action/service_instance.go
+++ b/actor/v7action/service_instance.go
@@ -10,6 +10,11 @@ import (
 	"code.cloudfoundry.org/cli/util/railway"
 )
 
+type ServiceInstanceUpdateManagedParams struct {
+	Tags       types.OptionalStringSlice
+	Parameters types.OptionalObject
+}
+
 func (actor Actor) GetServiceInstanceByNameAndSpace(serviceInstanceName string, spaceGUID string) (resources.ServiceInstance, Warnings, error) {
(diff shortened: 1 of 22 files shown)


actor/v7action/service_instance.go
```diff
@@ -87,13 +91,12 @@ func (actor Actor) UpdateManagedServiceInstance(serviceInstanceName, spaceGUID s
 
 	switch err.(type) {
 	case nil:
+		return Warnings(warnings), nil
```
kirederik, line 94: Theoretically this case is not needed (the default would cover it) but might be a good idea for make the behaviour explicit anyways

actor/v7action/service_instance.go
```diff
@@ -10,6 +10,11 @@ import (
 	"code.cloudfoundry.org/cli/util/railway"
 )
 
+type ServiceInstanceUpdateManagedParams struct {
+	Tags       types.OptionalStringSlice
+	Parameters types.OptionalObject
+}
+
```
kirederik, line 17: It wasn't immediate to me why we pulled this out of the `resources` package into the `v7action` package. Looking at the other commands, some implement something similar, some use the resource struct like the one we were using before. I'm inclined to believe it should as it was implemented before 🤔

blgm: It’s needed for the next story (service plans, because the command layer will specify a plan name, while the resource only has a GUID field). I agree that it’s not technically necessary here. At one point I thought that it was, and once I realised it was only needed for the next flag I decided to leave it in place.

PR #2013, opened 2020-08-04
v8(services): delete service instance using V3 API

blgm:
[#172205350](https://www.pivotaltracker.com/story/show/172205350)

### DIFF
--- a/actor/v7action/service_instance.go
+++ b/actor/v7action/service_instance.go
@@ -184,6 +184,58 @@ func (actor Actor) CreateManagedServiceInstance(params ManagedServiceInstancePar
 
 }
 
+type ServiceInstanceDeleteState int
+
+const (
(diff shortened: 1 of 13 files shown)


actor/v7action/service_instance.go
```diff
@@ -184,6 +184,58 @@ func (actor Actor) CreateManagedServiceInstance(params ManagedServiceInstancePar
...
+			switch {
+			case jobURL == "":
+				return
+			case wait:
+				return actor.CloudControllerClient.PollJob(jobURL)
+			default:
+				return actor.CloudControllerClient.PollJobForState(jobURL, constant.JobPolling)
+			}
+		},
+	)
```
FelisiaM, line 223: I like that `railway.Sequentially` offers the option to make the code more readable. I think if the functions are also extracted it would be even more readable.

PR #2014, opened 2020-08-04
v8(services): v3-update-service validations

pivotal-marcela-campo:
[#173416529](https://www.pivotaltracker.com/story/show/173416529)

A few notes:

1. Removed upgrade related flags as upgrade will be implemented as a separate command
2. At the actor, chose to not include the plan lookup as part of the railway function as it returns a different type of warnings and it is only executed if the `plan` was specified

### DIFF
--- a/actor/v7action/service_instance.go
+++ b/actor/v7action/service_instance.go
@@ -11,8 +11,9 @@ import (
 )
 
 type ServiceInstanceUpdateManagedParams struct {
-	Tags       types.OptionalStringSlice
-	Parameters types.OptionalObject
+	ServicePlanName types.OptionalString
+	Tags            types.OptionalStringSlice
+	Parameters      types.OptionalObject
 }
 
 func (actor Actor) GetServiceInstanceByNameAndSpace(serviceInstanceName string, spaceGUID string) (resources.ServiceInstance, Warnings, error) {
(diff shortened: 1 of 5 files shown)


actor/v7action/service_instance.go
```diff
@@ -59,16 +60,29 @@ func (actor Actor) UpdateUserProvidedServiceInstance(serviceInstanceName, spaceG
...
 func (actor Actor) UpdateManagedServiceInstance(serviceInstanceName, spaceGUID string, serviceInstanceUpdates ServiceInstanceUpdateManagedParams) (Warnings, error) {
 	var (
-		original resources.ServiceInstance
-		jobURL   ccv3.JobURL
+		original    resources.ServiceInstance
+		jobURL      ccv3.JobURL
+		allWarnings Warnings
 	)
 
+	if serviceInstanceUpdates.ServicePlanName.IsSet {
```
blgm, line 68: I think there are a couple of ways this could be improved:
1) Ideally we should only query plans from the same service offering as the existing service plan
2) This could be incorporated as a step in the "railway" structure below

As there are follow-on stories, I'm happy to defer improvements to them

command/v7/update_service_command.go
```diff
@@ -42,17 +31,18 @@ func (cmd UpdateServiceCommand) Execute(args []string) error {
 		return err
 	}
 
-	_, err := cmd.processParameters()
-	if err != nil {
-		return err
+	if cmd.noFlagsProvided() {
```
blgm, line 34: This abstraction makes the intent of the code very clear

Issue #1794, opened 2019-09-16
"cf create-service" does not handle whitespace consistently


kimthomasrehmann:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [X] I reviewed open and closed github issues that may be related to my problem.
- [X] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [X] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [X] I am reporting a bug that others will be able to reproduce.
- [ ] If this is an issue for the v7 beta release, I've read through the [official docs](https://docs.cloudfoundry.org/cf-cli/v7.html) and the [release notes](https://github.com/cloudfoundry/cli/releases).

**Describe the bug and the command you saw an issue with**

Trying to create a service with a trailing whitespace character reports a failure, however a service is created with a name that strips the whitespace

**What happened**

A service with the incorrect name was created

**Expected behavior**

A service with the correct name is created, or no service is created and an error message is displayed

**To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:

#cf cs service_name service_plan 'Kim Whitespace Test '
Creating service instance Kim Whitespace Test  in org  xxx / space yyy as automation_user...
FAILED
Service instance Kim Whitespace Test  not found
#cf service --guid 'Kim Whitespace Test '
Service instance Kim Whitespace Test  not found
FAILED
#cf service --guid 'Kim Whitespace Test'  # without trailing whitespace
xxxxxxxxx-yyyy-yyyy-yyyy-zzzzzzzz # a valid service GUID is displayed

**Provide more context**

SLES15 with CF CLI 6.37.0

- platform and shell details ( e.g. Mac OS X 10.11 iTerm)
- version of the CLI you are running
- version of the CC API Release you are on

Note: As of January 2019, we no longer support API versions older than CF Release v284/CF Deployment v1.7.0 (CAPI Release: 1.46.0 (APIs 2.100.0 and 3.35.0).

Note: In order to complete the v7 beta cf CLI in a timely matter, we develop and test against the latest CAPI release candidate. When v7 cf CLI is generally available, we will start supporting official CC API releases again.

abbyachau:
Hey @waterlink wondering if you have any thoughts on this as you build v7 services related commands.

waterlink:
@abbyachau @kimthomasrehmann 

[...]

PR #2019, opened 2020-08-11
v8(services): consistently handle whitespaces

kirederik:
Solves #1794

[#168509681](https://www.pivotaltracker.com/story/show/168509681)

### DIFF
--- a/command/flag/arguments.go
+++ b/command/flag/arguments.go
@@ -1,5 +1,7 @@
 package flag
 
+import "code.cloudfoundry.org/cli/types"
+
 type AppName struct {
 	AppName string `positional-arg-name:"APP_NAME" required:"true" description:"The application name"`
 }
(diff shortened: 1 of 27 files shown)


kirederik:
> A couple of thoughts:
> 
> * this fix is in the CLI, did you consider fixes in the CC?

The API ignores whitespaces when creating/updating, but doesn't handle them correctly when querying. Might be good to fix that on the API as well 🤔 

> * I would slightly prefer an approach that involves implementing a custom unmarshal like this: https://github.com/cloudfoundry/cli/blob/0d68f88697b720575fa0bad866b24d2f9354c0aa/command/flag/credentials_or_json.go#L18
> 
> ```go
> func (s *ServiceInstance) UnmarshalFlag(input string) error {
>   *s = strings.TrimSpace(input)
>   return nil
> }
> ```

Nice. I'll take a look at that.

PR #2025, opened 2020-08-19
v8(services): share-service initial setup

FelisiaM:
- command signature
- check space is targeted

[#173658279](https://www.pivotaltracker.com/story/show/173658279)

### DIFF
--- a/command/v7/share_service_command_test.go
+++ b/command/v7/share_service_command_test.go
@@ -1,0 +1,63 @@
+package v7_test
+
+import (
+	"errors"
+
+	"code.cloudfoundry.org/cli/command/commandfakes"
+	. "code.cloudfoundry.org/cli/command/v7"
+	"code.cloudfoundry.org/cli/command/v7/v7fakes"
+	"code.cloudfoundry.org/cli/util/ui"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+)
+
+var _ = Describe("share-service Command", func() {
+	var (
+		cmd             ShareServiceCommand
+		testUI          *ui.UI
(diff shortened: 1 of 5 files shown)


command/v7/share_service_command_test.go
```diff
@@ -0,0 +1,63 @@
...
+	JustBeforeEach(func() {
+		executeErr = cmd.Execute(nil)
+	})
+
+	Context("user not targeting space", func() {
+		BeforeEach(func() {
+			fakeSharedActor.CheckTargetReturns(errors.New("space not targeted"))
+		})
+
+		It("checks the user is logged in, and targeting an org and space", func() {
```
blgm, line 50: I don't think this test should be in this context, because the checks happen regardless of the result.

integration/v7/isolated/share_service_command_test.go
```diff
@@ -0,0 +1,77 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("share-service command", func() {
+	const shareServiceCommand = "v3-share-service"
+
+	FDescribe("help", func() {
```
blgm, line 15: Focused test. `ginkgo blur` will fix it.

blgm (changes requested):

integration/v7/isolated/share_service_command_test.go
FelisiaM: I had made the change but forgot to commit it 👍

command/v7/share_service_command_test.go
FelisiaM: yep makes sense - did the change

PR #2027, opened 2020-08-24
Validation of orguments for share-service

FelisiaM:
[#173658278](https://www.pivotaltracker.com/story/show/173658278)

### DIFF
--- a/actor/v7action/service_instance_sharing.go
+++ b/actor/v7action/service_instance_sharing.go
@@ -1,0 +1,49 @@
+package v7action
+
+import (
+	"code.cloudfoundry.org/cli/actor/actionerror"
+	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3"
+	"code.cloudfoundry.org/cli/resources"
+	"code.cloudfoundry.org/cli/types"
+	"code.cloudfoundry.org/cli/util/railway"
+)
+
+type ServiceInstanceSharingParams struct {
+	SpaceName string
+	OrgName   types.OptionalString
+}
+
+func (actor Actor) ShareServiceInstanceToSpaceAndOrg(
+	serviceInstanceName, targetedSpaceGUID, targetedOrgGUID string,
+	sharedToDetails ServiceInstanceSharingParams,
+) (Warnings, error) {
+
+	return handleServiceInstanceErrors(railway.Sequentially(
+		func() (warnings ccv3.Warnings, err error) {
+			_, _, warnings, err = actor.CloudControllerClient.GetServiceInstanceByNameAndSpace(serviceInstanceName, targetedSpaceGUID)
+			return
+		},
+		func() (warnings ccv3.Warnings, err error) {
+			if sharedToDetails.OrgName.IsSet {
+				orgName := sharedToDetails.OrgName.Value
+
+				var organizations []resources.Organization
+				organizations, warnings, err = actor.CloudControllerClient.GetOrganizations(
+					[]ccv3.Query{ccv3.Query{Key: ccv3.NameFilter, Values: []string{orgName}}}...,
+				)
+				if err == nil && len(organizations) == 0 {
+					err = actionerror.OrganizationNotFoundError{Name: orgName}
+				}
+			}
+			return
+		},
+		func() (warnings ccv3.Warnings, err error) {
+			var spaceWarnings Warnings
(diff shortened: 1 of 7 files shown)


command/v7/share_service_command_test.go
```diff
@@ -60,4 +68,67 @@ var _ = Describe("share-service Command", func() {
 		})
 	})
 
+	Context("user is logged in", func() {
```
kirederik, line 71: Should this be "There's a space and org targeted" or something like that?

command/v7/share_service_command_test.go
```diff
@@ -60,4 +68,67 @@ var _ = Describe("share-service Command", func() {
...
+			It("calls the actor to share in specified space and org", func() {
+				Expect(fakeActor.ShareServiceInstanceToSpaceAndOrgCallCount()).To(Equal(1))
+
+				actualServiceInstance, actualTargetedSpace, actualTargetedOrg, actualSharingParams := fakeActor.ShareServiceInstanceToSpaceAndOrgArgsForCall(0)
+				Expect(actualServiceInstance).To(Equal(expectedServiceInstanceName))
+				Expect(actualTargetedSpace).To(Equal(expectedTargetedSpaceGuid))
+				Expect(actualTargetedOrg).To(Equal(expectedTargetedOrgGuid))
+				Expect(actualSharingParams).To(Equal(v7action.ServiceInstanceSharingParams{
+					SpaceName: expectedSpaceName,
+					OrgName:   types.NewOptionalString(expectedSpecifiedOrgName),
```
kirederik, line 117: This is the only difference between this `It` and the previous one. I'd delete all the other expectations have just that (personal preference; feel free to ignore)

actor/v7action/service_instance_sharing_test.go
```diff
@@ -0,0 +1,171 @@
...
+									Values: []string{shareToSpaceName},
+								},
+								{
+									Key:    ccv3.OrganizationGUIDFilter,
+									Values: []string{shareToOrgGUID},
+								},
+							}))
+					})
+
+					It("returns an actor error and warnings", func() {
```
kirederik, line 136: I missed a "when the space cannot be found in the specified org". Reading this spec, we end up with:

When the org to share to is also specified
  It searches for the space in the specified org
  It returns an actor error and warnings

Not a big deal though

kirederik (approved):

actor/v7action/service_instance_sharing.go
```diff
@@ -0,0 +1,50 @@
...
+		func() (warnings ccv3.Warnings, err error) {
+			if sharedToDetails.OrgName.IsSet {
+				var (
+					orgWarnings  Warnings
+					organization resources.Organization
+				)
+
+				organization, orgWarnings, err = actor.GetOrganizationByName(sharedToDetails.OrgName.Value)
+				warnings = ccv3.Warnings(orgWarnings)
+				if err == nil {
```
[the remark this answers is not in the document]
FelisiaM: Not sure I understand the part that is not needed. 
If it is the err `GetOrganizationByName` returns an error when the org list is empty

command/v7/share_service_command_test.go
FelisiaM: done

command/v7/share_service_command_test.go
FelisiaM: updated the naming ✅

actor/v7action/service_instance_sharing_test.go
FelisiaM: My idea was 
- space cannot be found
  - it errors in default case (targeted org)
  - when org is also specified
   - It searches for the space in the specified org
   - It returns an actor error and warnings


Might be worth revisiting this after all tests have been written to see the flow of al the test cases

actor/v7action/service_instance_sharing.go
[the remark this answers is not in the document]
kirederik: The `if err == nil` is the bit I pointed out as unnecessary. Even when `GetOrganizationByName` errors, the organization will be initialised (with empty, yes, but it will nonetheless).

If we expand the code we have here with the railways code, we end up with something like:

```
organization, orgWarnings, err = actor.GetOrganizationByName(sharedToDetails.OrgName.Value)
if err == nil {
  // do something
}
if err != nil {
  return err, warning
}
```

PR #2029, opened 2020-08-25
v8(services): v3-unshare-service args, flags and help

pivotal-marcela-campo:
[#173659464](https://www.pivotaltracker.com/story/show/173659464)

### DIFF
--- a/command/v7/unshare_service_command.go
+++ b/command/v7/unshare_service_command.go
@@ -1,0 +1,24 @@
+package v7
+
+import "code.cloudfoundry.org/cli/command/flag"
+
+type UnshareServiceCommand struct {
+	BaseCommand
+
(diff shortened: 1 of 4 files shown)


command/v7/unshare_service_command.go
```diff
@@ -0,0 +1,24 @@
+package v7
+
+import "code.cloudfoundry.org/cli/command/flag"
+
+type UnshareServiceCommand struct {
+	BaseCommand
+
+	RequiredArgs    flag.ShareServiceArgs `positional-args:"yes"`
```
kirederik, line 8: Not sure if that's a pattern with other commands but, if it is, just ignore this comment.

I see why reusing the arguments from Share makes sense (they are indeed the same) but this has the side effect of, if we ever add more flags to one command, it will inadvertently impacts the other one.

kirederik (approved):

PR #2040, opened 2020-09-02
Unshare service instance 

FelisiaM:
This PR makes sure that the unshare request is sent to the API

It does not specifically handle the case when a relationship between a space and a SI did not previously exist. That will require API changes and will be done as a separate task. 

[#173659461](https://www.pivotaltracker.com/story/show/173659461)


actor/v7action/service_instance_sharing_test.go
```diff
@@ -248,13 +249,14 @@ var _ = Describe("Service Instance Sharing", func() {
 			})
 
 			It("returns an error and warnings", func() {
+				Expect(executionError).ToNot(BeNil())
```
blgm, line 252: The idiomatic way to do this in Gomega is:
```go
Expect(executionError).To(MatchError("cannot share the instance"))
```

actor/v7action/service_instance_sharing_test.go
```diff
@@ -248,13 +249,14 @@ var _ = Describe("Service Instance Sharing", func() {
...
 				Expect(executionError.Error()).To(Equal("cannot share the instance"))
 				Expect(warnings).To(ConsistOf("some-space-warning", "some-service-instance-warning", "some-share-warning"))
 			})
 		})
 	})
 
-	Describe("UnShareServiceInstanceToSpaceAndOrg", func() {
+	Describe("UnshareServiceInstanceToSpaceAndOrg", func() {
```
blgm, line 259: Should it be UnshareServiceInstanceFromSpaceAndOrg to match the method name?

actor/v7action/service_instance_sharing_test.go
```diff
@@ -273,5 +275,68 @@ var _ = Describe("Service Instance Sharing", func() {
...
+
+				fakeCloudControllerClient.GetServiceInstanceByNameAndSpaceReturns(
+					resources.ServiceInstance{GUID: expectedServiceInstanceGUID},
+					ccv3.IncludedResources{},
+					ccv3.Warnings{"some-service-instance-warning"},
+					nil,
+				)
+			})
+
+			It("makes a request to the cloud controller", func() {
```
blgm, line 298: This is ok, but a way that I prefer to see tests is:
```go
It("makes the right request", ...)

When("the request is successful", func() {
  It("returns warnings and no error", ...)
})

When("the request fails", func() {
  It("returns warnings and an error", ...)
})
```

This is because the first test happens regardless of the response.

actor/v7action/service_instance_sharing_test.go
```diff
@@ -273,5 +275,68 @@ var _ = Describe("Service Instance Sharing", func() {
...
+				Expect(fakeCloudControllerClient.GetServiceInstanceByNameAndSpaceCallCount()).To(Equal(1))
+				Expect(fakeCloudControllerClient.GetSpacesCallCount()).To(Equal(1))
+				Expect(fakeCloudControllerClient.UnshareServiceInstanceFromSpaceCallCount()).To(Equal(1))
+
+				actualServiceInstanceGUID, actualSpace := fakeCloudControllerClient.UnshareServiceInstanceFromSpaceArgsForCall(0)
+				Expect(actualServiceInstanceGUID).To(Equal(expectedServiceInstanceGUID))
+				Expect(actualSpace).To(Equal(expectedSpaceGUID))
+
+				Expect(executionError).To(BeNil())
```
blgm, line 307: Idiomatic Gomega is
```go
Expect(executionError).NotTo(HaveOccurred())
```

api/cloudcontroller/ccv3/service_instance_test.go
```diff
@@ -645,6 +645,85 @@ var _ = Describe("Service Instance", func() {
...
+				spaceGUID = "some-space-guid"
+			})
+
+			JustBeforeEach(func() {
+				warnings, executeErr = client.UnshareServiceInstanceFromSpace(serviceInstanceGUID, spaceGUID)
+			})
+
+			When("no errors occur deleting the shared space relationship", func() {
+				BeforeEach(func() {
+					server.AppendHandlers(
```
blgm, line 668: Another way to test this is to use the `FakeRequester` like this: https://github.com/cloudfoundry/cli/blob/167262b87a4c9dc5341dcfb049dba027748d25d6/api/cloudcontroller/ccv3/service_instance_test.go#L504

blgm (approved):

PR #2054, opened 2020-09-23
v8(services): v3-service cmd to show bound apps

blgm:
Note: the current CAPI version sometimes returns `null` for the last
opertation of a binding, which would cause this code to fail. There is work in the
pipeline to resolve this, and for now this is being delivered without
integration tests, so that we do not break CI.

[#171843320](https://www.pivotaltracker.com/story/show/171843320)

### DIFF
--- a/actor/v7action/service_instance_details.go
+++ b/actor/v7action/service_instance_details.go
@@ -46,9 +46,10 @@ type ServiceInstanceDetails struct {
 	Parameters        ServiceInstanceParameters
 	SharedStatus      SharedStatus
 	UpgradeStatus     ServiceInstanceUpgradeStatus
+	BoundApps         []resources.ServiceCredentialBinding
 }
 
-func (actor Actor) GetServiceInstanceDetails(serviceInstanceName string, spaceGUID string) (ServiceInstanceDetails, Warnings, error) {
+func (actor Actor) GetServiceInstanceDetails(serviceInstanceName string, spaceGUID string, omitApps bool) (ServiceInstanceDetails, Warnings, error) {
 	var serviceInstanceDetails ServiceInstanceDetails
 
 	warnings, err := railway.Sequentially(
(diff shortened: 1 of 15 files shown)


actor/v7action/service_instance_details.go
```diff
@@ -215,6 +220,19 @@ func (actor Actor) getServiceInstanceUpgradeStatus(serviceInstanceDetails Servic
 	}
 }
 
+func (actor Actor) getServiceInstanceBoundApps(serviceInstanceGUID string, omitApps bool) (boundApps []resources.ServiceCredentialBinding, warnings ccv3.Warnings, err error) {
+	switch omitApps {
```
pivotal-marcela-campo, line 224: I find this a bit weird. I would find out if we need the apps before calling this method at all. It would made the intent clearer. From reading the caller code I thought maybe we were just fetching ids and the flag was to include the whole resource, but it turns out it is meant to skip the method altogether.

api/cloudcontroller/ccv3/service_credential_binding_test.go
```diff
@@ -26,45 +26,50 @@ var _ = Describe("Service Instance", func() {
...
-				types := map[bool]resources.ServiceCredentialBindingType{true: resources.KeyBinding, false: resources.AppBinding}
+				types := []resources.ServiceCredentialBindingType{resources.KeyBinding, resources.AppBinding}
 				for i := 1; i <= 3; i++ {
 					Expect(requestParams.AppendToList(resources.ServiceCredentialBinding{
 						GUID:                fmt.Sprintf("credential-binding-%d-guid", i),
 						Name:                fmt.Sprintf("credential-binding-%d-name", i),
 						ServiceInstanceGUID: fmt.Sprintf("si-%d-guid", i),
 						AppGUID:             fmt.Sprintf("app-%d-guid", i),
-						Type:                types[typeSwitch],
+						Type:                types[i%2],
```
pivotal-marcela-campo, line 45: this looks nicer :)

pivotal-marcela-campo (changes requested):

PR #2105, opened 2020-11-05
v8(services): unbind-service cmd use V3 API 

blgm:
[#175356881](https://www.pivotaltracker.com/story/show/175356881)
[#175356884](https://www.pivotaltracker.com/story/show/175356884)


actor/v7action/service_app_binding.go
```diff
@@ -54,6 +60,48 @@ func (actor Actor) CreateServiceAppBinding(params CreateServiceAppBindingParams)
...
+		app             resources.Application
+		binding         resources.ServiceCredentialBinding
+		jobURL          ccv3.JobURL
+		stream          chan PollJobEvent
+	)
+
+	warnings, err := railway.Sequentially(
+		func() (warnings ccv3.Warnings, err error) {
+			serviceInstance, _, warnings, err = actor.getServiceInstanceByNameAndSpace(params.ServiceInstanceName, params.SpaceGUID)
+			return
```
kirederik, line 75: Just out of curiosity, are those `return` statements needed? 🤔

blgm: Yes, you always need a `return` in Go if the function has a return value. With named return values you can omit the variable names, but you still need the `return` statement.

command/v7/unbind_service_command.go
```diff
@@ -0,0 +1,88 @@
...
+	case actionerror.ServiceBindingNotFoundError:
+		cmd.UI.DisplayText("Binding between {{.ServiceInstanceName}} and {{.AppName}} did not exist", cmd.names())
+		cmd.UI.DisplayOK()
+		return nil
+	default:
+		return err
+	}
+
+	completed, err := shared.WaitForResult(stream, cmd.UI, cmd.Wait)
+	switch {
```
kirederik, line 44: I'm particularly not a big fan of this type of `switch`. I'm just used to believe that all cases are assertions on the same "variable" (if that makes sense). Not advocating to change it though, since it seems to be [idiomatic](https://golang.org/doc/effective_go.html#switch) to use `switch` this way 🤔

kirederik (approved):

actor/v7action/service_app_binding.go
kirederik: Yeah It just passed me that the function signature has return values 🤦

PR #2129, opened 2021-01-21
v8(services): delete-service-key use V3 API

blgm:
[#175573217](https://www.pivotaltracker.com/story/show/175573217)


FelisiaM (approved):
Looks good!

PR #2130, opened 2021-01-22
v8(services): create-service-key use V3 API

blgm:
[#175182328](https://www.pivotaltracker.com/story/show/175182328)
[#175185924](https://www.pivotaltracker.com/story/show/175185924)


command/v7/create_service_key_command.go
```diff
@@ -0,0 +1,108 @@
...
+	default:
+		cmd.UI.DisplayOK()
+		cmd.UI.DisplayText("Create in progress.")
+		return nil
+	}
+}
+
+func (cmd CreateServiceKeyCommand) Usage() string {
+	return `
+CF_NAME create-service-key SERVICE_INSTANCE SERVICE_KEY [-c PARAMETERS_AS_JSON] [--wait]
```
pivotal-marcela-campo, line 61: We should change PARAMETERS_AS_JSON to JSON_PARAMETERS as per notes in story

blgm: This triggered an investigation into what all the commands do to ensure consistency. Almost all commands use `PARAMETERS_AS_JSON` so we decided to standardise on that in this commit: https://github.com/cloudfoundry/cli/commit/146fa478c2aec6cdf93ab8833959d8cac3127b38

PR #2141, opened 2021-02-16
service: --params flag retrieves service instance params

FelisiaM:
[#173917704](https://www.pivotaltracker.com/story/show/173917704)

### DIFF
--- a/actor/v7action/service_instance_details.go
+++ b/actor/v7action/service_instance_details.go
@@ -160,15 +160,19 @@ func (actor Actor) getServiceInstanceDetails(serviceInstanceName string, spaceGU
 
 func (actor Actor) getServiceInstanceParameters(serviceInstanceGUID string) (ServiceInstanceParameters, ccv3.Warnings) {
 	params, warnings, err := actor.CloudControllerClient.GetServiceInstanceParameters(serviceInstanceGUID)
-	if err != nil {
-		if e, ok := err.(ccerror.V3UnexpectedResponseError); ok && len(e.Errors) > 0 {
-			return ServiceInstanceParameters{MissingReason: e.Errors[0].Detail}, warnings
-		} else {
-			return ServiceInstanceParameters{MissingReason: err.Error()}, warnings
-		}
-	}
 
-	return ServiceInstanceParameters{Value: params}, warnings
+	switch err := err.(type) {
+	case nil:
+		return ServiceInstanceParameters{Value: params}, warnings
+	case ccerror.ResourceNotFoundError: // Old API versions did not produce a helpful error
+		return ServiceInstanceParameters{MissingReason: "This service does not support fetching service instance parameters."}, warnings
+	case ccerror.ServiceInstanceParametersFetchNotSupportedError:
+		return ServiceInstanceParameters{MissingReason: err.Message}, warnings
+	case ccerror.V3UnexpectedResponseError:
+		return ServiceInstanceParameters{MissingReason: err.Errors[0].Detail}, warnings
+	default:
+		return ServiceInstanceParameters{MissingReason: err.Error()}, warnings
+	}
 }
 
 func (actor Actor) getServiceInstanceSharedStatus(serviceInstanceDetails ServiceInstanceDetails, targetedSpace string) (SharedStatus, ccv3.Warnings, error) {
--- a/actor/v7action/service_instance_details.go
+++ b/actor/v7action/service_instance_details.go
@@ -33,8 +33,7 @@ type SharedStatus struct {
 }
 
 type ServiceInstanceParameters struct {
-	Value         types.JSONObject
-	MissingReason string
+	Value types.JSONObject
 }
 
 type ServiceInstanceUpgradeState int
@@ -102,7 +101,7 @@ func (actor Actor) GetServiceInstanceParameters(serviceInstanceName string, spac
 			return
 		},
 		func() (warnings ccv3.Warnings, err error) {
-			serviceInstanceDetails.Parameters, warnings = actor.getServiceInstanceParameters(serviceInstanceDetails.ServiceInstance.GUID)
+			serviceInstanceDetails.Parameters, warnings, err = actor.getServiceInstanceParameters(serviceInstanceDetails.ServiceInstance.GUID)
 			return
 		},
 	)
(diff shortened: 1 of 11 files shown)


actor/v7action/service_instance_details.go
```diff
@@ -97,6 +93,26 @@ func (actor Actor) GetServiceInstanceDetails(serviceInstanceName string, spaceGU
 	return serviceInstanceDetails, Warnings(warnings), nil
 }
 
+func (actor Actor) GetServiceInstanceParameters(serviceInstanceName string, spaceGUID string) (ServiceInstanceDetails, Warnings, error) {
```
blgm, line 96: It feels like a `GetServiceInstanceParameters()` function should return `ServiceInstanceParameters` rather than `ServiceInstanceDetails`

actor/v7action/service_instance_details.go
```diff
@@ -144,15 +160,19 @@ func (actor Actor) getServiceInstanceDetails(serviceInstanceName string, spaceGU
...
-	if err != nil {
-		if e, ok := err.(ccerror.V3UnexpectedResponseError); ok && len(e.Errors) > 0 {
-			return ServiceInstanceParameters{MissingReason: e.Errors[0].Detail}, warnings
-		} else {
-			return ServiceInstanceParameters{MissingReason: err.Error()}, warnings
-		}
-	}
 
-	return ServiceInstanceParameters{Value: params}, warnings
+	switch err := err.(type) {
```
blgm, line 164: As discussed it would be better to return errors rather than expose a MissingReason, as this pattern came from an old implementation that was trying to do something else.

actor/v7action/service_instance_details_test.go
```diff
@@ -764,4 +708,184 @@ var _ = Describe("Service Instance Details Action", func() {
...
+			})
+		})
+
+		Describe("getting the parameters", func() {
+			It("makes the correct call to get the parameters", func() {
+				Expect(fakeCloudControllerClient.GetServiceInstanceParametersCallCount()).To(Equal(1))
+				Expect(fakeCloudControllerClient.GetServiceInstanceParametersArgsForCall(0)).To(Equal(serviceInstanceGUID))
+			})
+
+			When("getting the parameters fails with a V3UnexpectedResponseError", func() {
```
blgm, line 814: I don't think we should have a specific code path for a `V3UnexpectedResponseError`

api/cloudcontroller/ccv3/errors_test.go
```diff
@@ -107,6 +107,26 @@ var _ = Describe("Error Wrapper", func() {
...
+      {
+         "detail": "This service does not support fetching service instance parameters.",
+         "title": "CF-ServiceFetchInstanceParametersNotSupported",
+         "code": 120004
+      }
+   ]
+}`
+						})
+
+						It("returns a BadRequestError", func() {
```
blgm, line 125: Should this say ServiceInstanceParametersFetchNotSupportedError?

command/v7/service_command.go
```diff
@@ -32,7 +33,12 @@ func (cmd ServiceCommand) Execute(args []string) error {
 		err      error
 	)
 
-	cmd.serviceInstance, warnings, err = cmd.Actor.GetServiceInstanceDetails(string(cmd.RequiredArgs.ServiceInstance), cmd.Config.TargetedSpace().GUID, false)
+	switch {
```
blgm, line 36: I don't like having two `switch` statements here - it feels like overlapping code paths. I prefer:
```go
switch {
case cmd.ShowGUID:
  return cmd.fetchAndDisplayGUID()
case cmd.Params:
  return cmd.fetchAndDisplayParams()
default:
  return cmd.fetchAndDisplayDetails()
}
```
So the code path split is explicit.

command/v7/service_command.go
```diff
@@ -177,15 +184,7 @@ func (cmd ServiceCommand) displayParametersEmpty() {
...
-		map[string]interface{}{
-			"ServiceInstanceName": cmd.serviceInstance.Name,
-		},
-	)
-	cmd.UI.DisplayNewline()
-
-	data, err := json.Marshal(cmd.serviceInstance.Parameters.Value)
+	data, err := json.MarshalIndent(cmd.serviceInstance.Parameters.Value, "", "  ")
 	if err != nil {
 		panic(err)
```
blgm, line 189: I think I wrote this originally. What do you think about it? The reasoning is that since the parameters were once JSON, it's impossible for the Marshal to return an error - since errors would be because of types that cannot be marshalled. However looking at it again, it looks like a code small. I wonder if there's a better way to encapsulate it? Or just return the error so that the code looks "normal" even though there will never be an error?

command/v7/service_command_test.go
```diff
@@ -738,6 +653,101 @@ var _ = Describe("service command", func() {
 		})
 	})
 
+	When("the --params flag is specified", func() {
+		const (
+			parameters = `{"foo":"bar"}`
```
blgm, line 658: Is this used?

integration/v7/isolated/service_command_test.go
```diff
@@ -548,5 +515,44 @@ var _ = Describe("service command", func() {
...
+						broker.FirstServiceOfferingName(),
+						broker.FirstServicePlanName(),
+						serviceInstanceName,
+						"-c", fmt.Sprintf(`{"%s":"%s"}`, key, value),
+					}
+					Eventually(helpers.CF(command...)).Should(Exit(0))
+					Eventually(helpers.CF(serviceCommand, serviceInstanceName)).Should(Say(`status:\s+create succeeded`))
+				})
+
+				FIt("reports the service instance parameters JSON", func() {
```
blgm, line 545: Might need to run `ginkgo blur`

blgm (changes requested):

actor/v7action/service_instance_details.go
```diff
@@ -142,17 +157,19 @@ func (actor Actor) getServiceInstanceDetails(serviceInstanceName string, spaceGU
...
+	case nil:
+	case ccerror.ResourceNotFoundError,
+		ccerror.ServiceInstanceParametersFetchNotSupportedError:
+		return ServiceInstanceParameters{}, warnings, actionerror.ServiceInstanceParamsFetchingNotSupportedError{}
+	default:
+		return ServiceInstanceParameters{}, warnings, err
 	}
 
-	return ServiceInstanceParameters{Value: params}, warnings
+	return ServiceInstanceParameters{Value: params}, warnings, nil
```
blgm, line 172: Given that this is just one line, I'd move it into the switch statement, and then there's no need for a return at the end of the function.

actor/v7action/service_instance_details.go
```diff
@@ -33,8 +33,7 @@ type SharedStatus struct {
 }
 
 type ServiceInstanceParameters struct {
```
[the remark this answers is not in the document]
blgm: Or just use the `types.JSONObject` type...

blgm (changes requested):

command/v7/service_command.go
FelisiaM: Hmm - What about just ignoring the error. We are confident that it is JSON and there are tests around the result.

blgm (approved):

PR #2148, opened 2021-02-22
Fix stack being over-ridden everytime in v7 push

tstannard:
v7

## Description of the Change

When v7 push is called with a manifest that has a stack field set, the cf CLI would override that field to an empty string. This PR makes it so that we only override the stack field in the manifest when a user provides a stack with the -s flag. 

## Why Is This PR Valuable?

When users set the stack field in their manifest, they would like to use it. 

Bug fix.

## Applicable Issues

https://github.com/cloudfoundry/cloud_controller_ng/issues/1819

## How Urgent Is The Change?

We would like this shipped in the next minor v7 release 

## Other Relevant Parties

@amhuber who reported this issue

### DIFF
--- a/actor/v7pushaction/handle_stack_override.go
+++ b/actor/v7pushaction/handle_stack_override.go
@@ -10,8 +10,8 @@ func HandleStackOverride(manifest manifestparser.Manifest, overrides FlagOverrid
 		if manifest.ContainsMultipleApps() {
 			return manifest, translatableerror.CommandLineArgsWithMultipleAppsError{}
 		}
+		manifest.Applications[0].Stack = overrides.Stack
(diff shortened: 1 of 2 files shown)


a-b (commented):
LGTM

a-b:
@thausler786 can you please review the proposed change from the product perspective?

thausler786:
This is a change in behavior we will have to call out in the release notes.   

Reading the command I believe the user would clearly expect that the CLI would use the stack in the manifest as given, so this is a straightforward bug fix that should be included, even if it does cause a change in behavior.

a-b:
cc @reedr3

PR #2154, opened 2021-03-11
Use`UI.DisplayJSON` for all services JSON output

FelisiaM:
Service instance param and service key details now use UI.DisplayJSON - which does not escape special chars

[#171951689](https://www.pivotaltracker.com/story/show/171951689)

### DIFF
--- a/actor/v7action/service_instance_details.go
+++ b/actor/v7action/service_instance_details.go
@@ -5,7 +5,6 @@ import (
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccerror"
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3"
 	"code.cloudfoundry.org/cli/resources"
-	"code.cloudfoundry.org/cli/types"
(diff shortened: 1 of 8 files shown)


actor/v7action/service_instance_details.go
```diff
@@ -32,7 +31,7 @@ type SharedStatus struct {
 	UsageSummary              []UsageSummaryWithSpaceAndOrg
 }
 
-type ServiceInstanceParameters types.JSONObject
```
blgm, line 35: The comment for this type says:
```go
// JSONObject represents a JSON object. When not initialized it serializes to `{}`,
// whereas a `map[string]interface{}` serializes to `null`.
```
So this is a change of behaviour. Is that intended?

FelisiaM: I think in this all the usages we initialise it. So it shouldn't be an issue. 
Also, the serialization was happening via json.MarshalIndent, which was the one causing the issue.

blgm (approved):

Issue #2069, opened 2020-10-07
CF7 REGRESSION terabytes are not supported for memory quota


mike1808:
**Describe the bug and the command you saw an issue with**
Set quota to 1 terabyte using `1T`.

**What happened**
CF CLI failed with `Incorrect Usage: Byte quantity must be an integer with a unit of measurement like M, MB, G, or GB`.

**Expected behavior**
The quota is set to 1 terabyte. This was possible in CF6.

**Exact Steps To Reproduce**
`cf update-quota -m 1T`

**Provide more context**
- `cf version 7.1.0+4c3168f9a.2020-09-09`

heyjcollins:
Thanks for bringing this to our attention @mike1808. 
This regression was not intentional, however given our overall priorities and the fact that a quota equivalent to a terabyte can be set by passing in a large `G` value, I'm not inclined to prioritize fixing this and I'm going to respectfully close this issue.

However... If you can help me understand the use-case(s) that **require** `T` support, I could be convinced otherwise..
Please feel free to re-open and comment if you'd like to continue discussing.

DanielJonesEB:
Y'all can't leave this on a cliff-hanger! I'm fascinated as to the use case for 1TiB apps. I know the JVM can be a little bloated, but...

PR #2167, opened 2021-05-17
Allow for configuration of quotas in terabytes

crsimmons:

v7

## Description of the Change

Allows for configuring quotas using memory values in terabytes.

All the code for converting values in TB to megabytes was already there. It was just limiting the valid input units to `m` and `g`. I've added `t` to that list.

I ran the unit tests locally and all the ones that ran passed. I got an error about failing to compile a database which failed the suite. I can't imagine how my change would break funcitonality elsewhere in the codebase.

## Why Is This PR Valuable?

This fixes a regression from v6. Configuring quotas in TB is valuable when configuring org quotas in large foundations.

## Applicable Issues

https://github.com/cloudfoundry/cli/issues/2069

### DIFF
--- a/command/flag/megabytes.go
+++ b/command/flag/megabytes.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	AllowedUnits = "mg"
+	AllowedUnits = "mgt"
 )
 
 type Megabytes struct {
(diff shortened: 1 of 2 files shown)


a-b:
I suspect this change should also go to the main (V8) branch.

PR #2177, opened 2021-07-15
Change logcache WalkBackOff Strategy to 250 ms

jdgonzaleza:
After the log cache team worked on [this spike](https://www.pivotaltracker.com/n/projects/2115367/stories/174515254) we went through and made the recommended change to backed off from log cache for 250 ms when there are no messages. 

@Benjamintf1 Does this looks good to you and is there anything else we could do from a validation perspective in order to make sure the performance issue is resolved.


Benjamintf1:
lgtm

a-b (approved):
LGTM

a-b:
The only suggestion is to consider making backoff configurable.

PR #2180, opened 2021-07-16
Delete v6 code

JenGoldstrich:
Deleted a lot of v6/v2 code that is no longer called, all legacy-plugin calls go through the ./cf directory in the root of the repo, so all of these packages were no longer used after the v3 service changes!

Deleted v2v3ation, v7action, command/v6, pushaction, and integration/v6, v2action, actionerrors, v6manifestparser, ccv2


a-b (approved):
LGTM

PR #2181, opened 2021-07-18
Submit adr for v8 code de-version

JenGoldstrich:
Submitting ADR 12 for discussion with the cf cli team and CF community members.

To clarify what I mean by de-version here, in the CLI code base there are a lot of directories/packages labeled as v6, v7, v2action, v3action, v7action, all with reference to some version of the cf CLI or CAPI.   I think this experience would be improved by trying as much as possible to remove the concept of version, and just having one version of a command per branch, check the markdown file in the PR for more info.

After reading that please feel free to leave feedback, emoji reactions, suggestions for changes, or any concerns on this PR.

### DIFF
--- a/doc/adr/0012-de-version-v8-code-base.md
+++ b/doc/adr/0012-de-version-v8-code-base.md
@@ -1,0 +1,37 @@
+# 1. De-version the V8 Code Base
+
+
+## Status
+
+PR Submitted
+
+## Context
+
+### What do you mean by de-version?
+
+De-version in this context means to remove the concept of a version from most of the CLI's codepaths and directory structure outside of the actual version command.
+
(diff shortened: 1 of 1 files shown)


a-b (approved):
I like the idea of simplifying `V8`. My main concern is the cost of sustaining. Right now project structure between all major versions is very similar. Propagating code updates with the proposed change will require more work. Which may is not that bad.

reedr3:
I really like this proposal and I think we'll be able to mitigate any changes in upkeep with cherry picking fixes back to v7

dominicrobertsc (approved):

reedr3 (approved):

jdgonzaleza (approved):
LGTM

PR #2183, opened 2021-07-27
Quote $PATH in Makefile to avoid word splitting

tjvman:
If the `$PATH` contains a path with a space, `make` can fail due to the `$PATH` being split into multiple words.

Does not modify the Golang source code.

## Description of the Change

Fixes a small quoting bug in the Makefile

## Why Is This PR Valuable?

Users affected by this bug can't run any `make` targets locally.

N/A

## Applicable Issues

N/A

## How Urgent Is The Change?

Not urgent, but also very simple.

## Other Relevant Parties

None.


a-b:
Hey @tjvman, can you please fix EasyCLA ?

Issue #2188, opened 2021-08-09
As a **app developer**, I can **set route destination protocols** when **mapping routes on the v8 CLI**


Gerg:
Copy-pasted from https://www.pivotaltracker.com/n/projects/2476914/stories/179076094

---

Given  I am using the v8 CLI

When I map a route to an app using the `--app-protocol` flag

Then I see that the resulting destination uses the specified protocol

```
$ cf map-route my-app example.com --app-protocol=http2
Creating route my-app.example.com for org my-org / space my-space as admin...
OK

Mapping route route my-app.example.com for org my-org / space my-space as admin...
OK
```

---

Consider if we should update the mapping flavor text to include the app protocol:
```
Mapping route route my-app.example.com for org my-org / space my-space as admin...
```

---

Handle TCP and internal routes

---

Consider whether `--app-protocol` is the best name for this flag

---

Notes:
- Lean on the API to implement most of the logic

---

Root issue: https://github.com/cloudfoundry/routing-release/issues/200

PR #2192, opened 2021-08-18
Add experimental destination-protocol flag to map-route

mariash:
Allows users to specify HTTP2 protocol can be used to reach application.

V8

## Description of the Change

Adding `--destination-protocol` flag to the map-route cli command. This will allow users to use HTTP2 protocol to communicate with their apps. It sets the `protocol` property in the map route destination in Cloud Controller which is passed to the Diego LRP and advertised to Gorouter. By default if `destination-protocol` is not set HTTP1 protocol will be used for HTTP route. We extended the validation when `map-route` is used for existing route: if everything is the same except for the destination protocol the error will be raised.

This command is documented in `map-route` command help message as an experimental command. We have not decided yet how this will be displayed for users in `routes` or another command. So there is a possibility that this flag name might change in the future.

## Why Is This PR Valuable?

Users will be able to use CLI to activate HTTP2 protocol to access their applications. 

This is a core functionality of the route destination. 

## Applicable Issues

cloudfoundry/cli#2188
cloudfoundry/routing-release#200

## How Urgent Is The Change?

We want to include this in V8.

## Other Relevant Parties

@ctlong @Gerg

### DIFF
--- a/command/v7/map_route_command.go
+++ b/command/v7/map_route_command.go
@@ -8,17 +8,19 @@ import (
 type MapRouteCommand struct {
 	BaseCommand
 
-	RequiredArgs    flag.AppDomain   `positional-args:"yes"`
-	Hostname        string           `long:"hostname" short:"n" description:"Hostname for the HTTP route (required for shared domains)"`
-	Path            flag.V7RoutePath `long:"path" description:"Path for the HTTP route"`
-	Port            int              `long:"port" description:"Port for the TCP route (default: random port)"`
-	relatedCommands interface{}      `related_commands:"create-route, routes, unmap-route"`
+	RequiredArgs        flag.AppDomain   `positional-args:"yes"`
+	Hostname            string           `long:"hostname" short:"n" description:"Hostname for the HTTP route (required for shared domains)"`
+	Path                flag.V7RoutePath `long:"path" description:"Path for the HTTP route"`
+	Port                int              `long:"port" description:"Port for the TCP route (default: random port)"`
+	DestinationProtocol string           `long:"destination-protocol" description:"[Beta flag, subject to change] Protocol for the route destination (default: http1). Only applied to HTTP routes"`
+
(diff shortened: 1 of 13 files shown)


ctlong:
Steps to test:
* Claim an env
* Upload releases and redeploy:
  * latest diego-release
  * latest routing-release
  * capi-release with [this cc_ng PR](https://github.com/cloudfoundry/capi-release)
* Push an app, curl `GET /v3/routes/:guid/destinations`, and validate the response includes `"protocol": "http1"`
* Run `cf map-route APP_NAME DOMAIN --hostname HOSTNAME --destination-protocol http2` with the same inputs for `DOMAIN` and `HOSTNAME` that the push used, and validate that the response is an error due to a pre-existing route mapping with a different protocol
* Unmap the route on the app
* Run `cf map-route APP_NAME DOMAIN --hostname HOSTNAME --destination-protocol http2` with the same inputs as before and see it succeed
*  curl `GET /v3/routes/:guid/destinations`, and validate the response includes `"protocol": "http2"`
* Create a new route, and then map your app to that route with `--destination-protocol tcp` and see it fail with an error related to trying to set tcp as the protocol for an http route

command/v7/map_route_command.go
```diff
@@ -106,7 +121,7 @@ func (cmd MapRouteCommand) Execute(args []string) error {
 		cmd.UI.DisplayOK()
```
ctlong, line 121: Should we include some extra error here if the protocol is set and doesn't match the retrieved route destination's protocol? Otherwise the user may believe that they've set the protocol successfully.

JenGoldstrich: I agree, we should add an extra statement here to indicate that nothing has changed.

JenGoldstrich:
Looks good to me, merging with the update that Carson listed above, tested it manually with @ctlong, @mariash, and @jdgonzaleza 

Still open questions about int testing of the http2 feature in CF but that can be solved in another PR

PR #2195, opened 2021-08-20
Add a CAPI version check for HTTP/2 to certain integration tests

ctlong:

v8

## Description of the Change

Adds a new CC version check to an integration test that (probably?) depends on certain CAPI. Paves the way for future HTTP/2 integration tests to be written in CLI that won't break pipelines.

Also adds one new integration test for if the route destination doesn't already exist and the DestinationProtocol flag is set.

## Why Is This PR Valuable?

* (Maybe) fix breaking pipelines
* No pressure to upgrade environments that CLI uses for integration tests to enable HTTP/2
* Easier testing of HTTP/2-specific tests, just point the integration tests at a CAPI of proper version and it'll work!

Fixes current core functionality.

## Applicable Issues

cloudfoundry/routing-release#200

## How Urgent Is The Change?

Somewhat urgent depending on if the last HTTP/2 PR is breaking pipelines.

### DIFF
--- a/api/cloudcontroller/ccversion/minimum_version.go
+++ b/api/cloudcontroller/ccversion/minimum_version.go
@@ -10,4 +10,6 @@ const (
 
 	MinVersionCreateServiceBrokerV3            = "3.72.0"
 	MinVersionCreateSpaceScopedServiceBrokerV3 = "3.75.0"
+
+	MinVersionHTTP2RoutingV3 = "3.104.0"
 )
--- a/integration/v7/isolated/map_route_command_test.go
+++ b/integration/v7/isolated/map_route_command_test.go
@@ -3,6 +3,7 @@ package isolated
 import (
 	"fmt"
 
+	"code.cloudfoundry.org/cli/api/cloudcontroller/ccversion"
 	. "code.cloudfoundry.org/cli/cf/util/testhelpers/matchers"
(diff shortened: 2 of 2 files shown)


ctlong:
From https://github.com/cloudfoundry/cli/blob/master/.github/CONTRIBUTING.md#api-versioning:

> The feature has an appropriate version check in the command layer to prevent use of that feature if the targeted API is below the minimum version. Note: commands should FAIL prior to execution when minimum version is not met for specified functionality.

Should the CC version check be used to determine if users even see the option to use the beta flag @JenGoldstrich?

JenGoldstrich:
I think this should be fine so long as CAPI will accept http2 as a destination protocol as of that CAPI version, I was under the impression we'd need to change a CAPI property on all of our environments but if all we need is new enough CAPI this should work great.

ctlong:
There is a feature flag in the form of a bosh property, but it's in GoRouter, not CAPI. Since the CAPI HTTP/2 code is theoretically idempotent with the GoRouter and Diego HTTP/2 code, a CAPI version check should suffice.

I'm making some assumptions with choosing `3.104.0`. The last CAPI release had CC V3 version `3.103.0` and the next CAPI release should have the HTTP/2 code.

jdgonzaleza (approved):
LGTM

PR #2210, opened 2021-09-23
Detect CF-on-K8s in `cf api`

gcapizzi:
This makes `cf api` able to detect the `cf-on-k8s` flag in the root endpoint response and persist this information in the config. This will allow us to enable CF-on-K8s-specific behaviour for every following command.

Issue: cloudfoundry/cf-k8s-api#10


gcapizzi:
Should have opened this against our own fork's `master` 😓  closing!

Issue #2159, opened 2021-04-14
CF7 security-groups returns 414 error


scult:
**Please fill out the issue checklist below and provide ALL the requested information.**
- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
We have security groups that are bound to a lot of spaces when we try to run ```cf security-groups``` we get ```414 Request-URI Too Large```.

**What happened**
It seems like to get the name of the spaces that are bound to a security group the v7 CLI queries /v3/spaces and adds each space guid as an URL parameter. With security groups that have a lot of spaces attached this results in a URI that's larger than the 8KB that are supported by gorouter.
```
REQUEST: xxxx
[...]

PR #2229, opened 2021-11-08
[Bugfix] CF security-groups returns 414 error

jdgonzaleza:
As described in #2159 the `cf security-groups` command was not able to respond properly when there were too many spaces in the foundation. 
In order to fix this behavior, as done in other commands, we refactored the `getSecurityGroupSpaces` to use the batcher.

### DIFF
--- a/actor/v7action/security_group_test.go
+++ b/actor/v7action/security_group_test.go
@@ -14,6 +14,7 @@ import (
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3"
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant"
 	"code.cloudfoundry.org/cli/resources"
+	"code.cloudfoundry.org/cli/util/batcher"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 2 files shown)


actor/v7action/security_group_test.go
```diff
@@ -598,7 +599,7 @@ var _ = Describe("Security Group Actions", func() {
 			})
 		})
 
-		When("the request errors", func() {
+		FWhen("the request errors", func() {
```
reedr3, line 602: Hey, looks like there's still a focus on this test. That should be removed.

reedr3:
Hey, I left one comment to address, but otherwise, looks good to me!

actor/v7action/security_group_test.go
a-b: I think GitHub Actions can check that for any PR.

a-b: Or GitHub PR checks

a-b (changes requested):
As soon as @reedr3 comment resolved - LGTM

actor/v7action/security_group_test.go
jdgonzaleza: Fixed in [1005665](https://github.com/cloudfoundry/cli/pull/2229/commits/b1f6916a865a66b2f7d56e8ce3c7dfd6cc87596e)

PR #2233, opened 2021-11-22
Introduce CF-on-K8s support

gcapizzi:
This introduces the first batch of changes necessary to support CF-on-K8s. The details of how we changed the architecture to accomodate these changes are in the [proposal document](https://docs.google.com/document/d/1SM0AH0E9Oy0ej9EvJf0OrXZbxixEr2Imu59C3xgwnOw).

We think this is pretty much all changes needed for CF-on-K8s support and we'd love to see it backported to the next `v8.x` release, maybe marking it as _experimental_. Before we can call this stable we probably want to:
* review all commands that we can't possibly support on Kubernetes, and introduce specific error messages for CF-on-K8s users that try to run them;
* review commands that we _can_ support on Kubernetes in some capacity, but still don't, and implement them (e.g. [`cf oauth-token`](https://github.com/cloudfoundry/cf-k8s-controllers/issues/277));
* introduce an end-to-end test suite, similar to CATs but for CF-on-K8s.

We will make further PRs to address the points above.

/cc @cloudfoundry/cf-k8s


jdgonzaleza (approved):
LGTM... 
Every commit was previously reviewed in the eirini-forks repo 😄

PR #2234, opened 2021-11-22
[Feature] Add app-protocol to cf routes command output

jdgonzaleza:
Currently we do not show the app-protocol in the output of the cf routes command, so a new column must be added to the output.
Also, the apps mapped to the route should be displayed in the apps column and must be separated with commas.

When running the `cf routes` command I'm able to see the app-protocol column in the output and the value for it should be a comma separated list that contains http1 and/or http2. This is in the case that the route is an HTTP route. 

```
$ cf routes
space         host                     domain         port   path   protocol   app-protocol     apps                     service instance
MySpace       myHost                   domain.cli.fun               http       http1,http2      app1, app
```

### DIFF
--- a/actor/v7action/route_test.go
+++ b/actor/v7action/route_test.go
@@ -677,6 +677,7 @@ var _ = Describe("Route Actions", func() {
 							App: resources.RouteDestinationApp{
 								GUID: "app-guid-1",
 							},
+							Protocol: "http1",
 						},
 					},
 					SpaceGUID: "fake-space-1-guid",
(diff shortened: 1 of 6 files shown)


hjcalderon10 (approved):

actor/v7action/route_test.go
```diff
@@ -928,13 +936,18 @@ var _ = Describe("Route Actions", func() {
 
 				for i := 0; i < batcher.BatchSize*batches; i++ {
 					port := i + 1000
+					appProtocol := "http1"
+					if i%2 == 0 {
```
a-b, line 940: clever ;)

a-b (commented):
LGTM

Issue #2219, opened 2021-10-05
CF8 space-users does not show the SpaceSupporter Role


WeiQuan0605:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [ ] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
The output of "cf8 space-users OrgName Spacename" doesn't show the SpaceSupporter Role even though assign was successful.
 
```
cf8 space-users Orgname Spacename                                
Getting users in org Orgname / space Spacename as admin...

SPACE MANAGER
  admin (uaa)

SPACE DEVELOPER
 admin (uaa)

SPACE AUDITOR
  No SPACE AUDITOR found
```

**What happened**
as above.

**Expected behavior**
With cf8 space-users command, the SpaceSupporter role can be got.

**Exact Steps To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. Run `cf set-space-role USERNAME ORG SPACE ROLE`
2. show space users by role `cf space-users ORG SPACE`
3. See error

**Provide more context**
- platform and shell details ( e.g. Mac OS X 10.11 iTerm)
- version of the CLI you are running: cf8
- version of the CC API Release you are on: v1.117.0

**Notes regarding V6 and V7 CLI support:** 
- V6:
  - Minimum supported version of CF Deployment: v7.0.0 (CAPI Release: 1.74.0 (APIs 2.128.0 and 3.63.0))
  - Maximum supported version of CF Deployment: v13.4.0 (CAPI Release: 1.94.0 (APIs 2.149.0 and 3.84.0))
- V7:
  - Minimum supported version of CF Deployment: v13.5.0 (CAPI Release: 1.95.0 (APIs 2.150.0 and 3.85.0))

sweinstein22:
The commit that originally closed this was [reverted due to pipeline failures](https://github.com/cloudfoundry/cli/commit/950da16e784ae79de52216bac9d0ddb6cecc7320). We have two new PRs that should solve the issue

PR #2235, opened 2021-11-24
Add space supporter to space users command

MerricdeLauney:
V8 only

## Description of the Change
Adds space supporter to space users command

Space supporter is only available on certain versions of CAPI. To ensure
space user command behavior is always being tested at the integration
level, we created a new test that only would be run on newer versions of
CAPI where the space supporter is able to be created and retrieved by
this command.

## Why Is This PR Valuable?
Allows users to access new capi feature

## Applicable Issues
#2219

## How Urgent Is The Change?
not time sensitive


jdgonzaleza (approved):
LGTM

PR #2243, opened 2022-01-20
Fix destination protocol flake in map-route test

mariash:
destination protocol variable is changing in context and that causes a test pollution when tests run in parallel.


a-b:
LGTM, thanks for the prompt turnaround!

PR #2250, opened 2022-02-09
[Bugfix] Idempotency when trying to create a service twice

hjcalderon10:

It modifies CLI V8

## Description of the Change

For the _**cf cs**_ command, before V8 it was able to work in a idempotent way. The expected behavior was to be able to fulfill with an OK status the command even if the service name was already taken. Now, V8 fails in this scenario.
By looking into the code, this pattern was implemented but left the error handler when CAPI V3 answered with a 422 status code with the message
"the service instance name is taken".

Adding the message into the error handling function and everything works as expected.

## Why Is This PR Valuable?

Enabling an expected behavior to the create-service command.

Bugifx.

## How Urgent Is The Change?

## Other Relevant Parties


a-b (changes requested):
Please add tests

a-b (approved):
LGTM

PR #2261, opened 2022-03-09
Online release process with Github Actions

a-b:
The current state of CF CLI release infrastructure is offline and quite outdated. To make our release process more transparent and inclusive we decided to move to build, signing, and release processes online to Github Actions


a-b:
Delivered outside of this PR.

PR #2262, opened 2022-03-10
Synchronize master and v8

a-b:
Moving features and fixes from `master` branch to `v8


reedr3:
I'm going to merge this PR and create a second one to add commits from v8 to master so that they're fully in sync. But I think we need to clarify our branching strategy moving forward so this doesn't keep happening.

PR #2266, opened 2022-03-11
Fix selfcontained tests failing due to /whoami

matt-royal:

This modifies CLI v8. The same changes may be necessary for v9 as well.

## Description of the Change
The tests under integration/v7/selfcontained were failing due to a missing stub for the GET /whoami endpoint. This commit adds that stub so that the existing tests will pass.

## Why Is This PR Valuable?
See above

See above

## Applicable Issues

N/A

## How Urgent Is The Change?

Not urgent, but if you intend to run the selfcontained tests in your pipeline then this will be necessary to make them green.

## Other Relevant Parties

@gcapizzi @kieron-dev

### DIFF
--- a/integration/v7/selfcontained/kubernetes_auth_test.go
+++ b/integration/v7/selfcontained/kubernetes_auth_test.go
@@ -4,14 +4,15 @@ import (
 	"net/http"
 	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


jdgonzaleza (approved):
LGTM

PR #2267, opened 2022-03-16
Add auth header to logcache API requests for cf-on-k8s

matt-royal:
This PR modifies v8. This same behavior will need to exist in v9. How should we handle that?

## Description of the Change
Set the correct `Authorization` header when requesting logs on a cf-on-k8s deployment. This uses the same logic for generating the header as exists in the current CLI, but refactored for reuse. The behavior is unchanged for cf-on-vms

## Why Is This PR Valuable?
This PR enables future work for cf-on-k8s to support logging during push, etc.

Logging is part of core behavior and we've already made changes to support auth for cf-on-k8s in core.

## Applicable Issues
https://github.com/cloudfoundry/cf-k8s-controllers/issues/370

## How Urgent Is The Change?
This change is necessary to unblock logging work for cf-on-k8s.

## Other Relevant Parties
@cloudfoundry/cf-k8s


cf/api/repository_locator.go
```diff
@@ -100,7 +101,11 @@ func NewRepositoryLocator(config coreconfig.ReadWriter, gatewaysByName map[strin
 	}
 
 	logCacheURL := configV3.ConfigFile.LogCacheEndpoint
-	logCacheClient := command.NewLogCacheClient(logCacheURL, configV3, nil)
+	logCacheClient, err := logcache.NewClient(logCacheURL, configV3, nil, v7action2.NewDefaultKubernetesConfigGetter())
+	if err != nil {
+		panic("handle this error!")
```
matt-royal, line 106: Now that creating the logcache client can yield an error, we need to handle that. The existing code also panic on error a few lines up, so we copied that pattern. We could also change this function and the caller function to return errors, and then bubble that out into the main func if you prefer.

matt-royal:
This PR is built on top of https://github.com/cloudfoundry/cli/pull/2266, so that PR should be accepted first.

api/shared/wrap_for_cf_on_k8s.go
```diff
@@ -0,0 +1,124 @@
...
+		}, nil
+	}
+
+	if transportConfig.WrapTransport == nil {
+		// i.e. not auth-provider or exec plugin
+		return transport.HTTPWrappersForConfig(transportConfig, roundTripper)
+	}
+
+	// using auth provider to generate token
+	return transportConfig.WrapTransport(roundTripper), nil
```
gcapizzi, line 77: We have recently found our that [this check is redundant](https://github.com/kubernetes/client-go/blob/6889a6b513d4b6b374bec632433d39f5aede033d/transport/round_trippers.go#L39-L42).

reedr3: should that check be removed before we merge the PR?

reedr3 (approved):
LGTM

jdgonzaleza (approved):
LGTM

Once merged into v8, We should backport into main/master as this must be included in future cli versions 😬

a-b (commented):
LGTM

a-b:
> Once merged into v8, We should backport into main/master as this must be included in future cli versions 😬
It  is hard for me to imagine anything in v8 that does not belong to the `main` (future v9) branch.

api/shared/wrap_for_cf_on_k8s.go
gcapizzi: Yes, we might get away with just `return transportConfig.HTTPWrappersForConfig(...)` but I'd try it first, just in case.

Issue #2131, opened 2021-01-25
Support for Apple M1 chips


nirajhx:
**Please provide details on the following items. Failure to do so may result in deletion of your feature request.** 
Not able to install on Apple MacBook with M1 chipset 

**What's the user value of this feature request?**
Devs using the latest MacBook will be able to use cli

**Who is the functionality for?**
Devs using the latest M1 chip macbook

**How often will this functionality be used by the user?**
Always

**Who else is affected by the change?**
This should not affect the other versions

**Is your feature request related to a problem? Please describe.**
Trying to install CLI on my M1 chip MacBook, but the installation fails with following error:
```
==> Installing cf-cli from cloudfoundry/tap
cf-cli: The x86_64 architecture is required for this software.
Error: An unsatisfied requirement failed this build.
```

**Describe alternatives you've considered**
I tried compiling the source code, but that also didn't workout. Got following error:
```
go build -ldflags "-w -s -X code.cloudfoundry.org/cli/version.binarySHA=$(git rev-parse --short HEAD) -X code.cloudfoundry.org/cli/version.binaryBuildDate=$(date -u +"%Y-%m-%d") -X code.cloudfoundry.org/cli/version.binaryVersion=$(cat BUILD_VERSION) " -o out/cf .
# code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant
assertion failed [abi_info.kind == AbiKind::TranslatedCode]: emulated forward to an arm pc that isn't in translated code. arm_pc=0x1020cf4d8 abi_kind=6 emulation_interval=[0x1020f8df8,0x1020f8e0c) instruction_interval=[0x1020f8de4, 0x1020f8e0c) x86_rip=0x101c91a
(ThreadContextRegisterState.cpp:677 move_to_instruction_boundary)
```

a-b:
Hello @nirajhx, thanks for reporting this issue. Did you tried running our macOS binary using `Rosetta Shell`?

nirajhx:
Yeah, binary did work for me. Thanks 😊

biolauri:
Could someone re-open this issue? Albeit the workaround with `Rosetta Shell` does work, it'd be very great to have the cf cli native for ARM-based architecture.

a-b:
@biolauri do you know if it is possible to transpile (compile the binary for ARM on Linux) Golang for ARM?

a-b:
[...]

PR #2279, opened 2022-05-31
add option to build for linux arm64

gflex:

No but can be included

## Description of the Change

This change simply adds the option to build for linux arm64 in Makefile

## Why Is This PR Valuable?

Useful for those who need cf able to run on linux arm64

## Applicable Issues

Similar to https://github.com/cloudfoundry/cli/issues/2131
Suitable for those who run cf in Linux docker images under M1


pivotalgeorge:
This looks good in that the new target works for me on an M1 mac, however I cannot reproduce the issue with the existing amd64 target.
`docker run --platform linux/arm64 -v $PWD:/cli -w /cli golang:1.18 make out/cf-cli_linux_x86-64` followed by `docker run --platform linux/arm64 -v $PWD:/cli -w /cli golang:1.18 ./out/cf-cli_linux_x86-64 version` works for me on an M1 mac.

I'd like to see a change in behavior before approving--can you provide reproduction steps?
+@rafikt1992

pivotalgeorge:
Could you please provide reproduction steps for the existing mac os target not working on arm? I could not reproduce this in a docker container.

rafikt1992:
i don't have an example at them moment, but sometimes when i have a script that performs cf commands i get an error similar to this

`qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`

which is what i get when i run any cli that doesn't have an ARM64

the alternative is to run things under rosetta, or if they have a darwin M1 build. but this complicate the workflow as most of the tools we use in our development is dockerized

a-b (approved):
LGTM

beyhan:
Hello @a-b,

Do you plan to offer build binaries for linux-arm64 after this is merged?

PR #2285, opened 2022-06-15
feat: add --wait flag to upgrade-service

blgm:

This PR is aimed for vNext (v9), but could be cherry-picked to v8 if that was considered appropriate.

## Description of the Change

Most of the service commands (e.g. "cf update-service") take a --wait
flag to wait for completion. For some reason "cf upgrade-service" did
not, and this commit resolves that inconsistency.

## Why Is This PR Valuable?

It makes the service commands more usable and more consistent

It's a modification to a core command

## Applicable Issues

None

## How Urgent Is The Change?

Not urgent at all

## Other Relevant Parties

No-one


jdgonzaleza (approved):
LGTM

PR #2294, opened 2022-07-14
Refresh token in logs is disabled for Korifi

Birdrock:
CLI v8

## Description of the Change
Disables refresh token call for the `cf logs` command when targeting Korifi (CF on K8s). We are not using UAA for AuthN/Z, so this call was resulting in errors.

## Why Is This PR Valuable?

This will allow users of Korifi to use the `cf logs` command without causing an error on the refresh token call.

This change extends previous core functionality to allow targeting of Korifi (CF on K8s) clusters.

## Applicable Issues

[[Feature]: CF CLI user can skip oauth/token requests on Korifi](https://github.com/cloudfoundry/korifi/issues/1299)

## How Urgent Is The Change?

Medium/high urgency - we would like to pair it with a beta release. This functionality will be needed for standard user flows.

## Other Relevant Parties

Korifi users.

### DIFF
--- a/command/v7/logs_command.go
+++ b/command/v7/logs_command.go
@@ -58,19 +58,22 @@ func (cmd LogsCommand) Execute(args []string) error {
 		return cmd.displayRecentLogs()
 	}
 
-	stop := make(chan struct{})
-	stoppedRefreshing := make(chan struct{})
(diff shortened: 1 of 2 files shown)


jdgonzaleza (approved):
LGTM

a-b (approved):
LGTM

PR #2295, opened 2022-07-14
Refresh token in logs is disabled for Korifi

Birdrock:
CLI v8. A corresponding PR is open for the v8 branch with #2294.

## Description of the Change
Disables refresh token call for the `cf logs` command when targeting Korifi (CF on K8s). We are not using UAA for AuthN/Z, so this call was resulting in errors.

## Why Is This PR Valuable?

This will allow users of Korifi to use the `cf logs` command without causing an error on the refresh token call.

This change extends previous core functionality to allow targeting of Korifi (CF on K8s) clusters.

## Applicable Issues

[[Feature]: CF CLI user can skip oauth/token requests on Korifi](https://github.com/cloudfoundry/korifi/issues/1299)

## How Urgent Is The Change?

This PR to update master is lower priority than #2294.

## Other Relevant Parties

Korifi users.

### DIFF
--- a/command/v7/logs_command.go
+++ b/command/v7/logs_command.go
@@ -58,19 +58,22 @@ func (cmd LogsCommand) Execute(args []string) error {
 		return cmd.displayRecentLogs()
 	}
 
-	stop := make(chan struct{})
-	stoppedRefreshing := make(chan struct{})
-	stoppedOutputtingRefreshErrors := make(chan struct{})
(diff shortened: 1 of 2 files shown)


MerricdeLauney:
Requested review from cli team as i no longer have cli context

jdgonzaleza:
Closing and moving conversation to https://github.com/cloudfoundry/cli/pull/2294.
If this is mento to go to v8 then we first review the pr to v8 and then cherry-pick

PR #2303, opened 2022-08-02
Push and scale app and run-task

Benjamintf1:
* `push` accepts --log-rate-limit / -l flag.
* `scale` accepts --log-rate-limit / -l flag.
* `run-task` accepts --log-rate-limit / -l flag.

### DIFF
--- a/command/v7/shared/app_summary_displayer_test.go
+++ b/command/v7/shared/app_summary_displayer_test.go
@@ -16,6 +16,9 @@ import (
 )
 
 var _ = Describe("app summary displayer", func() {
+
+	const instanceStatsTitles = `state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details`
+
 	var (
 		appSummaryDisplayer *AppSummaryDisplayer
 		output              *Buffer
@@ -249,7 +252,7 @@ var _ = Describe("app summary displayer", func() {
 				It("lists instance stats for process types that have > 0 instances", func() {
 					Expect(testUI.Out).To(Say(`type:\s+web`))
 					Expect(testUI.Out).To(Say(`sidecars: `))
-					Expect(testUI.Out).To(Say(`state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details`))
+					Expect(testUI.Out).To(Say(instanceStatsTitles))
 				})
 
 				It("does not show the instance stats table for process types with 0 instances", func() {
@@ -279,7 +282,7 @@ var _ = Describe("app summary displayer", func() {
 				It("displays the instances table", func() {
 					Expect(testUI.Out).To(Say(`type:\s+web`))
 					Expect(testUI.Out).To(Say(`sidecars: `))
-					Expect(testUI.Out).To(Say(`state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details`))
+					Expect(testUI.Out).To(Say(instanceStatsTitles))
 				})
 			})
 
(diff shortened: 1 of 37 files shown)


command/v7/shared/app_summary_displayer_test.go
```diff
@@ -197,7 +249,7 @@ var _ = Describe("app summary displayer", func() {
 				It("lists instance stats for process types that have > 0 instances", func() {
 					Expect(testUI.Out).To(Say(`type:\s+web`))
 					Expect(testUI.Out).To(Say(`sidecars: `))
-					Expect(testUI.Out).To(Say(`state\s+since\s+cpu\s+memory\s+disk\s+details`))
+					Expect(testUI.Out).To(Say(`state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details`))
```
jdgonzaleza, line 252: (Very) Nitpicky: Perhaps we can have a constant with the "title column" `state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details` ? 
I'm not against of the current implementation. It's just an idea.

fixtures/applications/example-app/Gemfile.lock
```diff
@@ -3,7 +3,7 @@ GEM
   specs:
     mustermann (1.1.1)
       ruby2_keywords (~> 0.0.1)
-    rack (2.2.3)
+    rack (2.2.3.1)
```
jdgonzaleza, line 6: Why was this changed?

integration/helpers/app_instance_table.go
```diff
@@ -58,8 +59,8 @@ func ParseV3AppProcessTable(input []byte) AppTable {
 			// instance row
 			columns := splitColumns(row)
 			details := ""
-			if len(columns) >= 7 {
-				details = columns[6]
+			if len(columns) >= 8 {
```
jdgonzaleza, line 62: Is it 8 the number of columns we have? if that's the case, could we create a constant for that instead of hardcoding it ?

integration/v7/isolated/restart_command_test.go
```diff
@@ -105,7 +105,7 @@ var _ = Describe("restart command", func() {
 					Eventually(session).Should(Say(`type:\s+web`))
 					Eventually(session).Should(Say(`instances:\s+1/1`))
 					Eventually(session).Should(Say(`memory usage:\s+\d+(M|G)`))
-					Eventually(session).Should(Say(`\s+state\s+since\s+cpu\s+memory\s+disk\s+details`))
+					Eventually(session).Should(Say(`\s+state\s+since\s+cpu\s+memory\s+disk\s+logging\s+details`))
```
jdgonzaleza, line 108: Same (very) nitpicky suggestion as above. 😄

jdgonzaleza (approved):
LGTM

integration/helpers/app.go
```diff
@@ -369,3 +369,17 @@ func WaitForAppDiskToTakeEffect(appName string, processIndex int, instanceIndex
...
+		session := CF("restart", appName)
+		Eventually(session).Should(Exit(0))
+	}
+
+	Eventually(func() string {
+		session := CF("app", appName)
+		Eventually(session).Should(Exit(0))
+		appTable := ParseV3AppProcessTable(session.Out.Contents())
+		return appTable.Processes[processIndex].Instances[instanceIndex].LogRate
+	}).Should(MatchRegexp(fmt.Sprintf(`\d+(\.\d+)?[KMG]?/s of %s/s`, expectedLogRateLimit)))
```
pivotalgeorge, line 384: this looks okay in the sense that it follows the pattern of the other WaitFor helpers above, but we'll have to keep an eye on whether we see an increase in pipeline flakes.

pivotalgeorge: nevermind. didn't grok that ShouldRestartFirst is flagged.

pivotalgeorge (approved):
Looks good + units pass for me

a-b (commented):
LGTM

PR #2316, opened 2022-09-13
Return error for log rate limits larger than math.MaxInt

rroberts2222:
This fix prevents the CLI from accepting a number
that would overflow for the log rate limit.

Co-authored-by: Rebecca Roberts <robertsre@vmware.com>

v8

## Description of the Change

CLI would overflow when a huge log rate limit is passed. This could cause the CLI to send a large negative number to CAPI. 
This change returns an error to prevent this.

## How Urgent Is The Change?

Not particularly urgent-- change to fix an edge case.

## Other Relevant Parties
 None

### DIFF
--- a/command/flag/bytes_with_unlimited.go
+++ b/command/flag/bytes_with_unlimited.go
@@ -1,6 +1,7 @@
 package flag
 
 import (
+	"math"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


a-b (approved):
LGTM, thanks for writing the descriptive test.

PR #2320, opened 2022-09-16
Add ADR for new Plugin API

will-gant:
Adds a proposed Architecture Decision Record (ADR) to create a new Plugin API Interface.

### DIFF
--- a/doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
+++ b/doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
@@ -1,0 +1,56 @@
+# 13. Provide a Plugin API backed by CAPI V3
+
+Date: 2022-09-16
+
+## Status
+
+Submitted
+
+## Context
+
+_The issue motivating this decision, and any context that influences or constrains the decision._
+
+Although all V8 CLI commands now call [V3](https://v3-apidocs.cloudfoundry.org/) of the CAPI API, CLI plugins are only able to call a Plugin API Interface whose methods invoke a 'legacy' version of the CLI that currently exists in the [`cf`](../../cf/) directory. This is, in effect, V6 of the CLI, which exclusively uses CAPI [V2](https://apidocs.cloudfoundry.org/). The last V6 release was in October 2020, and the `cf` directory has had few commits since then.
+
+The Plugin API exposes ten common CLI commands (e.g. `cf services`) via [methods](https://github.com/cloudfoundry/cli/blob/master/plugin/plugin.go#L17-L48) which return Go structs in the shape of CAPI V2 API resources. Other V6 CLI commands can be invoked by making a [CliCommand](https://github.com/cloudfoundry/cli/blob/d3bb1ecfc94ee12432f8defb30ab9459e83cc727/plugin/cli_connection.go#L72-L74) or [CliCommandWithoutTerminalOutput](https://github.com/cloudfoundry/cli/blob/d3bb1ecfc94ee12432f8defb30ab9459e83cc727/plugin/cli_connection.go#L68-L70), which both take string arguments and return a slice of strings. These methods can be used with `cf curl`, which allows users to make authenticated CAPI API calls, and some plugin authors do so for both [V2](https://github.com/orange-cloudfoundry/cf-plugin-bg-restage/blob/bcb1c4c712a43119d63d849b4b368eb65a86eb3b/repo.go#L72-L80) and [V3](https://github.com/generalmotors/cf-restage-all/blob/d6d9747182881a2b50c7bc601f4762778bb1cbfd/restage_all.go#L211-L237) calls.
+
+In early 2020 the CLI team developed a working POC for a new Plugin API backed by CAPI V3. Although included in beta releases of V7 of the CLI, these changes were mostly [reverted](https://www.pivotaltracker.com/n/projects/892938/stories/171032219) before general availability due to concerns at the scale of breaking changes for plugins, many of which are no longer maintained.
+
+## Decision
+
+_The change that we're proposing or have agreed to implement._
+
+Create a Plugin API V2 that exposes methods that plugins can use to call the same CLI commands used by V8 of the CLI today. Defer making a decision on how long to continue supporting the legacy Plugin API until the uptake of Plugin API V2 is apparent.
+
+The Plugin API V2 will be similar to that developed in early 2020, making use of the same `V7Actor` that V8 of the CLI uses to make V3 CAPI calls. Plugin metadata will declare whether the plugin in question supports the Plugin API V2. If it does not it will be run with the legacy Plugin API.
+
+## Consequences
+
+_What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated._
+
+### Benefits
+
+* **Avoids a hard switch that would break a large proportion of plugins**
+    * A hard switch to an entirely new Plugin API serving CAPI V3-shaped structs would break most plugins that invoke CLI commands.
+    * Introducing a new Plugin API that attempted to mimic its legacy predecessor by transforming V3 CAPI resources into V2-shaped structs would be disproportionately complex. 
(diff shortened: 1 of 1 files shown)


a-b (commented):
LGTM

a-b:
I'm curious about what @acrmp and @XenoPhex are thinking about this proposal.

doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
```diff
@@ -0,0 +1,56 @@
...
+
+In early 2020 the CLI team developed a working POC for a new Plugin API backed by CAPI V3. Although included in beta releases of V7 of the CLI, these changes were mostly [reverted](https://www.pivotaltracker.com/n/projects/892938/stories/171032219) before general availability due to concerns at the scale of breaking changes for plugins, many of which are no longer maintained.
+
+## Decision
+
+_The change that we're proposing or have agreed to implement._
+
+Create a Plugin API V2 that exposes methods that plugins can use to call the same CLI commands used by V8 of the CLI today. Defer making a decision on how long to continue supporting the legacy Plugin API until the uptake of Plugin API V2 is apparent.
+
+The Plugin API V2 will be similar to that developed in early 2020, making use of the same `V7Actor` that V8 of the CLI uses to make V3 CAPI calls. Plugin metadata will declare whether the plugin in question supports the Plugin API V2. If it does not it will be run with the legacy Plugin API.
```
XenoPhex, line 25: I’m unfamiliar with the plugin API that was developed in early 2020 - can y’all provide it for additional context / update this ADR with information surrounding said API?

Without that additional information, I’m left assuming that the proposal is suggesting that the general API remains the same as the current / V1 plugin API, but backed by the V3 API and returns V3 API like objects. This likely means the plugin API will have additional function calls for the newly added V3 resources.

If these assumptions are in line with this proposal than here are my suggestions (if not ignore and I’ll provide an updated response once I have the appropriate information):
1. Switch from existing RPC to a standardized gRPC framework. I honestly don’t remember why this was, I just remember having dozens of conversations with people and the conclusion ending here.
2. Be able to execute [g]RPC calls in a stream-like fashion (ie able to stream STDIN, STDOUT, STDERR as the API server executes requested API calls). This is especially important for plugins that are executing long running functions like `push`.
3. Have a `json` or “structured” output for every CLI function would significantly help plugin authors when parsing the output of certain plugin API calls.
4. _Do not directly tie the V2 Plugin API interface to the V3 API backed actors._ As long as there’s a defined, but separate, plugin API interface the V3 API backed actors can move freely / change without affects / breaking the V2 Plugin API.
5. Indirectly related - when plugins are installed, the `cf` CLI did not keep track of all the necessary information needed for safe / automated upgrade processes. Updating the `cf` CLI to store the following would help with that:
  - Where the plugin was installed from (repo source name, id, etc. Something verifiable?)
  - How the plugin was installed (installed from repo or local or something else?)

doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
```diff
@@ -0,0 +1,56 @@
...
+
+## Consequences
+
+_What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated._
+
+### Benefits
+
+* **Avoids a hard switch that would break a large proportion of plugins**
```
XenoPhex, line 33: Honestly - I would just break / replace the API along side a major version bump the CF CLI. It’s been a few years since I’ve been at VMware, but even before I left, I knew a lot of plugin authors were just shelling out to `cf curl` to call unprovided API endpoints / V3 API to do what they wanted. Given how much the current plugin API has languished, and the fact that the majority of plugins are years old / most likely no longer maintained, a hard break might actually be easier for everyone to handle.

doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
will-gant: Thanks for pondering this. For background on the abortive effort in early 2020, see [this epic](https://www.pivotaltracker.com/n/projects/892938/epics/4315861). The last commit before those changes were reverted was [this one](https://github.com/cloudfoundry/cli/tree/0d0655662c3143a6275dd7feed1ca40a0d2f89a2). [This doc](https://github.com/cloudfoundry/cli/blob/master/doc/v7-plugin-api-changes.md) also described some of the changes in the new Plugin API.

Having done worked only a tiny bit with gRPC before, I'll defer to others on this, but I do like the idea of being able to generate server and client-side code. And that could help achieve 3. - I agree that it'd be a really good thing to totally eliminate the need to parse responses as slices of strings.

doc/adr/0013-provide-a-plugin-api-backed-by-capi-v3
will-gant: Ah man I wrote a reply to this that I obviously didn't actually send... Doh!

Ok, so essentially my concern here is that we would end up in a situation where a large proportion of users just stick with whatever is the final version of the CLI is that supports legacy plugins, and stay on that forever.

I'm not sure proper data on usage exists, but my own anecdotal experience at various companies (I work for a consultancy) is that old, unmaintained plugins are used widely - both by humans and by tools like CI pipelines and scripts. A hard switch would cause these things to suddenly break all over the place, and the only good solution for a lot of people would be to ensure they use an older version of the CLI.

Some unknown but large proportion of plugins will never be updated by their original authors, who are long gone. Some wouldn't even respond to a PR if someone else did the work for them. 

I can't imagine many places, for example, building one image with an old version of `cf` for scripts that need to run plugin commands, and a newer one for ones that don't. So I'm concerned they'd default to using the older one for _everything_, which would be a worse situation than we've got right now.

But an important part of my ADR the view that resolving the question of how and when to retire the legacy plugin interface should not be a prerequisite to implementing a V2 interface. The latter should not be delayed by the former, regardless of whether a decision is made to deprecate it in the short-term, long-term or never.

XenoPhex: > But an important part of my ADR the view that resolving the question of how and when to retire the legacy plugin interface should not be a prerequisite to implementing a V2 interface. The latter should not be delayed by the former, regardless of whether a decision is made to deprecate it in the short-term, long-term or never.

Absolutely - I think regardless of the approach in handling the previous API, you shouldn't delay/bound the work based on that decision. My only [slight] concern around this is how much flexibility you'll have to implement the new interface while you're "maintaining" the old one.

> Ok, so essentially my concern here is that we would end up in a situation where a large proportion of users just stick with whatever is the final version of the CLI is that supports legacy plugins, and stay on that forever.

This is going to happen regardless of the strategy, a "transition" based model will just delay the version this happens with. The question is whether or not the team wants to take on the additional burden of maintaining multiple plugin interfaces at once (including how to manage the transition for a number of _years_). Unfortunately, without proper metrics on CLI, plugins and related interfaces usage, it's hard to say what strategy [with the most optimal user experience per developer cost] would work best here. In my opinion / experience, it's easier [for everyone] to just cut and move forward. If my memory of working with support teams / on customer issues remain accurate, most [paying] customers don't (usually) upgrade their CLIs for upwards of a year+; this gives plugin developers (at least the ones that are still active), more than enough time to upgrade their plugins to a new interface. The rest of those plugins are never going to get updated and it's not up to the CLI team to maintain a mess because other companies don't want to pay to update tools they rely on. (IE stop doing _unnecessary_ free labor for other companies.)

will-gant: Yeah, it's a fair point that there'd sometimes be work needed to support the legacy API. But if the past couple of years are a guide, it looks like it requires a fairly small portion of the team's efforts. 

I generated the output below with a [git-extras](https://github.com/tj/git-extras) command. It's not perfect (it just counts commits, regardless of size), but does show that the `cf` and `plugin` directories are updated pretty infrequently.

```console
# git effort * -- --since='Oct 18 2020'

path                                                                                       commits    active days

integration............................................................................... 103         68
command................................................................................... 92          60
actor..................................................................................... 75          43
api....................................................................................... 48          34
util...................................................................................... 24          14
go.mod.................................................................................... 20          12
resources................................................................................. 19          14
go.sum.................................................................................... 19          13
Makefile.................................................................................. 18          13
README.md................................................................................. 12          6
cf........................................................................................ 8           8
doc....................................................................................... 5           4
bin....................................................................................... 4           4
tools..................................................................................... 2           2
fixtures.................................................................................. 2           2
types..................................................................................... 1           1
plugin.................................................................................... 1           1
main.go................................................................................... 1           1
```

PR #2321, opened 2022-09-16
Allow user to rename app if buildpack is deleted

moleske:
cc @reedr3 

Going to default branch, probably will request a backport to v8

## Description of the Change
A user had an app that they wanted to rename, but the buildpack the app was using had previously been deleted from the cloud foundry deployment.  When the user called `cf rename` they received an error about the buildpack not being available.  CAPI supports only renaming an app, but the cf cli is passing not just the rename on `cf rename` but the stack and buildpack information.  This pr allows only rename by only passing to capi the new name information and not the stack and buildpack information.

## Why Is This PR Valuable?

CAPI allows the user to only do renames via its API so we should also support this

Not new functionality, just fixing existing functionality.

## Applicable Issues

## How Urgent Is The Change?

## Other Relevant Parties

### DIFF
--- a/actor/v7action/application.go
+++ b/actor/v7action/application.go
@@ -400,6 +400,17 @@ func (actor Actor) UpdateApplication(app resources.Application) (resources.Appli
 	return updatedApp, Warnings(warnings), nil
 }
 
+// UpdateApplicationName updates the name of an application
+func (actor Actor) UpdateApplicationName(newAppName string, appGUID string) (resources.Application, Warnings, error) {
+
+	updatedApp, warnings, err := actor.CloudControllerClient.UpdateApplicationName(newAppName, appGUID)
+	if err != nil {
+		return resources.Application{}, Warnings(warnings), err
+	}
+
+	return updatedApp, Warnings(warnings), nil
+}
+
 func (actor Actor) getDeployment(deploymentGUID string) (resources.Deployment, Warnings, error) {
 	deployment, warnings, err := actor.CloudControllerClient.GetDeployment(deploymentGUID)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


moleske:
There's a few questions in the commit messages, but the one I'm most concerned for is should there be an integration test added.

Current Integration tests all passed on a cf-deployment with capi 1.137.0, though I didn't enable some experimental features like route_sharing so didn't test those

PeteLevineA (approved):

reedr3 (approved):
looks good to me!

actor/v7action/application.go
```diff
@@ -400,6 +400,17 @@ func (actor Actor) UpdateApplication(app resources.Application) (resources.Appli
 	return updatedApp, Warnings(warnings), nil
 }
 
+// UpdateApplicationName updates the name of an application
+func (actor Actor) UpdateApplicationName(newAppName string, appGUID string) (resources.Application, Warnings, error) {
+
+	updatedApp, warnings, err := actor.CloudControllerClient.UpdateApplicationName(newAppName, appGUID)
+	if err != nil {
+		return resources.Application{}, Warnings(warnings), err
```
jdgonzaleza, line 408: Any reason why we are returning resources.Application{} instead of nil?

jdgonzaleza: seems like the norm here... :)

jdgonzaleza:
> There's a few questions in the commit messages, but the one I'm most concerned for is should there be an integration test added.
> 
> Current Integration tests all passed on a cf-deployment with capi 1.137.0, though I didn't enable some experimental features like route_sharing so didn't test those

@moleske Our pipelines have experimental features such as route_sharings already enabled :). 
Also, I'm wondering if by modifying the current implementation we will be changing the expectation of current clients that are waiting for buildpacks and stacks to change when running `cf rename` . What do you think about this?

moleske:
> Also, I'm wondering if by modifying the current implementation we will be changing the expectation of current clients that are waiting for buildpacks and stacks to change when running cf rename . What do you think about this?

@PeteLevineA and I thought about this.  We think it is not changing the expectation of the current clients.  This is because `cf rename` says it will just rename an application.  However the current behavior is that it will also attempt to restage with the buildpack and stack, even though the user didn't request those to be changed by `cf rename`.  The support issue we saw was `cf rename` failed because the buildpack was renamed.  If they wanted to change the buildpack or stack, they can use `cf push -b <newbuildpack> -s <newstack>`

jdgonzaleza (approved):
LGTM

PR #2326, opened 2022-09-30
GitHub Actions Integration Tests

a-b:
## Description of the Change

This introduces CF CLI integration tests


a-b:
I propose to have workflow names use shared namespace-like prefixes. For example 

| Current | Proposed |
|--------|--------|
| .github/workflows/setup-cf-env.yml | .github/workflows/cf-env-setup.yml |
| .github/workflows/unclaim-cf-env.yml | .github/workflows/cf-env-unclaim.yml |

.github/workflows/setup-cf-env.yml
```diff
@@ -0,0 +1,95 @@
...
+          ./tmp/aws/install
+
+          wget https://s3.amazonaws.com/bosh-cli-artifacts/bosh-cli-7.0.1-linux-amd64 --output-document="/usr/bin/bosh"
+          chmod +x /usr/bin/bosh
+
+      - name: Deploy edge CAPI
+        if: ${{ inputs.capi-version == 'edge' }}
+        run: |
+          # find latest capi
+          FILENAME="$(aws s3 ls capi-releases --no-sign-request --recursive --region us-east-1 | sort | tail -n 1 | awk '{print $4}')"
```
PeteLevineA, line 84: I assume this means it is possible to have the same release version and we're just grabbing the latest? Or that since this is edge just always grab the latest and we don't actually know the latest version of capi

PeteLevineA (approved):

moleske:
We don't know what the latest edge is, we just grab it.  This is from the output of [capi-create-release](https://ci.cake.capi.land/teams/main/pipelines/capi/jobs/create-capi-release), which passed unit tests but has not passed integration tests

.github/workflows/setup-cf-env.yml
moleske: We don't know what the latest edge is, we just grab it. This is from the output of [capi-create-release](https://ci.cake.capi.land/teams/main/pipelines/capi/jobs/create-capi-release), which passed unit tests but has not passed integration tests

Future PRs will allow specification of CAPI, to support running tests with a minimum supported version of CAPI

.github/workflows/setup-cf-env.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Install Tools
+        run: |
+          wget https://github.com/cloudfoundry/bosh-bootloader/releases/download/v8.4.110/bbl-v8.4.110_linux_x86-64 -P /tmp
+          mv /tmp/bbl-* /usr/local/bin/bbl
+          chmod +x /usr/local/bin/bbl
+
+          wget https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip -P /tmp
+          unzip -d /tmp /tmp/awscli-exe-linux-x86_64.zip
+          ./tmp/aws/install
```
moleske, line 75: This needs to be the absolute path `/tmp/aws/install` instead of the relative path `./tmp/aws/install`

.github/workflows/unclaim-cf-env.yml
```diff
@@ -0,0 +1,43 @@
...
+jobs:
+  build-env:
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+
+    steps:
+      - name: Download metadata
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.identifier-for-metadata-file }}
```
moleske, line 22: some testing in a private space seems to indicate that this line is not quite correct.  I'm getting a problem where this input is empty (`''`) causing all artifacts to be downloaded, which results in a pathing problem.  For example, if the environment name is `rosebud` then instead of downloading just the metadata json file, it puts in the json file in a directory called `rosebud`.  This causes the next step to fail on `cat metadata.json | jq -r .name` as the file is actually `rosebud/metadata.json`

I'm hoping I'm just not seeing something silly, like the input is badly passed in `integration.yml` or incorrectly exported out `setup-cf-env.yml`

[github docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow) for reference on anyone trying to help

.github/workflows/setup-cf-env.yml
```diff
@@ -0,0 +1,95 @@
...
+      environment:
+        required: true
+        type: string
+      capi-version:
+        required: true
+        type: string
+    outputs:
+      environment-name:
+        description: "Name of claimed environment"
+        value: ${{ jobs.setup-cf-env.outputs.environment-name }}
```
moleske, line 15: this should be `${{ steps.claim-toolsmiths-env.outputs.environment-name }}` as we want the output of the step, not an out put of the job.  See [this](https://github.com/cloudfoundry/cli/pull/2326/files#r986032667) for original issue

.github/workflows/unclaim-cf-env.yml
moleske: Ok thinking harder for a few minutes, the problem is actually over [here](https://github.com/cloudfoundry/cli/pull/2326/files#r986041041)

reedr3 (approved):
This looks good to me. I agree that integration tests should not be a blocker on PRs until we've addressed the failures/flakiness.

PR #2331, opened 2022-10-13
Explicity enable route_sharing feature on tests that need it

moleske:
- previously tests were not stating they needed this feature enable, only a minimum version of capi
- since the feature is not enabled by default, this adds a call to enable it
- attempt to disable it after each test, however when tests run in parallel, this caused the tests to fight each other on turning it on and off.  Since a minimum capi version is checked and we don't have tests that explicitly test the feature being disabled, it seems ok to just leave it on

*I checked the "allow edits and access to secrets by maintainers, so any approver should be able to push commits if they see small issues*

Modifies tests to make github actions integration test work better, so should be backported to all versions still under testing

## Description of the Change
Explicitly enable route_sharing feature on tests that need it

## Why Is This PR Valuable?
Will allow folks to just run the tests instead of configuring their cloud foundry install

n/a

## Applicable Issues

#2326 

## How Urgent Is The Change?
As urgent as we want integration tests to work on github actions

## Other Relevant Parties
@cloudfoundry/wg-app-runtime-interfaces-cli-approvers 
@cloudfoundry/wg-app-runtime-interfaces-cli-reviewers

### DIFF
--- a/integration/helpers/commonisolated/common_isolated_setup.go
+++ b/integration/helpers/commonisolated/common_isolated_setup.go
@@ -1,87 +1,87 @@
 package commonisolated
 
 import (
-	"fmt"
-	"testing"
-	"time"
+    "fmt"
+    "testing"
+    "time"
 
-	"code.cloudfoundry.org/cli/integration/helpers"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+    "code.cloudfoundry.org/cli/integration/helpers"
+    . "github.com/onsi/ginkgo"
+    . "github.com/onsi/gomega"
 )
 
 const (
-	CFEventuallyTimeout   = 300 * time.Second
-	CFConsistentlyTimeout = 500 * time.Millisecond
-	RealIsolationSegment  = "persistent_isolation_segment"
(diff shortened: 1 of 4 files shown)


jdgonzaleza:
I'm curious... didn't we have something for this in the integration tests setup? weren't we enabling some feature flags?

moleske:
> I'm curious... didn't we have something for this in the integration tests setup? weren't we enabling some feature flags?

The last commit here moves the enabling to a helper file which is also enabling the flags "diego_docker" and "service_instance_sharing" so adding "route_sharing" there seems to be consistent with what we've got

jdgonzaleza:
> > I'm curious... didn't we have something for this in the integration tests setup? weren't we enabling some feature flags?
> 
> The last commit here moves the enabling to a helper file which is also enabling the flags "diego_docker" and "service_instance_sharing" so adding "route_sharing" there seems to be consistent with what we've got

ohhhh Sorry!!! I didn't see which file was being modified!

moleske:
One concern I thought of is "route_sharing" is not available on all versions of capi.  So trying to enable this feature on a version of capi where "route_sharing" does not exist could fail badly.  The original location in each test had a check on the version of capi so it was only being enable on a version of capi that support "route_sharing"

PR #2332, opened 2022-10-13
Use default of 1GB memory instead of 32M

moleske:
- a regular deploy of cf-deployment has a default memory limit of 1G
- if we prefer to use 32M, we should change the tests to push an app specifying that as the limit instead

*I checked the "allow edits and access to secrets by maintainers, so any approver should be able to push commits if they see small issues*

Modifies tests to make github actions integration test work better, so should be backported to all versions still under testing

## Description of the Change
Use default of 1GB memory instead of 32M

## Why Is This PR Valuable?
Will allow folks to just run the tests instead of configuring their cloud foundry install

n/a

## Applicable Issues
#2326 

## How Urgent Is The Change?
As urgent as we want integration tests to work on github actions

## Other Relevant Parties
@cloudfoundry/wg-app-runtime-interfaces-cli-approvers 
@cloudfoundry/wg-app-runtime-interfaces-cli-reviewers 

### DIFF
--- a/integration/v7/isolated/scale_command_test.go
+++ b/integration/v7/isolated/scale_command_test.go
@@ -104,7 +104,7 @@ var _ = Describe("scale command", func() {
 				helpers.WithProcfileApp(func(appDir string) {
 					Eventually(helpers.CustomCF(helpers.CFEnv{WorkingDirectory: appDir}, "push", appName)).Should(Exit(0))
 				})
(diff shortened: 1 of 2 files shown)


PeteLevineA (approved):

jdgonzaleza (approved):
LGTM

PR #2347, opened 2022-11-07
update routes command to not require a space with --org-level argument

### LINKED ISSUE #2346: cf routes --org-level incorrectly requires a space to be targeted

- [x] I am reporting a bug that others will be able to reproduce.

When running `cf --org-level` with an organization targeted but not a space, I get this error:

```
No space targeted, use 'cf target -s SPACE' to target a space
```

**Expected behavior**

Since `cf --org-level` lists routes at the org level, I expect that the command should work without a space targeted.

**Exact Steps To Reproduce**

1. Target an organization, but **not a space**
2. Run `cf routes --org-level`
3. See error: 

```
No space targeted, use 'cf target -s SPACE' to target a space
```

**Investigation**

I think this is occurring because of this line in the code: https://github.com/cloudfoundry/cli/blob/v8/command/v7/routes_command.go#L28. `CheckTarget` is being called with a value of `true` for `targetedSpaceRequired`, [which causes `CheckTarget` to throw an error if a space is not targeted](https://github.com/cloudfoundry/cli/blob/v8/actor/sharedaction/check_target.go#L22). 

However, if `cmd.OrgLevel` is `true`, then we know that we're requesting routes at the org level and not space level, so I think we should pass `false` for `targetedSpaceRequired` to `CheckTarget`. 

I can work on a PR to fix this issue.

**Provide more context**

- Mac OS X, iTerm2
- CLI 8.5
- version of the CC API Release you are on

**Notes regarding V6 and V7 CLI support:** 
- V6:
  - Minimum supported version of CF Deployment: v7.0.0 (CAPI Release: 1.74.0 (APIs 2.128.0 and 3.63.0))
  - Maximum supported version of CF Deployment: v13.4.0 (CAPI Release: 1.94.0 (APIs 2.149.0 and 3.84.0))
- V7:
  - Minimum supported version of CF Deployment: v13.5.0 (CAPI Release: 1.95.0 (APIs 2.150.0 and 3.85.0))

markdboyd:

This PR modifies CLI v8.

## Description of the Change

As described in #2346, the CLI currently returns an error when running `cf routes --org-level` if a space is not targeted. However, this behavior is not desirable since listing routes at the organization level shouldn't require targeting a space. This PR fixes the behavior to remove the space requirement when `--org-level` is specified.

## Why Is This PR Valuable?

This PR fixes a minor bug affecting all users of CLI v8.

This PR is fixing a bug in the functionality that is already a part of the core

## Applicable Issues

Closes #2346 

## How Urgent Is The Change?

Not very urgent

## Other Relevant Parties

All users of CLI v8 are affected by the bug and thus would be affected by the fix.

### DIFF
--- a/cf/commands/route/routes.go
+++ b/cf/commands/route/routes.go
@@ -53,7 +53,11 @@ func (cmd *ListRoutes) Requirements(requirementsFactory requirements.Factory, fc
 	reqs := []requirements.Requirement{
 		usageReq,
 		requirementsFactory.NewLoginRequirement(),
-		requirementsFactory.NewTargetedSpaceRequirement(),
+	}
+
+	orglevel := fc.Bool("orglevel")
+	if !orglevel {
+		reqs = append(reqs, requirementsFactory.NewTargetedSpaceRequirement())
 	}
 
 	return reqs, nil
--- a/cf/commands/route/routes_test.go
+++ b/cf/commands/route/routes_test.go
@@ -64,6 +64,12 @@ var _ = Describe("routes command", func() {
 			Expect(runCommand()).To(BeFalse())
 		})
 
+		It("does not fail when an org is not targeted and --orglevel is an argument", func() {
+			requirementsFactory.NewTargetedSpaceRequirementReturns(requirements.Failing{Message: "not logged in"})
+
+			Expect(runCommand("--orglevel")).To(BeTrue())
+		})
+
 		Context("when arguments are provided", func() {
 			var cmd commandregistry.Command
 			var flagContext flags.FlagContext


FloThinksPi (approved):
Dont have write permissions on the CLI repo but i think this is a nice improvement and also compatible since the behaviour is extended and in case on has an org and space targeted the behaviour stays the same. Thanks for the PR 👍🏼 

The CLA must still be signed though :)

a-b (approved):
LGTM

a-b:
@markdboyd can you please make a PR for the main branch?

markdboyd:
@a-b Sure, heres a PR to `main`: https://github.com/cloudfoundry/cli/pull/2358

PR #2370, opened 2023-01-26
Replace 'set-output' deprecated command in GHA

ccjaimes:

No, just GHA

## Description of the Change

Replaces a deprecated command used to set environment variables used in GHA workflows, now uses a native concat to insert these variables.

## Why Is This PR Valuable?

The command will stop working soon, these changes will prevent our workflows to break.


a-b:
This change got to the main via the different PR. Clothing this one

PR #2380, opened 2023-03-01
Support `cf login -u` when targeting Korifi / Kubernetes

matt-royal:
The PR modifies v8

## Description of the Change
`cf login -u` now works when authenticating against Kubernetes. It will warn if a password is provided. We added the same warning to `cf auth`

## Why Is This PR Valuable?
It furthers support for Korifi by supporting more workflows

To further support integration with Korifi

## Applicable Issues
https://github.com/cloudfoundry/korifi/issues/2195

## How Urgent Is The Change?
This change isn't urgent but merging sooner will avoid us needing to rebasing later

## Other Relevant Parties
We aren't aware of any


gururajsh:
Looks good to me. @a-b What do you think?

a-b (approved):
LGTM

a-b:
We will observe the integration test run in V8 branch, since running tests off PR branches is now broken.

PR #2400, opened 2023-04-11
fix: hide password for create-service-broker (main)

blgm:
## Does this PR modify CLI v6, CLI v7, or CLI v8?
- v8

Please see the contribution doc above or review [Architecture Guide](https://github.com/cloudfoundry/cli/wiki/Architecture-Guide).

## Description of the Change
The `cf create-service-broker` command did not allow the password to be specified as an environment variable or via a prompt, meaning that users with certain logging options ended up with a password in the log. This fix brings the `cf create-service-broker` and `cf update-service-broker` into line with `cf auth` and `cf login` which allow credentials to be specified via environment variables and interactive prompts.

## Why Is This PR Valuable?

It fixes a security issue that users have raised

## Why Should This Be In Core?

Because it's a fix to an existing core command

## Applicable Issues

List any applicable GitHub Issues here

## How Urgent Is The Change?

Is the change urgent? If so, explain why it is time-sensitive.

## Other Relevant Parties

Who else is affected by the change?

PR #2402, opened 2023-04-12
feat: --update-if-exists flag for create-service-broker (main)

blgm:
- v8

## Description of the Change

A very common pattern is to have scripting like:
```
cf create-service-broker $SERVICE_BROKER_NAME $USERNAME $PASSWORD $APP_URL || cf update-service-broker $SERVICE_BROKER_NAME $USERNAME $PASSWORD $APP_URL
```

This is not ideal because:
- it's a duplicated pattern that the CF CLI could handle better
- it does not check the error type
- it results in a failure being logged, when in fact there was none

Adding a --update-if-exists flag to the CF CLI allows the following
alternative:
```
cf create-service-broker $SERVICE_BROKER_NAME $USERNAME $PASSWORD $APP_URL --update-if-exists
```

## Why Is This PR Valuable?

It gives script users a better and safer experience using the CF CLI

It's a change to a core command

## Applicable Issues

None

## How Urgent Is The Change?

It's likely that a lot of create-broker scripts will need to be updated for password hiding enabled in #2400. Although not strictly urgent, it makes sense for this to be in the same release.

## Other Relevant Parties

None

### DIFF
--- a/command/v7/create_service_broker_command.go
+++ b/command/v7/create_service_broker_command.go
@@ -3,6 +3,7 @@ package v7
 import (
 	"os"
 
+	"code.cloudfoundry.org/cli/actor/actionerror"
 	"code.cloudfoundry.org/cli/command"
 	"code.cloudfoundry.org/cli/command/flag"
 	"code.cloudfoundry.org/cli/resources"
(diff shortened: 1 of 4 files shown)


gururajsh (approved):

gururajsh (changes requested):
Need to resolve conflict

PR #2414, opened 2023-05-03
Invoke `POST /v3/apps/app-guid/actions/restart` on `cf restart app-name`

georgethebeatle:
Previously, `cf restart` would synchronously stop and then start the app.
Although this behaviour is documented by the CF V3 API, the cli could use the
dedicated `restart` endpoint. Thus API implementations are able to figure out
how to properly implement a restart vs stop+start

For example, the CF on K8S implementation - Korifi - could benefit from such a
change to implement app restart by leveraging the native K8S behaviour of a
rolling retart, i.e. an app restart (and also restage, or re-push) could have
no app downtime

With this draft PR we would like to check whether you are willing to accept
such a PR. If so, we could polish it - small refactoring, introducing tests,
etc.


georgethebeatle:
After further discussion in the team we decided to try implementing `/v3/deployments` part of the cf api in korifi instead of changing the cf cli. This way we would be more compliant to the v3 api and users would be able to use the --strategy rolling to achieve no downtime. Closing this PR.

PR #2491, opened 2023-07-20
Bump go dependencies to fix security vulnerability

ccjaimes:
Updating versions of net, text and crypto as updates included in go 1.20.6. Additionally, a file race condition fix has been implemented for Command Parser

Co-authored-by: Shwetha Gururaj <gururajsh@vmware.com>


go.mod
a-b, line 1: I suspect we might also want to change golang version [here](https://github.com/ccjaimes/cli/blob/1e435370acd4cd5f14f9b9ba683a41e8437a7129/go.mod#L3)?

a-b (changes requested):

a-b:
windows-latest units [looks broken](https://github.com/cloudfoundry/cli/actions/runs/5615063985/job/15214525479?pr=2491#step:6:343) - can we fix them?

a-b (approved):
LGTM

util/command_parser/command_parser_test.go
```diff
@@ -1,25 +1,24 @@
...
 var _ = Describe("Command 'Parser'", func() {
 	var (
 		pluginUI *ui.UI
+		v3Config *configv3.Config
 	)
 	BeforeEach(func() {
 		var err error
-		fakeConfig := new(commandfakes.FakeConfig)
-		pluginUI, err = ui.NewPluginUI(fakeConfig, ioutil.Discard, ioutil.Discard)
+		v3Config = new(configv3.Config)
```
a-b, line 20: Do we need a real `configv3` instead of simple fake in the command parser test?

ccjaimes: Given that CommandParsers was pulling a v3Config everytime it was created in every execution of BeforeEach, it could be more efficient to create a v3Config separately, pass it as parameter instead and use it in PluginUi as well

gururajsh (commented):
LGTM

Issue #2504, opened 2023-08-18
V3 view field Suspended for orgs


fredga:
The V3 API allows you to see the status of organizations. make this field visible in order to make it available for other developments that use the cloudfoundry go cli library.

Exemple, the community cf_exporter release uses the go library from the cloudfoundry cli in order to retrieve the metrics on the organizations. Currently only the name, guid and quota metrics are reported.
Would it be possible to recover the status of the organizations.

This feature is necessary for applications that use the cli product in go development such as cf_exporter in order to have additional metrics on the org.

This feature don't cause a breaking change.  Users will not be affected by this change

Make the Suspended value returned by the V3 cloudfoundry API available to orgs metrics.

![Capture d’écran du 2023-08-18 16-56-17](https://github.com/cloudfoundry/cli/assets/19749556/d50a03bd-3ecc-419c-8204-1ef78fb4f490)

joaopapereira:
It looks like there is already a PR that fixes this issue. I am going to close it, let me know if there is any work that is still needed.

PR #2511, opened 2023-08-24
add field suspended for orgs and tests

fredga:

- [x] [v7](https://github.com/cloudfoundry/cli/tree/v7)

## Description of the Change

The V3 API allows you to see the status of organizations. make this field visible in order to make it available for other developments that use the cloudfoundry go cli library.

Exemple, the community cf_exporter release uses the go library from the cloudfoundry cli in order to retrieve the metrics on the organizations. Currently only the name, guid and quota metrics are reported.
Would it be possible to recover the status of the organizations.

This feature is necessary for applications that use the cli product in go development such as cf_exporter in order to have additional metrics on the org.

This feature don't cause a breaking change. Users will not be affected by this change

Make the Suspended value returned by the V3 cloudfoundry API available to orgs metrics.

All units-full are SUCCESS.

## Applicable Issues

[issue 2504](https://github.com/cloudfoundry/cli/issues/2504)

### DIFF
--- a/api/cloudcontroller/ccv3/organization_test.go
+++ b/api/cloudcontroller/ccv3/organization_test.go
@@ -450,6 +450,7 @@ var _ = Describe("Organizations", func() {
 
 				expectedBody := map[string]interface{}{
 					"name": "some-org-name",
+					"suspended": false,
 				}
 
 				server.AppendHandlers(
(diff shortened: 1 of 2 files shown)


a-b (approved):
LGTM

a-b:
@fredga If this change is expected in v8, please create a new PR.

PR #2560, opened 2023-09-07
Included GHA package ecosystem for dependabot in v7 and v8

ccjaimes:
This change has the purpose of keeping all pipeline configuration synced and up to date across main branches (v7 and v8)


moleske (approved):
This will help keep things in sync across the three branches in the future as they all should be using the same versions of actions

PR #2573, opened 2023-09-11
GHA: Workflow to bump GPG key expiration date

a-b:
This pipeline creates an ephemeral environment loaded with a signing key, so the operator can perform manipulations.


moleske (approved):
Seems reasonable

PR #2359, opened 2022-12-08
Add --wait, -w flag to run-task

stephanme:
When the wait flag is supplied to run-task, the cli will not exit until the task enters a terminal state; SUCCEEDED or FAILED.

The cli exits 0 when the task succeeds and exits 1 when the task fails.

This improves the scriptability of tasks. They can be run serially in a script.

Cherry-pick of 8671a5b (#1861) to v8
Addresses Issue #1104, #2238

Authored-by: Zach Robinson <zrobinson@pivotal.io>
Co-authored-by: Stephan Merker <stephan.merker@sap.com>

## Does this PR modify CLI v6, CLI v7, or CLI v8?

CLI v8

## Description of the Change

Adds --wait flag to run-task command as available in CLI v7.

## Why Is This PR Valuable?

This improves the scriptability of tasks. They can be run serially in a script.
The --wait flag is supported for the run-task command in v7.

## Why Should This Be In Core?

New cmd line flag for existing core command.

## Applicable Issues

#2238

Original issue for v7 was #1104

## How Urgent Is The Change?

not urgent

## Other Relevant Parties

n/a

Issue #2238, opened 2021-12-08
run-task `--wait` parameter is missing from cf8


mogul:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
Provide details on what you were trying to do (and why).

I was trying to run `cf run-task [PARAMS] --wait` to wait for the output before proceeding in a script.

**What happened**
`Incorrect Usage: unknown flag '--wait'`

**Expected behavior**
The command does not error out, and does not return until the task succeeds or fails.

**Exact Steps To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. `brew install cloudfoundry/tap/cf-cli@7`
2. `cf run-task --wait` (errors out because of missing parameters, but no complaints about `--wait`)
4. `brew uninstall cloudfoundry/tap/cf-cli@7`
1. `brew install cloudfoundry/tap/cf-cli@8`
2. `cf run-task --wait` (this is not sufficient parameters to actually run a task, but it demonstrates that `--wait` no longer works in cf8) 
3. See error about the `--wait` parameter
 
**Provide more context**
- platform and shell details ( e.g. Mac OS X 10.11 iTerm)
  - This was in Ubuntu Linux, using the cf-cli 8 installed via Homebrew (which now supports Linux)
- version of the CLI you are running
  - CF7: `cf version 7.4.0+e55633fed.2021-11-15`
  - CF8: `cf version 8.0.0+e8d41cf8e.2021-09-16`
- version of the CC API Release you are on
  - N/A

See [the Slack thread that precipitated this issue](https://cloudfoundry.slack.com/archives/C032824SM/p1638818105015600) for additional info; note that the --help output and CLI docs site for CF8 will need updating when this fixed as well.

mogul:
This is causing problems between teams who share tooling that uses the `cf` CLI... People keep wanting to iterate the tooling to `cf8` when certain tasks are only possible with `cf7`. Any chance it can be addressed soon?

stephanme:
+1

[...]

PR #1861, opened 2019-12-06
Add --wait, -w flag to run-task

zrob:
When the wait flag is supplied to run-task, the cli will
not exit until the task enters a terminal state; SUCCEEDED or FAILED.

The cli exits 0 when the task succeeds and exits 1 when the task fails.

This improves the scriptability of tasks. They can be run
serially in a script.

Addresses Issue #1104

---

This affects v7 only.

Does not address the --log request in the associated issue.

Issue #1104, opened 2017-03-27
Request: cf run-task --wait


gajwani:
## Request:
AS a user of the CF CLI,
GIVEN that I have run a task
THEN the CLI waits for the task to complete

## Proposed API
`cf run-task <APP_NAME> <COMMAND> --wait`

Since the CLI already polls in other cases, and also is able to talk to the API to get the app/task status, it seems reasonable that polling for it's status and telling a user about completeness would be desirable.

/cc @tjgrathwell

gajwani:
@mpivotal We were told you might also have some thoughts about this.

dkoper:
We considered this when exposing tasks and decided to wait and let people try it out as is to see what direction to take with this feature. So thanks for submitting the issue!

I understand it's not uncommon for tasks to take a long time (their default timeout I believe is something like 48hrs for that reason), so an alternative considered was to "attach" the terminal to the task, like an ssh session as Heroku does. That requires feature work on Diego and CC, so curious to know whether @zrob and @ematpl already have something on their roadmap.

tjgrathwell:
Attaching to the task would be ideal if streamed the output as it happened.

In our polling implementation, we do a `cf logs --recent` after the task enters `FINISHED` state and print out the lines that look like they match our task, but we're at the whim of however well `cf logs --recent` feels like treating us. Sometimes that command has even crashed with a mysterious cloud controller error.

But if it takes a significant amount of extra work to do the attach thing, I wouldn't want it hold up a simpler waiting technology. The lack of a wait makes it hard to do simple things like run a series of tasks one after another.

I don't know how much it makes sense to optimize for extremely long running tasks, if I were to do something long running in my application I would probably only use `cf run-task` to feed something into a worker queue.

aeijdenberg:
In case it's helpful, we wrote an experimental CLI plug-in called `run-and-wait` to do some of this, ie start the task, poll it, exit with 0 or 1 depending on success or failure. We're beginning to use this in some of our CI tasks:

<https://github.com/govau/cf-run-and-wait>

oppegard:
We (Tracker team) would appreciate something like this as well for use in automation/CI scenarios. It'd be even better if there were a `--logs` option in addition to `--wait`, so that logs would be streamed to stdout until the task either succeeds or fails.

kelapure:
This story is critical and needed; in the interim we wrote a workaround for it here https://gist.github.com/kelapure/ef79419796e35a68629a4e772e4646e4

[...]

PR #2611, opened 2023-10-12
Add --wait, -w flag to run-task 

ccjaimes:
When the wait flag is supplied to run-task, the cli will not exit until the task enters a terminal state; SUCCEEDED or FAILED.

The cli exits 0 when the task succeeds and exits 1 when the task fails.

This improves the scriptability of tasks. They can be run serially in a script.

Cherry-pick of 8671a5b (#1861) to v8
Addresses Issue #1104, #2238

Authored-by: Zach Robinson <zrobinson@pivotal.io>

Taking the fix from v8 #2359

### DIFF
--- a/actor/actionerror/task_failed_error.go
+++ b/actor/actionerror/task_failed_error.go
@@ -1,0 +1,7 @@
+package actionerror
+
+type TaskFailedError struct{}
+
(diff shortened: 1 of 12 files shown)


moleske (approved):
yay forward ports!

PR #2617, opened 2023-10-17
Don't fetch unneeded pages

### LINKED ISSUE #2599: Don't fetch unneeded pages
While looking through some logs, I've stumbled across the following request to Cloud Controller:
```
/v3/packages?app_guids=<some_guid>&order_by=-created_at&page=900&per_page=1
```

This request was sent by the `cf` cli, so I searched for its origin. Inside the `GetUnstagedNewestPackageGUID` function, [all packages](https://github.com/cloudfoundry/cli/blob/d83fe8aef8302e209366d20f7fdcefb3d44102a1/actor/v7action/application.go#L226-L238) for an app are fetched (with page size 1), but only two things are done with the result: (a) there is a check if the returned list of packages is empty, (b) the first item from the list is taken.

The underlying problem is that many of the `CloudControllerClient` functions (like `GetPackages` in this case) use the [`MakeListRequest`](https://github.com/cloudfoundry/cli/blob/d83fe8aef8302e209366d20f7fdcefb3d44102a1/api/cloudcontroller/ccv3/package.go#L51) function which automatically does [pagination](https://github.com/cloudfoundry/cli/blob/d83fe8aef8302e209366d20f7fdcefb3d44102a1/api/cloudcontroller/ccv3/requester.go#L98) until [no further page exists](https://github.com/cloudfoundry/cli/blob/d83fe8aef8302e209366d20f7fdcefb3d44102a1/api/cloudcontroller/ccv3/paginate.go#L29-L31).

This behavior is fine as long as the returned list is needed, but there seem to be several cases where actually only the first item in a list (i.e. the newest one) is used.

So I decided to start working on a change that makes use of the [`page`](https://v3-apidocs.cloudfoundry.org/version/3.148.0/index.html#pagination) parameter. Until now I did not open a PR, but just pushed the changes to a fork. My idea is to introduce the [`Page` `QueryKey`](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-c1850802d9c9a4e69ee2d48382b5488b46f1cc22b873d93a1252e671399693aaR98) and (besides forwarding it to the Cloud Controller), [use it](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-64f90ef68b9a423698ad3c4e026be95a4d0283802be528acb45406927d68fb9aR98-R106) to [stop](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-6df3e82d49aaedfeac073b3c68a67464c7faa2ad01c2dafc596bb243b7791135R29) the automatic pagination in the cli code.

This means that the call to `GetPackages` can simply be changed to [include this `QueryKey`](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-0659c25c0870d191835de8179e4be5b8591aa5fd97e0849c1d9a5d9fcd03727fR230).

I then started to look through the code to identify more places where (a) a list endpoint is called and (b) only the first list item is used (or no item at all, but only a check if the list was empty is done). My idea is to use `PerPage` and `Page` consistently in all these cases even if the query ensures that only a single item matches.

I also found the `GetEvents` function that [wasn't using the `MakeListRequest` function](https://github.com/cloudfoundry/cli/blob/d83fe8aef8302e209366d20f7fdcefb3d44102a1/api/cloudcontroller/ccv3/event.go#L43-L45) internally as the only caller just wanted the first page to be returned. I've [corrected this](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-feefd45cc0268f42f7cf4bedeb084839f2201d44cb033ea7b09bb34bce6442c1R46) and simply [added the `Page` `QueryKey` to the caller](https://github.com/sap-contributions/cli/commit/9adea4c4fffff9975f3815d2a5fbf5324278339e#diff-ba77be84832cef50274e6270942e650e7c1307a760c2bc16c39e11fe359a5908R33).

Before continuing my work (there are still other places where the `Page` `QueryKey` can be added) and opening a PR, I would like to get some feedback if I'm heading in the right direction with this change.

philippthun:
- introduce `Page` query key
- when set, stop automatic pagination
- add `PerPage` and `Page` query keys to all list endpoints where only a single item from the returned list is used
- use `MakeListRequest` function in `GetEvents` and set `Page` query key in caller
- adapt `StagePackage` function to include `GUIDFilter` instead of iterating over the returned package list and comparing guids

Fixes #2599

### DIFF
--- a/actor/v7action/application.go
+++ b/actor/v7action/application.go
@@ -226,7 +226,8 @@ func (actor Actor) GetUnstagedNewestPackageGUID(appGUID string) (string, Warning
 	packages, warnings, err := actor.CloudControllerClient.GetPackages(
 		ccv3.Query{Key: ccv3.AppGUIDFilter, Values: []string{appGUID}},
 		ccv3.Query{Key: ccv3.OrderBy, Values: []string{ccv3.CreatedAtDescendingOrder}},
-		ccv3.Query{Key: ccv3.PerPage, Values: []string{"1"}})
+		ccv3.Query{Key: ccv3.PerPage, Values: []string{"1"}},
+		ccv3.Query{Key: ccv3.Page, Values: []string{"1"}})
 	allWarnings = append(allWarnings, warnings...)
 	if err != nil {
 		return "", allWarnings, err
@@ -241,6 +242,7 @@ func (actor Actor) GetUnstagedNewestPackageGUID(appGUID string) (string, Warning
 		newestPackage.GUID,
 		ccv3.Query{Key: ccv3.StatesFilter, Values: []string{"STAGED"}},
 		ccv3.Query{Key: ccv3.PerPage, Values: []string{"1"}},
+		ccv3.Query{Key: ccv3.Page, Values: []string{"1"}},
 	)
 	allWarnings = append(allWarnings, warnings...)
 	if err != nil {
--- a/actor/v7action/application_test.go
+++ b/actor/v7action/application_test.go
@@ -1959,6 +1959,7 @@ var _ = Describe("Application Actions", func() {
 						ccv3.Query{Key: ccv3.AppGUIDFilter, Values: []string{"some-app-guid"}},
 						ccv3.Query{Key: ccv3.OrderBy, Values: []string{ccv3.CreatedAtDescendingOrder}},
 						ccv3.Query{Key: ccv3.PerPage, Values: []string{"1"}},
+						ccv3.Query{Key: ccv3.Page, Values: []string{"1"}},
 					))
 				})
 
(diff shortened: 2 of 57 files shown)


ccjaimes:
Hello @philippthun Could you please create an additional PR with these changes for v8? We are looking to merge this contribution for next release!

a-b (approved):
@dalvarado and I reviewed V8 version of this proposal and it looks good to us.

PR #2665, opened 2023-11-13
Change status query into a information message when finishing a rolling restage - v7

ccjaimes:
This change will prevent the issue reported in Jira's ticket CLI-50 which creates a race condition when trying to query the app's rolling restage status at the same moment the old restage group gets deleted.


a-b (approved):
LGTM

PR #2666, opened 2023-11-13
Change status query into a information message when finishing a rolling restage - main

ccjaimes:
This change will prevent the issue reported in Jira's ticket CLI-50 which creates a race condition when trying to query the app's rolling restage status at the same moment the old restage group gets deleted.

Link for v8: https://github.com/cloudfoundry/cli/pull/2664
Link for v7: https://github.com/cloudfoundry/cli/pull/2665

### DIFF
--- a/command/v7/shared/app_stager.go
+++ b/command/v7/shared/app_stager.go
@@ -172,6 +172,10 @@ func (stager *Stager) StartApp(
 		if err != nil {
 			return err
 		}
+		if noWait == true {
+			stager.UI.DisplayText("First instance restaged correctly, restaging remaining in the background")
+			return nil
+		}
 	} else {
 		user, err := stager.Actor.GetCurrentUser()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


a-b (changes requested):
This change should be ok, please add tests.

pivotalgeorge:
as this change was made to resolve a tricky-to-reproduce issue, going to document the setup here:
- used a TAS 4.0 testing environment, set up such that the `compute` bosh job (which hosts [garden](https://github.com/cloudfoundry/garden-runc-release) and therefore app instances) was on a GCP VM with 32GB memory, 131GB root disk, and 2vCPUs
- used a version of the cf-cli built off the v8 branch - specifically c7d46b6d4
- 10 sample apps, each using the [dora](https://github.com/cloudfoundry/cf-acceptance-tests/tree/main/assets/dora) code but named `dora-1`, `dora-2`, etc.
- 3-at-a-time restaging of those apps, i.e. `cf restage "dora-$i" --strategy rolling --no-wait &` for `i := {1,2,3}`, followed by the next 3, etc.
- observed `Process not found  FAILED`, etc.
- rotating too many (all 10) apps at a time also resulted in `Process not found`, but this was a red herring as it was caused by `InsufficientResources` (out of memory)
- observed that the same testing procedure did **not** result in an error using a version of the binary built from this PR (d444bc11e)

PR #2763, opened 2024-01-30
add redact option for env block when pushing an app with a manifest.yml

nouseforaname:
IF the app is pushed with a manifest
    AND the app needs secrets set via environment variables
    THEN currently there is no way to avoid leaking secrets in the output of
    cf push commands.

    technically this could be avoided by not using a manifest and running:
    ```cf push --no-start
    cf set-env ...
    cf set-env ...
    cf set-env ...
    ...
    `

    but that escalates quickly if the APP contains many secret things in it's
    env requirements.

    We face this in an errand where:
    - the bosh release renders the application.yml
    - the application.yml contains secrets in the env block
    - the errand VM streams it's logs to a logging server

    This can be remediated by the suggested redacttion.
    `--redact-env` to indiscriminately change all values
    to `<redacted>` if they're contained within an apps `env` block.


a-b (approved):
Thanks for your contribution. LGTM!

a-b:
@nouseforaname, If you want to see this change in the next V8 release, please create a pull request for the respective branch.

integration/v7/push/manifest_diff_test.go
```diff
@@ -53,4 +53,43 @@ var _ = Describe("displaying manifest differences between pushes", func() {
 			})
 		})
 	})
+	FWhen("there is an app that has been pushed and we push it again with a different manifest", func() {
```
a-b, line 56: @nouseforaname, can you please remove the focus?

integration/v7/push/manifest_diff_test.go
```diff
@@ -53,4 +53,43 @@ var _ = Describe("displaying manifest differences between pushes", func() {
...
+		FIt("displays the diff from the manifest", func() {
```
a-b, line 57: Same focus

a-b (changes requested):
Please remove the focus on tests.

Also we should ensure the linter catches issues earlier.

cc @moleske

nouseforaname:
> @nouseforaname, If you want to see this change in the next V8 release, please create a pull request for the respective branch.

That is actually an interesting point. Should this go anywhere else than v8?

integration/v7/push/manifest_diff_test.go
nouseforaname: Done, thanks for spotting this.

integration/v7/push/manifest_diff_test.go
nouseforaname: Done, thanks for spotting this.

moleske:
> That is actually an interesting point. Should this go anywhere else than v8?

v7 is the only other place for this to potentially go.  So if needed on the v7 cli than a pr to the v7 branch would be needed

a-b (approved):
No more focused tests. LGTM!

PR #2783, opened 2024-02-12
Docs: Remove deprecated V6 mentions

a-b:


moleske (approved):
so long v6, maybe some day we'll drop you out of the [bosh cf cli release](https://github.com/cloudfoundry/bosh-package-cf-cli-release/pull/11)

PR #2787, opened 2024-02-14
V8 add redact option for env block when pushing an app with a manifest.yml 

nouseforaname:
V8 PR for
https://github.com/cloudfoundry/cli/pull/2763


moleske:
as #2763 was approved and merged, merging this for v8

PR #2793, opened 2024-02-15
Use `max_page` size for apps, service keys & routes

johha:

- [x] [v8](https://github.com/cloudfoundry/cli/tree/v8) -> #2794 

## Description of the Change

To reduce the number of db queries in cloud controller it is beneficial to use a larger page size (5k) for potentially large entities.
This was already done for `marketplace` and `services` in commit b823ae53f394acf64ad7fb09f550c508a3c46e0a. This commit does the same for `apps`, `routes` and `service-keys`

### DIFF
--- a/actor/v7action/application_summary.go
+++ b/actor/v7action/application_summary.go
@@ -40,6 +40,7 @@ func (actor Actor) GetAppSummariesForSpace(spaceGUID string, labelSelector strin
 	keys := []ccv3.Query{
 		{Key: ccv3.SpaceGUIDFilter, Values: []string{spaceGUID}},
 		{Key: ccv3.OrderBy, Values: []string{ccv3.NameOrder}},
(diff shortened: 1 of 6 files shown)


moleske:
Hi @johha , since you've indicated this should go to the v8 and v7 releases, can you also open PRs against the v8 and v7 branches as well?

johha:
> Hi @johha , since you've indicated this should go to the v8 and v7 releases, can you also open PRs against the v8 and v7 branches as well?
V8 is sufficient actually. I opened a PR: #2794

moleske (approved):
This seems reasonable to me, especially as it will reduce stress on cloud controller

chombium (approved):
Cool optimization @johha

johha:
@moleske @a-b Can we merge this PR?

PR #2794, opened 2024-02-16
Use max_page size for apps, service keys & routes [V8]

johha:
See #2793


moleske:
Main branch pr merged, so doing this one

Issue #2734, opened 2024-01-11
'cf a' should print out available information even if an app process was not found


stephanme:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**

`cf a` in a space that contains ~800 apps takes ~5min (see also #2733 for one reason). When an app in the space is deleted e.g. by CI while `cf a` is running, cli fails and prints out nothing. 

**What happened**

```
% cf a
Getting apps in org test / space test as stephan.merker@sap.com...

Process not found
FAILED
```

The 'process not found' error result most likely from `GET /v3/processes/<process guid>/stats` requests.

**Expected behavior**

`cf a` should print out available info for better UX, especially when it takes long time in large spaces:

```
% cf a
Getting apps in org test / space test as stephan.merker@sap.com...

name               requested state   processes                          routes
app1               started           web:1/1                            app1.cfapps.eu12.hana.ondemand.com
app2               started           not found
...
```
(when app2 got deleted during execution of `cf a`)

Since multiple CF API calls are involved in `cf a`, cli can't guarantee a consistent output.

**Exact Steps To Reproduce**

- push several hundred apps in a space
- run `cf a`
- while `cf a` is running, delete one of the last apps (sorted by app name)

**Provide more context**
- platform and shell details:  Mac OS
- version of the CLI you are running: 8.7.6+a659013.2023-12-14
- version of the CC API Release you are on: 3.153.0

**Notes regarding V6 and V7 CLI support:** 
- V6:
  - Minimum supported version of CF Deployment: v7.0.0 (CAPI Release: 1.74.0 (APIs 2.128.0 and 3.63.0))
  - Maximum supported version of CF Deployment: v13.4.0 (CAPI Release: 1.94.0 (APIs 2.149.0 and 3.84.0))
- V7:
  - Minimum supported version of CF Deployment: v13.5.0 (CAPI Release: 1.95.0 (APIs 2.150.0 and 3.85.0))

johha:
Fixed with #2795

PR #2795, opened 2024-02-19
Improve `cf apps` with `--no-stats` and better error handling

johha:

- [x] [v8](https://github.com/cloudfoundry/cli/tree/v8) -> #2796

### Introduce `--no-stats` for `cf apps`

Fetching process stats requires a `/stats` request for each process.
If a space contains a large number apps or processes this slows down `cf apps`.
In case a user is not interested in stats the `--no-stats` flag can be helpful.

### Handle `ProcessNotFoundError` in `cf apps`

As a user calls `cf apps` in a large space it might happen that a app
was deleted while the CLI is putting all the different information (processes, stats & routes) together.
This leads to a failing `cf apps` command.
This change prevents this by catching `ProcessNotFoundError` errors and returning `nil` instead.
Related to #2734

### DIFF
--- a/actor/v7action/application_summary.go
+++ b/actor/v7action/application_summary.go
@@ -32,7 +32,7 @@ func (a ApplicationSummary) hasIsolationSegment() bool {
 		len(a.ProcessSummaries[0].InstanceDetails[0].IsolationSegment) > 0
 }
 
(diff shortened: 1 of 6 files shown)


a-b (approved):

a-b:
Thank you @johha  for your contribution. We appreciate the quality of your commit and PR description.

Issue #2812, opened 2024-03-15
Output CPU Entitlement metrics rather than the old CPU metric


acrmp:
**What's the user value of this feature request?**

Currently the cf CLI displays a CPU metric which is difficult to reason about. See #1194 for background here.

Instead we'd like the cf CLI to show CPU Entitlement usage which is more useful.

**Who is the functionality for?**

End users of the cf CLI.

**How often will this functionality be used by the user?**

Frequently. We would expect the CPU Entitlement usage stat to be presented when `cf app` is run or following `cf push` (as currently).

**Who else is affected by the change?**

This is potentially a breaking change dependent on whether this is implemented by replacing the existing CPU column or adding an additional column.

**Is your feature request related to a problem? Please describe.**

Developers find it difficult to reason about their application use of CPU without knowledge of the underlying platform.

**Describe the solution you'd like**

Modify the cf CLI to output CPU Entitlement in place of the old CPU metric for environments that expose the CPU Entitlement process stat.

```
     state     since                  cpu entitlement   memory        disk           logging       details
#0   running   2024-02-10T00:30:38Z   5.0%              47.7M of 1G   129.6M of 1G   0/s of 16K/s
#1   running   2024-02-12T18:17:49Z   5.0%              47.7M of 1G   129.6M of 1G   0/s of 16K/s
```

For environments that do not expose the CPU Entitlement information the cf CLI would continue to show the existing output.

**Describe alternatives you've considered**

Adding an additional column towards the end of the column output would be more backwards compatible but potentially confusing:

```
     state     since                  cpu    memory        disk           logging        cpu entitlement   details
#0   running   2024-02-10T00:30:38Z   0.2%   47.7M of 1G   129.6M of 1G   0/s of 16K/s   5.0%
#1   running   2024-02-12T18:17:49Z   0.1%   47.3M of 1G   129.6M of 1G   0/s of 16K/s   2.5%
```

**Additional context**

https://github.com/cloudfoundry/diego-release/issues/897
https://github.com/cloudfoundry/cloud_controller_ng/pull/3641

There was some great work done to emit absolute usage and entitlement metrics earlier but these were not directly consumable via the core cf CLI.

https://github.com/cloudfoundry/cpu-entitlement-plugin

Benjamintf1:
If we go with the initial proposed option, I think we should consider giving the user a compatibility flag (IE. USE_CPU_USAGE=true) to use cpu usage regardless of whatever capi returns.

[...]

PR #2840, opened 2024-04-01
Feat: Add cpu entitlement to app process table

ctlong:

- [ ] [main](https://github.com/cloudfoundry/cli/tree/main) - all changes should by default start here

## Description of the Change

Extend the v8 cf CLI to support showing the new CPU Entitlement metric as a non-breaking change, with the `cpu entitlement` column appearing before the details column:

```
     state     since                  cpu    memory        disk           logging        cpu entitlement   details
#0   running   2024-02-10T00:30:38Z   0.2%   47.7M of 1G   129.6M of 1G   0/s of 16K/s   5.0%
#1   running   2024-02-12T18:17:49Z   0.1%   47.3M of 1G   129.6M of 1G   0/s of 16K/s   2.5%
```

If the cpu entitlement stats are not available (e.g. deployment does not support it, app is stopped, etc.) then the cf CLI should show an empty value for the process column.

## Why Is This PR Valuable?

Extend the v8 cf CLI to support showing the new CPU Entitlement metric, where applicable.

## Applicable Issues

- #2812 
- https://www.pivotaltracker.com/story/show/187310450

## How Urgent Is The Change?

Slightly less than urgent.

## Other Relevant Parties

@rroberts2222


ctlong:
We're not sure why this change is failing unit tests in macOS... it seems unrelated to our changes. When we run unit tests on the `main` branch on my local mac, we see these tests flake one out of every 7 runs or so.

ctlong:
Using the following command, I was able to run unit tests for this branch on my local mac 23 times in a row without failure:
```bash
CF_HOME=$PWD/fixtures CF_USERNAME="" CF_PASSWORD="" ginkgo -r -randomize-all -require-suite -randomize-suites -skip-package integration,cf/ssh,plugin,cf/actors/plugin,cf/commands/plugin,cf/actors/plugin,util/randomword -until-it-fails
```

moleske:
I like clicking rerun so I've unstuck the pr checks (for now)

Please also open a pr against the main branch as any changes in the v8 branch we'll want in main as well

ctlong:
We've another change in progress for `main` right now.

Where this change represents the "alternative option" described in https://github.com/cloudfoundry/cli/issues/2812, for cf CLI v9 we're aiming for the proposed solution in that issue, which we prefer 😄

a-b (approved):

dalvarado (approved):

Issue #2669, opened 2023-11-20
cf share-service fails with exit code 1 if it is already shared


svenXY:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**

With cf cli v8, the command 
`cf share-service srv-name -o org -s space` fails (FAILED) and exits with 1 if the service is already shared with the org/space

Until recently, it just reported that it is already shared (OK) and ended with exitcode 0.

Looking at other cli commands like e.g. cf create-service, I'm under he impression that for idempotence reasons, commands that would only try to produce a state that already exists should not fail.

**What happened**

Gitlab pipelines fail all of a sudden due to new version of cf cli tools

**Expected behavior**

Pipelines should continue to work. If failing is going to be the new behaviou (to which I'd highly disagree!), it should at least be mentioned in the documentation (that does not make any mention of exit codes at all). Furthermore, there is no possibility to find out if a sharing is already in place except by grepping the output of `cf service`...

**Exact Steps To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. Run `cf share-service srv-name -o org -s space` on a service that is already shared with the named org/space
2. See error

svenXY:
RESPONSE: [2023-11-20T15:53:35+01:00]
HTTP/1.1 422 Unprocessable Entity
Content-Type: application/json; charset=utf-8
Date: Mon, 20 Nov 2023 14:53:35 GMT
Referrer-Policy: strict-origin-when-cross-origin
Server: nginx
Vary: Accept
X-B3-Spanid: 533b6174dda84c69
X-B3-Traceid: 3642895499764d5d533b6174dda84c69
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
X-Runtime: 0.051301
X-Vcap-Request-Id: 36428954-9976-4d5d-533b-6174dda84c69::6c4c873b-d255-4110-a1e1-aa02b9c3c17d
X-Xss-Protection: 1; mode=block
{
  "errors": [
    {
      "code": 10008,
[...]

PR #2861, opened 2024-04-29
[Bugfix] Idempotency when trying to share a service twice

vchrisb:

## Description of the Change

While `cf create-service` has an Exitcode of `0` if a service with the name already exist, does `cf share-service` have an Exitcode of `1` for the similar situation.
This behavior is different to cf cli v7.

cli v7:
```
Sharing service instance foo into org test / space dev2 as admin...
Service instance foo is already shared with that space.
OK
```

cli v8:
```
Sharing service instance foo into org test / space dev2 as admin...
A service instance called foo has already been shared with dev2.
FAILED
```
 
The PR is changing the behavior of `cf8 share-service` to return an exit code of `0` if the service is already shared and make it consistent with `v7`.

## Why Is This PR Valuable?

Make `cf share-service` idempotent and consistent with v7

## Applicable Issues

#2669

## How Urgent Is The Change?

## Other Relevant Parties


gururajsh:
@vchrisb We had some environment issues and integration tests did not run. Can you please update the branch and reopen the PR? Also, can you please create a separate PR for main?

PR #2862, opened 2024-04-29
Use shepherd envs instead on toolsmiths for tests (main)

gururajsh:

- [] [v8](https://github.com/cloudfoundry/cli/tree/v8)
- [] [v7](https://github.com/cloudfoundry/cli/tree/v7)

## Description of the Change

- CLI integration tests used toolsmiths envs. This change is to use shepherd environments instead since toolsmiths envs are not longer supported.

- Min Capi job is disabled until we figure out CLI versioning policy (how far back we want to support cf-deployment version).

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

Its time-sensitive since we have not run tests/merge PRS/release CLI because of envs not being available. 

## Other Relevant Parties


.github/workflows/tests-integration-reusable.yml
```diff
@@ -80,38 +82,36 @@ jobs:
...
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 401" headers.txt && exit 1
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 404" headers.txt && exit 2
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 500" headers.txt && echo "$ERR_500"
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 200" headers.txt && break
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 429" && echo "$ERR_429"
-          grep -q -E "HTTP/[[:digit:]\.]{1,3} 409" && echo "$ERR_409"
+        shepherd login service-account ${account_token}
+        lease_id=$(shepherd create lease --duration 8h --pool ${pool_name} --pool-namespace official --namespace tas-devex --json | jq -r .id) 
 
+        # Give somtime for the lease to complete. Shepherd may take upto an 3 hours to create an env
```
moleske, line 94: oh boy for that wait time.  though I guess the other stalled if there were no environments anyway

.github/workflows/tests-integration.yml
```diff
@@ -33,16 +33,16 @@ jobs:
...
-    name: MIN CAPI
-    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
-    uses: ./.github/workflows/tests-integration-reusable.yml
-    with:
-      capi-version: min
-      run-with-client-creds: false
-      os: ubuntu-latest
-      name: Integration MIN CAPI
-    secrets: inherit
+  # run-integration-tests-cf-env-with-min-capi:
```
moleske, line 36: I'm assuming this is commented out until it is figured out what we actually want to test for min version?  Otherwise I'd delete it

moleske (approved):
I have no way of testing this, but I trust folks have tested this.  Plus no place like live in a pr to test 😄

.github/workflows/tests-integration.yml
gururajsh: Yes.. Once we figured out min version, this will come back.

.github/workflows/tests-integration-reusable.yml
gururajsh: Yup.. shepherd takes upto 3 hrs to create an env, so better to wait than come back

gururajsh:
> I have no way of testing this, but I trust folks have tested this. Plus no place like live in a pr to test 😄

I have tested this in my fork for all three branches. They passed 😄

.github/workflows/tests-integration.yml
a-b: Motivation is a bit different, the intention is to put them back as we figure out issues.

a-b (approved):

.github/workflows/tests-integration-reusable.yml
```diff
@@ -30,7 +29,7 @@ jobs:
       run:
         shell: bash
     runs-on: ${{ inputs.os }} 
-
+    container: us-west2-docker.pkg.dev/shepherd-268822/shepherd2/concourse-resource:latest
```
a-b, line 32: This seems to be a Google package repository. Can we rely on the stability of this URL, specifically the `268822`?

.github/workflows/tests-integration-reusable.yml
```diff
@@ -56,6 +56,8 @@ jobs:
     - id: read-min-capi
       name: Read MIN CAPI
       run: |
+        wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq &&\
```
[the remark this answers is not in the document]
gururajsh: Did not remove any code related to min capi job since it'll be enable again.

.github/workflows/tests-integration-reusable.yml
gururajsh: This is the work around for now to use shepherd envs in workflows

reedr3 (approved):
lgtm

PR #2863, opened 2024-04-29
Use shepherd envs instead on toolsmiths for tests (v7) 

gururajsh:

- [ ] [main](https://github.com/cloudfoundry/cli/tree/main) - all changes should by default start here
- [ ] [v8](https://github.com/cloudfoundry/cli/tree/v8)
- [x] [v7](https://github.com/cloudfoundry/cli/tree/v7)

## Description of the Change

- CLI integration tests used toolsmiths envs. This change is to use shepherd environments instead since toolsmiths envs are not longer supported.

- Min Capi job is disabled until we figure out CLI versioning policy (how far back we want to support cf-deployment version).

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

Its time-sensitive since we have not run tests/merge PRS/release CLI because of envs not being available.

## Other Relevant Parties

### DIFF
--- a/.github/workflows/release-build-sign-upload.yml
+++ b/.github/workflows/release-build-sign-upload.yml
@@ -78,7 +78,7 @@ jobs:
 
   setup:
     name: Setup
-    # needs: test
+    if: github.triggering_actor.role == 'Write'
     runs-on: ubuntu-latest
 
     outputs:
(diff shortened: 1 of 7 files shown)


reedr3 (dismissed):
lgtm

reedr3 (approved):

PR #2880, opened 2024-05-06
Release workflow improvements [v7]

gururajsh:
## Description of the Change

This change is to update our release process. It also

Uploads artifacts to github release
Adds release notes template
Creates release notes for github release
Updates CLAW with new release version

## Why Is This PR Valuable?

It improves time needed to do a CLI release.

## Applicable Issues

None

## How Urgent Is The Change?

## Other Relevant Parties


.github/workflows/release-build-sign-upload.yml
```diff
@@ -1033,4 +1006,88 @@ jobs:
...
+        ref: develop
+        path: CLAW
+        ssh-key: ${{ secrets.GIT_SSH_KEY_CLAW }}  
+
+    - name: Add new version to claw variables
+      run: |
+        set -ex
+
+        pushd CLAW
+          echo "- ${VERSION_BUILD}" >> claw-variables.yml
```
a-b, line 1079: Do we have an understanding of how CLAW will react if we add the same version multiple times? 
I'm wondering if it makes sense to put some check, e.g., `grep ${VERSION_BUILD:?} claw-variables` before appending the var file.
Is there any scenario where we can get merge conflicts on the claw repo side?

gururajsh: Merge conflict should fail the job. Will create a separate PR to add a duplicate check since other PRs are already merged

.github/workflows/release-build-sign-upload.yml
```diff
@@ -1033,4 +1006,88 @@ jobs:
...
+    - setup
+    - s3-upload
+    env:
+      VERSION_BUILD: ${{ needs.setup.outputs.version-build }}
+    steps:
+    - name: Checkout CLAW
+      uses: actions/checkout@v4
+      with:
+        repository: cloudfoundry/CLAW
+        ref: develop
```
a-b, line 1070: Do we still want to merge PRs manually on the CLAW side? Otherwise perhaps we could bump versions on the `master` (we have to rename it long time ago) branch.

gururajsh: We don't merge manually on CLAW. develop to master promotion happens automatically
Yes.. we need to change the name

a-b (approved):

PR #2881, opened 2024-05-06
Add "total_service_keys" to quota json unmarshalling

sleungcy-sap:
- [x] [main](https://github.com/cloudfoundry/cli/pull/2881)
- [x] [v7](https://github.com/cloudfoundry/cli/pull/2884)
- [x] [v8](https://github.com/cloudfoundry/cli/pull/2887)

# Description of the Change
Consumers of this library can access the total_service_keys field from the api with this change.

# Why Is This PR Valuable?
The value of this parameters is not included into unmarshalled data and is missing at the moment. The value should be available from the API as documented in https://v3-apidocs.cloudfoundry.org/version/3.163.0/#organization-quotas-in-v3


gururajsh (commented):
Thank you for your contribution to CLI. Looks like this change needs to be applied at other places like quota_resource_test.go and space and org quota tests in different packages and several other non test files. Can you please update the PR with these changes?

resources/quota_resource.go
```diff
@@ -71,6 +71,7 @@ func (al *AppLimit) UnmarshalJSON(rawJSON []byte) error {
 type ServiceLimit struct {
 	TotalServiceInstances *types.NullInt `json:"total_service_instances,omitempty"`
 	PaidServicePlans      *bool          `json:"paid_services_allowed,omitempty"`
+	TotalServiceKeys      *types.NullInt `json:"total_service_keys,omitempty"`
 }
 
 func (sl *ServiceLimit) UnmarshalJSON(rawJSON []byte) error {
```
Samze, line 77: **question** Do we need some custom logic + test to handle the nil case? Seems like we have that already for `TotalServiceInstances`. See https://github.com/cloudfoundry/cli/pull/2881/files#diff-305a3cec50fa59b749c31f4599960a65a465a66bfca7afe1ccce4fcb58fad06eL87-L92

Samze:
LGTM, two quick questions I had is that while this change adds the definition to the go-lang structs, this does not add them to any CLI output https://github.com/cloudfoundry/cli/blob/main/command/v7/shared/quota_displayer.go#L57-L66 

Firstly, @sleungcy-sap I want to double check that was the intention - this is just a change in the CLI library for consumers of the go module right? 

Secondly (non-blocking on this MR), @a-b / @gururajsh if we have this in the struct - should we be adding this to the CLI output? The column output is treated like an API contract right, is there a process for adding new columns (e.g. can we just add a new column on the end or is this considered a breaking change)?

Samze:
@sleungcy-sap we discussed this on the [WG call](https://docs.google.com/document/d/1wP21mVwaoZw45eU2YoDJg6KbvYJuD5ATrrhSY6_yHmM/edit) this week. While the cli is [listed](https://docs.cloudfoundry.org/devguide/capi/client-libraries.html#unofficial) as an experimental client, ideally we would not add functionality to the cli library without exposing the feature in the UX of the CLI itself. 

As an alternative have you considered using https://github.com/cloudfoundry/go-cfclient which is now listed as the `Go tools` area in https://github.com/cloudfoundry/community/blob/main/toc/working-groups/app-runtime-interfaces.md. 

If you are interested in expanding the PR to include contributions for UX changes, then we can discuss the next steps for that.

gururajsh:
Closing this PR for now, feel free to create an issue that includes a proposal for the UX changes.

PR #2906, opened 2024-05-10
Add version check to CLAW update [main]

gururajsh:
## Description of the Change

## Why Is This PR Valuable?

## Applicable Issues

None

## How Urgent Is The Change?

## Other Relevant Parties


moleske (approved):
Looks suspiciously similar to those prs for the other branches 😄

PR #2907, opened 2024-05-10
Add version check to CLAW update [v8]

gururajsh:
## Description of the Change

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

## Other Relevant Parties


.github/workflows/release-build-sign-upload.yml
```diff
@@ -1000,4 +1000,93 @@ jobs:
...
+        ref: master
+        path: CLAW
+        ssh-key: ${{ secrets.GIT_SSH_KEY_CLAW }}  
+
+    - name: Add new version to claw variables
+      run: |
+        set -ex
+
+        pushd CLAW
+          if  grep -q "${VERSION_BUILD}" "claw-variables.yml" ; then
```
a-b, line 1073: I'm wondering if we can skip `if`. `grep --quiet --invert-match` returns non zero exit code if pattern detected. This will interrupt script execution and fail this step.

I strongly suggest we put full flag names in pur scripts. This improves readability and comprehension.

.github/workflows/release-build-sign-upload.yml
```diff
@@ -1000,4 +1000,93 @@ jobs:
...
+
+            git add claw-variables.yml
+
+            if ! [ -z "$(git status --porcelain)"]; then
+              git config user.name github-actions
+              git config user.email github-actions@github.com
+              git commit -m "Add CF CLI ${VERSION_BUILD}"
+            else
+                echo "no new version to commit"
+            fi
```
a-b, line 1087:
```sh

if git status --porcelain; then
  echo "no new version to commit"
else
  git ..
fi

```

a-b (changes requested):
Please consider suggestions above

.github/workflows/release-build-sign-upload.yml
gururajsh: What is wrong if we check the other way round?

.github/workflows/release-build-sign-upload.yml
gururajsh: Will keep that in mind, thank you!!

PeteLevineA (approved):

.github/workflows/release-build-sign-upload.yml
a-b: Improved readability, less complexity.

gururajsh: I think its one's preference of how an if statement is written

gururajsh: I see where you are coming from. The reason I used this statement is because Its consistent with how its written in other [places](https://github.com/cloudfoundry/cli/blob/main/.github/workflows/release-update-repos.yml#L193C11-L200C1).  May be we can have a story to change the if statements the way you suggested.

gururajsh: Bump

moleske (approved):
seems like a reasonable approach to me

gururajsh:
Thank you @moleske .

a-b (approved):
LGTM

PR #2908, opened 2024-05-10
Add version check to CLAW update [v7]

gururajsh:
## Description of the Change

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

## Other Relevant Parties


PeteLevineA (approved):

moleske (approved):
Looks suspiciously similar to those prs for the other branches 😄

.github/workflows/release-build-sign-upload.yml
```diff
@@ -1067,7 +1067,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: cloudfoundry/CLAW
-        ref: develop
+        ref: master
         path: CLAW
         ssh-key: ${{ secrets.GIT_SSH_KEY_CLAW }}
```
a-b, line 1072: Is using (read/write) GITHUB_TOKEN to work with this repo possible in this case?

https://docs.github.com/en/actions/security-guides/automatic-token-authentication
https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp#

PR #2920, opened 2024-05-14
Install ginkgo "out of the box"

chinigo:
## Description of the Change

This change installs the version of Ginkgo specified in `go.mod`, modifies the `Makefile` to do this before each flavor of test, and simplifies the CI tasks that invoke these tests.

## Why Is This PR Valuable?

Prior to this change, the `Makefile` assumes the presence of a system-level installation of Ginkgo. That installation of Ginkgo will not necessarily match the version specified in our `go.mod`.

This introduces the (slight) possibility of subtle differences in the behavior of the test library. It also complicates spinning up a new environment, either for a developer or in CI.

## Applicable Issues

n/a

## How Urgent Is The Change?

Not urgent.

## Related PRs

- #2926 - for the `v7` branch
- #2927 - for the `v8` branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,6 @@ install-test-deps: ## Install "global" dependencies needed to run tests
 # Running `go install <pkg>` without a version specifier will use version specified in go.mod
 # See https://go.dev/ref/mod#go-install
 	go install github.com/onsi/ginkgo/v2/ginkgo
-	go install github.com/onsi/gomega/matchers
 
 integration-cleanup:
 	$(CURDIR)/bin/cleanup-integration
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,12 @@ fly-windows-units:
 format: ## Run go fmt
 	go fmt ./...
 
+install-test-deps: ## Install "global" dependencies needed to run tests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -71,6 +71,12 @@ fly-windows-units:
 format: ## Run go fmt
 	go fmt ./...
 
+install-test-deps: ## Install "global" dependencies needed to run tests
+# Running `go install <pkg>` without a version specifier will use version specified in go.mod
+# See https://go.dev/ref/mod#go-install
+	go install github.com/onsi/ginkgo/v2/ginkgo
+	go install github.com/onsi/gomega/matchers
```
Samze, line 78: **question/non-blocking** Unrelated directly to this change, but why do we install `github.com/onsi/gomega/matchers`? I don't believe thats a binary, its just the library used in the test cases for assertions.

moleske (approved):
Looks reasonable to me.  You'll want to open this change against v8 and v7 branches as well

chinigo:
> Looks reasonable to me. You'll want to open this change against v8 and v7 branches as well

@moleske thank you for the reminder. I've opened #2926 and #2927 to add these changes to those branches as well.

Samze (approved):
LGTM

PR #2921, opened 2024-05-14
Remove custom linter

Samze:
## Description of the Change

There appears to be have been a custom linter introduced a long time ago, this is currently failing when running `make all` with:
```
....
Running custom linters...
bash -c "go run bin/style/main.go api util"
api/shared/wrap_for_cf_on_k8s.go +99 type declaration certRoundTripper defined after a function declaration
api/shared/wrap_for_cf_on_k8s.go +104 method certRoundTripper.RoundTrip defined after function getCert
exit status 1
make: *** [custom-lint] Error 1
```

Instead of fixing the custom linter, lets just remove it and rely on `golangci-lint`

## Why Is This PR Valuable?

Allows developers to do `git clone && make` on the repo.

## How Urgent Is The Change?

Not urgent


joaopapereira (approved):
The change looks good.
The only non-blocking question is if we want to find a linter that checks if we use a function before we define it. I am ok with *not* adding that check.

Samze:
> The change looks good. The only non-blocking question is if we want to find a linter that checks if we use a function before we define it. I am ok with _not_ adding that check.

Yeah, since unlike some interpreted languages golang has compile checks for functions and ordering doesn't matter, I'd personally be ok with not having this check.

gururajsh (approved):
LGTM

PR #2927, opened 2024-05-15
Install ginkgo "out of the box" (v8 branch)

chinigo:
## Note

Functionally a duplicate of #2920, applied to the `v8` branch.

## Description of the Change

This change *to the `v8` branch* installs the version of Ginkgo specified in `go.mod`, modifies the `Makefile` to do this before each flavor of test, and simplifies the CI tasks that invoke these tests.

## Why Is This PR Valuable?

Prior to this change, the `Makefile` assumes the presence of a system-level installation of Ginkgo. That installation of Ginkgo will not necessarily match the version specified in our `go.mod`.

This introduces the (slight) possibility of subtle differences in the behavior of the test library. It also complicates spinning up a new environment, either for a developer or in CI.

## Applicable Issues

n/a

## How Urgent Is The Change?

Not urgent.

## Related PRs

- #2920 - for the `main` branch
- #2926 - for the `v7` branch

### DIFF
--- a/.github/workflows/tests-integration-reusable.yml
+++ b/.github/workflows/tests-integration-reusable.yml
@@ -249,8 +249,6 @@ jobs:
 
         make build
 
-        go install github.com/onsi/ginkgo/v2/ginkgo
-
         export CF_PASSWORD=${CF_INT_PASSWORD}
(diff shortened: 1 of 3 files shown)


gururajsh (approved):
LGTM

PR #2931, opened 2024-05-17
Fix race condition in config handling (v8)

### LINKED ISSUE #2232: Error writing config: rename ~/.cf/temp-config929711920 ~/.cf/config.json: no such file or directory

- [x] I am reporting a bug that others will be able to reproduce.

Provide details on what you were trying to do (and why).

**What happened**
When I was using GNU parallel to run some scripts that can execute multiple cf commands in parallel, I got the below error for one job, but others worked:

| |
|-|
| **Error writing config: rename /Users/user/.cf/temp-config929711920 /Users/user/.cf/config.json: no such file or directory**

I compared the **config.json** file and found that only the AccessToken field was different when that file was changed. And after inspecting the AccessToken, especially the `iat` and `exp` fields, where the latter is only 1-hour after the former. 

And thanks to  https://github.com/cloudfoundry/cli/pull/1594, the token must have been automatically refreshed when my script was running. So more then one cf commands tried to refresh the token and it caused contentions even they try to use temp files.

**Expected behavior**
Whether there's an even smarter way to avoid parallel file writing contention, I would expect the life span of the CF CLI access token can be extended to something like 1 day, or at least 12 hours.

The default 1-hour life span of the CF access token is really just too short, especially when one needs to work with multiple CF spaces, BTW I also use $CF_HOME to work with multiple CF spaces simultaneously.

**Exact Steps To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. Run `cf login`, to a space where multiple apps or services are deployed
2. Take the **`AccessToken`** field in **$CF_HOME/.cf/config.json**
3. Use a JWT decoder to inspect the above **AccessToken**, check the `exp` field
4. Convert the value of the `exp` field to a readable date, e.g. using `date -d@1636967840`
5. At the time when the token is about to expire, launch some command that uses GNU parallel to run multiple cf commands concurrently, e.g., `cf apps | awk 'NR>4 {print $1}' | parallel -N1 cf env`

**Provide more context**
- platform and shell details: Mac OS 12.0.1, iTerm 3.4.12,  Bash 5.1.8 installed via Homebrew
- version of the CLI you are running: `cf version 8.0.0+e8d41cf8e.2021-09-16`
- version of the CC API Release you are on:  3.108.0"

**Notes regarding V6 and V7 CLI support:** 
- V6:
  - Minimum supported version of CF Deployment: v7.0.0 (CAPI Release: 1.74.0 (APIs 2.128.0 and 3.63.0))
  - Maximum supported version of CF Deployment: v13.4.0 (CAPI Release: 1.94.0 (APIs 2.149.0 and 3.84.0))
- V7:
  - Minimum supported version of CF Deployment: v13.5.0 (CAPI Release: 1.95.0 (APIs 2.150.0 and 3.85.0))

f-blass:
## Description of the Change

The CLI encounters race conditions when handling config files if multiple commands are executed in parallel.

Currently, the CLI writes a new config for every command executed due to the code in [command_parser.go](https://github.com/cloudfoundry/cli/blob/ae68111aeda81647283e94e3aee84115e4d27544/util/command_parser/command_parser.go#L72-L77). However, it also deletes all temporary config files when reading the config, leading to race conditions between multiple processes. This PR introduces two changes to mitigate this problem:

1. **Ignore "File Not Found" errors when attempting to delete a file.** This change addresses the issue where multiple processes list and delete temporary files simultaneously.
2. **Only delete temp config files older than 5 minutes.** This change resolves the problem where one process creates a temporary config that it later wants to rename, while a second process deletes this temp config in the meantime. Orphan files should be rare, as they are normally pruned by the signal handler, as seen in https://github.com/cloudfoundry/cli/blob/04df8ae6dd68b64c2080c863c1ddf4c3e06abcd6/util/configv3/write_config.go#L63-L66

## Why Is This PR Valuable?

Enables usage of the CF CLI in scripts which execute commands in parallel

## Applicable Issues

fixes #2232 

## How Urgent Is The Change?

Medium

## Other Relevant Parties

Anyone using the CLI in scripts with parallel execution. Multiple users have encountered this issue, as discussed in #2232.

## Related PRs

- [v7](https://github.com/cloudfoundry/cli/pull/2936)
- [main](https://github.com/cloudfoundry/cli/pull/2937)


gururajsh:
Hi @f-blass,

Thank you for your contribution. Looks like some tests need to be [addressed](https://github.com/cloudfoundry/cli/actions/runs/9133676652/job/25117683388#step:15:409) as part of this fix. Can you please look into it.

f-blass:
Hi @gururajsh 
thank you! Should be fixed now.

gururajsh (commented):
Changes look good. Can you please create PRs for main and v7 branches as well? Asking for v7, since its a fix for an issue.

f-blass:
Hi @gururajsh 
thank you. PRs for v7 and main branch are created and linked.

gururajsh (approved):
LGTM

PR #2932, opened 2024-05-17
Add appGUID attribute to droplet resource

gmllt:

- [x] [main](https://github.com/cloudfoundry/cli/tree/main) - all changes should by default start here ([PR on main branch](https://github.com/cloudfoundry/cli/pull/2837))

## Description of the Change

The V3 API allows you to access to the relationships existing beween a droplet and its related app.
This PR make this relationship visible trough a new `AppGUID` attribute in the resource `droplet`.

The community `cf_exporter` currently use the `/v2/spaces/:guid/summary` endpoint to gain access to buildpack associated to an application. Have this relationship visible would help get rid of calls on this endpoint and retrieve buildpacks attached to an application through its droplets.

This feature is necessary for applications using the cli product in go development to established relationship between an `application` resource and its droplets.

This feature will not cause breaking change.

### DIFF
--- a/resources/droplet_resource.go
+++ b/resources/droplet_resource.go
@@ -1,12 +1,16 @@
 package resources
 
 import (
+	"code.cloudfoundry.org/cli/api/cloudcontroller"
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant"
+	"encoding/json"
 )
 
 // Droplet represents a Cloud Controller droplet's metadata. A droplet is a set of
 // compiled bits for a given application.
 type Droplet struct {
+	// AppGUID is the unique identifier of the application associated with the droplet.
+	AppGUID string `json:"app_guid"`
 	//Buildpacks are the detected buildpacks from the staging process.
 	Buildpacks []DropletBuildpack `json:"buildpacks,omitempty"`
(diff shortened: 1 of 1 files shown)


resources/droplet_resource.go
```diff
@@ -35,3 +39,77 @@ type DropletBuildpack struct {
...
+		GUID string `json:"guid,omitempty"`
+	}
+
+	type RelationshipData struct {
+		Data Data `json:"data,omitempty"`
+	}
+
+	type Relationships struct {
+		App RelationshipData `json:"app,omitempty"`
+	}
```
joaopapereira, line 54: Any particular reason for the difference between these structs and the ones for the unmarshal (named structs vs anonymous structs)? My suggestion is if we are keeping these inside functions personally I would prefer to have them anonymous.

a-b:
@dependabot rebase

resources/droplet_resource.go
a-b: @gmllt wanted to follow up. Do you have any feedback on this?

gmllt: @joaopapereira & @a-b sorry for the huge delay, here is the implementation with anonymous structures

joaopapereira (approved):
LGTM

joaopapereira (approved):
LGTM

PR #2964, opened 2024-06-05
Min capi

joaopapereira:
## Description of the Change

Re-enable the tests to ensure that CF CLI is compatible with the minimum version of CAPI


gururajsh (approved):
LGTM

PR #2967, opened 2024-06-10
Start using cflinuxfs3

joaopapereira:
## Description of the Change

This change will add a new opsfile to enable integration tests in edge to use cflinuxfs3.

### DIFF
--- a/.github/ops-files/use-cflinuxfs3.yml
+++ b/.github/ops-files/use-cflinuxfs3.yml
@@ -1,0 +1,105 @@
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/diego/droplet_destinations/cflinuxfs3?
+  value: /home/vcap
+- type: replace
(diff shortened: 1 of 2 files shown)


moleske (commented):
Why are we adding cflinuxfs3?  It is no longer support in the open source community.  I am vaguely aware of some companies providing private cflinuxfs3 as extended support but not convinced that is something we should be testing on this open source repo.  Especially as the repo is now [archived](https://github.com/cloudfoundry/cflinuxfs3-release) and the last release was in June 2023

joaopapereira:
We are in the process of changing pools we are using to run tests for cli and to do that, we need to ensure that linuxfs-3 is present in the deployment, or else a big group of tests would have to be changed. This change is also necessary to support the current min capi version tests we will re-add.

Another point here is that we are planning on stopping support v7 around Oct, so the amount of changes that we would need to do is too big compared with just adding support for Linux-3 is too big for something that we plan on stopping supporting in a few months' time.

moleske:
Definitely curious which tests need to change.  Though I'm assuming these tests are only on the v7 branch.

As long as this is old contained to the v7 branch, seems reasonable.

a-b (approved):
LGTM

PR #2968, opened 2024-06-11
Fix pr issues

joaopapereira:
## Description of the Change

Fix github action failing while testing PRs. We found out that when we were using workflow_run, we were always retrieving the workflow files from the main branch. The main change that is made here is to stop relying on workflow_run and to call the integration tests after the unit tests are successful.

### DIFF
--- a/.github/workflows/tests-integration-reusable.yml
+++ b/.github/workflows/tests-integration-reusable.yml
@@ -28,12 +28,6 @@ on:
       pool-namespace:
         type: string
         default: 'official'
-    secrets:
-      SHEPHERD_SERVICE_ACCOUNT_TOKEN:
-        required: true
-      CLIENT_SECRET:
-        required: true
-
 jobs:
   run-integration-tests:
     defaults:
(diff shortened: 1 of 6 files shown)


moleske (approved):
I like this flow better cause then the branches won't be dependent on main.  That has been tricking people for awhile.

I'm approving cause this flow seems reasonable.  I'm assuming the errors about `Incorrect number of arguments provided` will be resolved before merging.  Also not sure if this is one of the cases where stuff needs to be merged first before seeing it work (that was often the case with the previous method of getting the workflow from the main branch)

a-b:
The integration tests workflow was designed in that specific way to enable PRs from forks to run integration tests using repository secrets containing credentials for the Shepherd. It also requires approval to run tests for new contributors, which helps maintain security for sensitive information. 

Do we know if the proposed change operates in the equivalent way?

joaopapereira:
The only change this PR introduces is how the integration tests are triggered. Instead of using a "global" trigger, each workflow explicitly runs the integration tests, which is the only change.

gururajsh (approved):
LGTM

a-b (commented):
LGTM

PR #2973, opened 2024-06-20
[v8] Fix pr issues

joaopapereira:
## Description of the Change

Fix github action failing while testing PRs. We found out that when we were using workflow_run, we were always retrieving the workflow files from the main branch. The main change that is made here is to stop relying on workflow_run and to call the integration tests after the unit tests are successful.

### DIFF
--- a/.github/workflows/tests-integration-reusable.yml
+++ b/.github/workflows/tests-integration-reusable.yml
@@ -27,7 +27,7 @@ on:
         default: ${{ vars.SHEPHERD_POOL_NAME }}
       pool-namespace:
         type: string
-        default: 'tas-devex'
+        default: 'official'
       is-pr:
         type: boolean
         default: true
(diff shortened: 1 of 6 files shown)


gururajsh (approved):
LGTM

.github/workflows/tests-integration-reusable.yml
```diff
@@ -40,17 +36,6 @@ jobs:
     runs-on: ${{ inputs.os }} 
     container: us-west2-docker.pkg.dev/shepherd-268822/shepherd2/concourse-resource:latest
     steps:
-    - uses: LouisBrunner/checks-action@v2.0.0
```
a-b, line 43: @joaopapereira, does the removal of this step mean that the integration tests run will no longer be reflected in the check?

joaopapereira: Given that the tests are now part of the unit tests workflow, there is no need to add extra check tests

PR #2998, opened 2024-07-08
Add missing file close

Samze:
## Description of the Change

Add missing close, from https://github.com/cloudfoundry/cli/pull/2919


a-b (approved):
Good catch!

PR #3004, opened 2024-07-09
Update inno setup x64 flag to x64compatible

weresch:
## Description of the Change

Inno Setup is used to create installers for release, and it no longer supports the ia64 flag and recommends x64compatible instead. See https://jrsoftware.org/ishelp/index.php?topic=setup_architecturesinstallin64bitmode

## Why Is This PR Valuable?

Removes an error from building the release

## How Urgent Is The Change?

Needed before next release


gururajsh (approved):
LGTM

PR #3014, opened 2024-07-10
Add path to checkout action [v8]

gururajsh:
To keep the downloaded artifacts and checked out repo path separate, added path to checkout action
## Description of the Change

Fix for creating github draft release job. Downloaded artifacts were getting deleted by the checkout action since it was being checked out to $GITHUB_WORKSPACE. So added path to checkout action.

## Why Is This PR Valuable?

Helps with release process

## Applicable Issues

## How Urgent Is The Change?

## Other Relevant Parties


joaopapereira (approved):
LGTM

PR #3028, opened 2024-07-15
Use latest CAPI

gururajsh:
Use the latest version of CAPI to run integration tests.


integration/shared/isolated/error_handling_test.go
```diff
@@ -19,7 +19,7 @@ var _ = Describe("curl command", func() {
 		It("returns the expected request with verbose output", func() {
 			session := helpers.CF("curl", "-v", "/v2/banana")
 			Eventually(session).Should(Say("GET /v2/banana HTTP/1.1"))
-			Eventually(session).Should(Say(`"error_code": "CF-NotFound"`))
+			Eventually(session).Should(Say(`"error_code":"CF-NotFound"`))
```
Samze, line 22: **nit/non-blocking** Perhaps just searching for the string `CF-NotFound` would be more robust.

PR #3039, opened 2024-07-22
Remove unused property [main]

gururajsh:
The integration workflow file had an unused property which was causing the workflow to fail. This PR removes it.


moleske (approved):
haven't tested myself to see it work (won't see it work on the pr til it is fixed)

since the problem is this line, seems reasonable to just delete it

PR #3042, opened 2024-07-22
Display deployment status for cf app command [main]

gururajsh:
When the cf app command is executed with strategy rolling, it displays the rolling deployment status(DEPLOYING or CANCELING) if one is in progress.
Ex. Rolling deployment is DEPLOYING


Samze (approved):
LGTM

Samze (approved):

PR #3054, opened 2024-07-26
Max in flight

joaopapereira:
## Description of the Change

Add max-in-flight flag to the push command. This will allow the users to  set the maximum number of instances that they want to roll at any given time when a new version of the app is being pushed


joaopapereira:
New PR was created for this

PR #3061, opened 2024-07-29
`cf revisions` no longer reports EXPERIMENTAL notice [v8]

chinigo:
## Description of the Change

This commit changes the `cf revision` and `cf revisions` commands to no longer print a notice that the command is experimental.

## Why Is This PR Valuable?

The `cf revision` and `cf revisions` commands were introduced over 4 years ago and are no longer experimental.

## Applicable Issues

n/a

## How Urgent Is The Change?

Not urgent.

## Other Relevant Parties

None

## Related PRs

- #3060 - v7 
- #3061 - v8 (this PR)
- #3062 - main

PR #3060, opened 2024-07-29
`cf revisions` no longer reports EXPERIMENTAL notice [v7]

chinigo:
## Description of the Change

This commit changes the `cf revision` and `cf revisions` commands to no longer print a notice that the command is experimental.

## Why Is This PR Valuable?

The `cf revision` and `cf revisions` commands were introduced over 4 years ago and are no longer experimental.

## Applicable Issues

n/a

## How Urgent Is The Change?

Not urgent.

## Other Relevant Parties

None

## Related PRs

- #3060 - v7 (this PR) 
- #3061 - v8
- #3062 - main

PR #3062, opened 2024-07-29
`cf revisions` no longer reports EXPERIMENTAL notice [main]

chinigo:
## Description of the Change

This commit changes the `cf revision` and `cf revisions` commands to no longer print a notice that the command is experimental.

## Why Is This PR Valuable?

The `cf revision` and `cf revisions` commands were introduced over 4 years ago and are no longer experimental.

## Applicable Issues

n/a

## How Urgent Is The Change?

Not urgent.

## Other Relevant Parties

None

## Related PRs

- #3060 - v7 
- #3061 - v8
- #3062 - main (this PR)


a-b (approved):
LGTM

joaopapereira (changes requested):
Given that the test was there to ensure that the message was present I do not think these tests make sense anymore.
If you come back to the code and see this test, it is strange 😄

chinigo:
- Negative test assertions removed (including one whole file that contained nothing but).
- Moved the `revisions` and `revision` command into the `APPS` section of the `cf help` output.

command/common/internal/help_all_display.go
```diff
@@ -172,7 +173,7 @@ var ExperimentalHelpCategoryList = []HelpCategory{
 	{
 		CategoryName: "EXPERIMENTAL COMMANDS:",
 		CommandList: [][]string{
-			{"revisions", "revision", "rollback"},
```
Samze, line 175: **issue** Revision should not be made unexperimental since it does not do anything

chinigo: Ah you're right, it was too aggressive to move _both_ of these commands out of the EXPERIMENTAL block. I also see now that [the ticket](https://jira.eng.vmware.com/browse/TPCF-10927) mentions in passing that the `rollback` command should also not be marked EXPERIMENTAL.

I'll submit another revision to all three of these PRs that:
- Restores `revision` to `EXPERIMENTAL` status.
- Moves `rollback` _out_ of `EXPERIMENTAL` status.

a-b (approved):

command/common/internal/help_all_display.go
chinigo: Updated PR as promised:

- `cf revision` remains in `EXPERIMENTAL` status.
- `cf revisions` is no longer in `EXPERIMENTAL` status.
- `cf rollback` is no longer in `EXPERIMENTAL` status.

Samze (approved):

joaopapereira (approved):
LGTM

PR #3069, opened 2024-07-31
Fix unit tests to use the local timezone [main]

weresch:
## Description of the Change

Some unit tests recently added expected an EST timezone. This fix has it use the local environment timezone.


weresch:
Unit tests don't seem to be rerunning on the latest commit as the test failure messages are old. Closing this PR to open a fresh one.

PR #3075, opened 2024-08-02
Temporarily move integration tests to the push event

joaopapereira:
## Description of the Change

Temporarily move integration tests to the push event

- We will change this commit in the near future to ensure that the integration tests run again on pull requests

### DIFF
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -24,6 +24,11 @@ on:
       - "Tests"
     types:
       - completed
+  push:
+    branches:
+      - main
+      - v8
+      - v7
 
 jobs:
   run-integration-tests-cf-env:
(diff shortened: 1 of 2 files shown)


weresch:
LGTM

.github/workflows/tests-integration.yml
```diff
@@ -24,6 +24,11 @@ on:
       - "Tests"
     types:
       - completed
+  push:
+    branches:
```
a-b, line 28: question: Is it possible to run integration tests only if the unit tests pass successfully?

joaopapereira: At this point we decided only to run integration tests on push (after the pr is merged) This is a little bit strange but in reality, this was what was happening with the bug we found yesterday.
This solution is a stop-gap solution and I hope we can start the work to re-add the integration tests into the PR workflow at the beginning of next week.

a-b (approved):

PR #3085, opened 2024-08-07
[v8] Max in flight flag

joaopapereira:
## Description of the Change

Add the max-in-flight flag to all the commands


actor/v7pushaction/create_deployment_for_push_plan.go
```diff
@@ -8,10 +8,15 @@ import (
 func (actor Actor) CreateDeploymentForApplication(pushPlan PushPlan, eventStream chan<- *PushEvent, progressBar ProgressBar) (PushPlan, Warnings, error) {
 	eventStream <- &PushEvent{Plan: pushPlan, Event: StartingDeployment}
 
-	var dep resources.Deployment
-	dep.DropletGUID = pushPlan.DropletGUID
-	dep.Strategy = pushPlan.Strategy
-	dep.Relationships = resources.Relationships{constant.RelationshipTypeApplication: resources.Relationship{GUID: pushPlan.Application.GUID}}
+	dep := resources.Deployment{
```
pivotalgeorge, line 11: [comment, no action] good improvement / we should be using this style. I wonder if our linter catches the assignment approach you've replaced

actor/v7pushaction/setup_deployment_information_for_push_plan.go
```diff
@@ -0,0 +1,13 @@
+package v7pushaction
+
+import "code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant"
+
+func SetupDeploymentInformationForPushPlan(pushPlan PushPlan, overrides FlagOverrides) (PushPlan, error) {
```
pivotalgeorge, line 5: [suggestion / question / nonblocking] why is this separate from create_deployment_for_push_plan

command/v7/copy_source_command.go
```diff
@@ -17,6 +17,7 @@ type CopySourceCommand struct {
 	RequiredArgs        flag.CopySourceArgs     `positional-args:"yes"`
 	usage               interface{}             `usage:"CF_NAME copy-source SOURCE_APP DESTINATION_APP [-s TARGET_SPACE [-o TARGET_ORG]] [--no-restart] [--strategy STRATEGY] [--no-wait]"`
 	Strategy            flag.DeploymentStrategy `long:"strategy" description:"Deployment strategy can be canary, rolling or null"`
+	MaxInFlight         int                     `long:"max-in-flight" default:"-1" description:"Defines the maximum number of instances that will be actively being started. Only applies when --strategy flag is specified."`
```
pivotalgeorge, line 20: [suggestion, blocking] it feels like the default should be the same as CAPI (1) or at least a valid value (resolves to the same, since 0 and negative numbers are invalid here)

command/v7/copy_source_command.go
```diff
@@ -48,6 +49,14 @@ func (cmd *CopySourceCommand) ValidateFlags() error {
 		}
 	}
 
+	if cmd.Strategy.Name == constant.DeploymentStrategyDefault && cmd.MaxInFlight > 0 {
+		return translatableerror.RequiredFlagsError{Arg1: "--max-in-flight", Arg2: "--strategy"}
+	}
+
+	if cmd.Strategy.Name != constant.DeploymentStrategyDefault && (cmd.MaxInFlight < -1 || cmd.MaxInFlight == 0) {
```
pivotalgeorge, line 56: [suggestion, nonblocking] (cmd.MaxInFlight <= 0)

actor/v7pushaction/create_deployment_for_push_plan_test.go
```diff
@@ -106,6 +107,51 @@ var _ = Describe("CreateDeploymentForApplication()", func() {
...
+				Expect(dep).To(Equal(resources.Deployment{
+					Strategy: "rolling",
+					Options:  resources.DeploymentOpts{MaxInFlight: 10},
+					Relationships: resources.Relationships{
+						constant.RelationshipTypeApplication: resources.Relationship{GUID: "some-app-guid"},
+					},
+				}))
+			})
+
+			It("returns errors and warnings", func() {
```
pivotalgeorge, line 145: [suggestion, nonblocking] do we want a happy-path case covered?

command/v7/push_command.go
```diff
@@ -91,6 +91,7 @@ type PushCommand struct {
 	LogRateLimit            string                              `long:"log-rate-limit" short:"l" description:"Log rate limit per second, in bytes (e.g. 128B, 4K, 1M). -l=-1 represents unlimited"`
 	PathToManifest          flag.ManifestPathWithExistenceCheck `long:"manifest" short:"f" description:"Path to manifest"`
 	Memory                  string                              `long:"memory" short:"m" description:"Memory limit (e.g. 256M, 1024M, 1G)"`
+	MaxInFlight             int                                 `long:"max-in-flight" default:"-1" description:"Defines the maximum number of instances that will be actively being started. Only applies when --strategy flag is specified."`
```
pivotalgeorge, line 94: see other comment about default of -1

command/v7/push_command_test.go
```diff
@@ -1290,5 +1292,23 @@ var _ = Describe("push Command", func() {
 					"--task", "--strategy=canary",
 				},
 			}),
+
+		Entry("max-in-flight is passed without strategy",
+			func() {
+				cmd.MaxInFlight = 10
+			},
+			translatableerror.RequiredFlagsError{
```
pivotalgeorge, line 1300: [question, non-blocking] does it feel to you/anyone else like we are over-testing? I feel like I've seen several of the same error expectations at different levels
edit: nevermind if this is the same test for the different commands

command/v7/restage_command.go
```diff
@@ -16,6 +16,7 @@ type RestageCommand struct {
 
 	RequiredArgs        flag.AppName            `positional-args:"yes"`
 	Strategy            flag.DeploymentStrategy `long:"strategy" description:"Deployment strategy can be canary, rolling or null."`
+	MaxInFlight         int                     `long:"max-in-flight" default:"-1" description:"Defines the maximum number of instances that will be actively being restaged. Only applies when --strategy flag is specified."`
```
pivotalgeorge, line 19: same as earlier comment about default value of -1

command/v7/restart_command.go
```diff
@@ -50,6 +52,11 @@ func (cmd RestartCommand) Execute(args []string) error {
 		return err
 	}
 
+	err = cmd.ValidateFlags()
```
pivotalgeorge, line 55: [comment, nonblocking] I like this pattern

command/v7/rollback_command.go
```diff
@@ -19,6 +20,7 @@ type RollbackCommand struct {
 	Version         flag.Revision `long:"version" required:"true" description:"Roll back to the specified revision"`
 	relatedCommands interface{}   `related_commands:"revisions"`
 	usage           interface{}   `usage:"CF_NAME rollback APP_NAME [--version VERSION] [-f]"`
+	MaxInFlight     int           `long:"max-in-flight" default:"-1" description:"Defines the maximum number of instances that will be actively being rolled back."`
```
pivotalgeorge, line 23: same as previous default value comment

command/v7/restart_command.go
```diff
@@ -6,12 +6,14 @@ import (
...
 	"code.cloudfoundry.org/cli/command"
 	"code.cloudfoundry.org/cli/command/flag"
+	"code.cloudfoundry.org/cli/command/translatableerror"
 	"code.cloudfoundry.org/cli/command/v7/shared"
 )
 
 type RestartCommand struct {
 	BaseCommand
 
+	MaxInFlight         int                     `long:"max-in-flight" default:"-1" description:"Defines the maximum number of instances that will be actively restarted at any given time. Only applies when --strategy flag is specified."`
```
pivotalgeorge, line 16: same as previous default value comment

integration/v7/isolated/continue_deployment_test.go
```diff
@@ -87,7 +87,7 @@ var _ = Describe("Continue Deployment", func() {
 			When("There is a canary deployment", func() {
 				It("succeeds", func() {
 					helpers.WithHelloWorldApp(func(appDir string) {
-						helpers.CF("push", appName, "-p", appDir, "--strategy=canary", "--no-wait").Wait()
+						helpers.CF("push", appName, "-p", appDir, "--strategy=canary").Wait()
```
pivotalgeorge, line 90: [question, nonblocking] do we still want to test --no-wait somewhere

resources/deployment_resource.go
```diff
@@ -22,6 +23,10 @@ type Deployment struct {
 	Strategy         constant.DeploymentStrategy
 }
 
+type DeploymentOpts struct {
+	MaxInFlight int `json:"max_in_flight"`
```
pivotalgeorge, line 27: 👍

resources/deployment_resource.go
```diff
@@ -50,6 +56,14 @@ func (d Deployment) MarshalJSON() ([]byte, error) {
 		ccDeployment.Strategy = d.Strategy
 	}
 
+	var b DeploymentOpts
+	if d.Options != b {
+		ccDeployment.Options = &d.Options
+		if d.Options.MaxInFlight < 1 {
+			ccDeployment.Options.MaxInFlight = 1
```
pivotalgeorge, line 63: [concern, blocking] what happens when we talk to an older CAPI, I.e. one that doesn't have max-in-flight? in #3088 we approached this by assuming that a value of 0 was "max-in-flight unset in the response" but I'm reading this block to mean that we'll always be setting max-in-flight to the default value of 1 and including it in a POST to CAPI
edit: Unless I am missing something and this object is only for unmarshalling the CAPI response

pivotalgeorge (commented):
Looks mostly-good to me. some comments are concerns, most are minor suggestions

actor/v7pushaction/setup_deployment_information_for_push_plan.go
joaopapereira: `actor.PreparePushPlanSequence` looked like the place where we were configuring the push plan by translating all the needed flags.

command/v7/push_command_test.go
joaopapereira: We do this test for all the commands we changed

resources/deployment_resource.go
joaopapereira: The concern is valid. When max-in-flight gets to this point with a value we just send it to CAPI and CAPI will return an error, which I am going to address in the next commit.
Also did change the way we are defining if --max-in-flight is present or not by using a pointer instead of the -1 value which should make everything much easier to reason about.

integration/v7/isolated/continue_deployment_test.go
joaopapereira: I talked with @gururajsh and we decided that it would be ok for right now to skip the --no-wait test.

gururajsh (dismissed):
LGTM

command/v7/copy_source_command.go
gururajsh: @pivotalgeorge looks like this concern has been addressed. Are you ok for this conversation to be resolved?

actor/v7pushaction/setup_deployment_information_for_push_plan.go
pivotalgeorge: Okay makes sense (the [pushplansequence](https://github.com/cloudfoundry/cli/blob/383393adc517d887f7c880fca9cb65bd9de07da0/actor/v7pushaction/actor.go#L70)). In that case what's strangely-shaped is the location (siblings within v7pushaction) of two deployment functions that both take pushPlans. I see that this parallels the situation for e.g. droplet config, so I'm happy to move on from this topic for now and leave it up to a separate refactor

command/v7/copy_source_command.go
```diff
@@ -48,6 +49,14 @@ func (cmd *CopySourceCommand) ValidateFlags() error {
 		}
 	}
 
+	if cmd.Strategy.Name == constant.DeploymentStrategyDefault && cmd.MaxInFlight != nil {
```
pivotalgeorge, line 52: [suggestion, blocking] I believe this is now incorrect: MaxInFlight can take a value of 0, which should be invalid for this flag

pivotalgeorge (changes requested):

command/v7/copy_source_command.go
pivotalgeorge: [code since changed, comment in new review]

command/v7/push_command_test.go
pivotalgeorge: sounds good, thanks

resources/deployment_resource.go
pivotalgeorge: agreed on the nullable pointer being easy to reason about. If we handle the version issue in followup work I'm happy to consider this resolved.

command/v7/copy_source_command.go
pivotalgeorge: LGTM

pivotalgeorge:
LGTM

pivotalgeorge (approved):
thanks for addressing previous comments
LGTM

PR #3096, opened 2024-08-09
[main] Max in flight flag

joaopapereira:
* Allow the user to set max-in-flight while pushing app
* Add max-in-flight flag to the restart command
* Add max-in-flight flag to restage command
* Add max-in-flight flag to rollback command
* Add max-in-flight flag to copy-src command
* Remove `--no-wait` to ensure no flakes


gururajsh (approved):
LGTM

PR #3097, opened 2024-08-12
[v8] Print max-in-flight value alongside deployment status

weresch:
## Description of the Change

This prints `max-in-flight: <value>` after the deployment status, when there is an Active deployment for an app and when max-in-flight is set to a non-default value

With #3085 merged into `v8`, this PR takes the new changes from #3088 and targets the `v8` branch.

## Why Is This PR Valuable?

It provides the user with information/confirmation that they have set a non-default max-in-flight value

+@joaopapereira @weresch @pivotalgeorge

PR #3098, opened 2024-08-12
Print max-in-flight value alongside deployment status [main]

pivotalgeorge:
## Description of the Change
This prints `max-in-flight: <value>` after the deployment status, when there is an Active deployment for an app and when max-in-flight is set to a non-default value

## Why Is This PR Valuable?
It provides the user with information/confirmation that they have set a non-default max-in-flight value

## Applicable Issues/PRs

this PR is based on #3096 and is therefore blocked on that PR getting merged into `main`

+@joaopapereira @weresch

### DIFF
--- a/command/v7/shared/app_summary_displayer_test.go
+++ b/command/v7/shared/app_summary_displayer_test.go
@@ -698,10 +698,58 @@ var _ = Describe("app summary displayer", func() {
 		When("there is an active deployment", func() {
 			var LastStatusChangeTimeString = "2024-07-29T17:32:29Z"
 			var dateTimeRegexPattern = `[a-zA-Z]{3}\s\d{2}\s[a-zA-Z]{3}\s\d{2}\:\d{2}\:\d{2}\s[A-Z]{3}\s\d{4}`
+			var maxInFlightDefaultValue = 1
 
 			When("the deployment strategy is rolling", func() {
 				When("the deployment is in progress", func() {
-					When("last status change has a timestamp", func() {
+					When("last status change has a timestamp and max-in-flight is non-default", func() {
+						BeforeEach(func() {
+							summary = v7action.DetailedApplicationSummary{
+								Deployment: resources.Deployment{
+									Strategy:         constant.DeploymentStrategyRolling,
+									StatusValue:      constant.DeploymentStatusValueActive,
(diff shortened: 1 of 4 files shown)


pivotalgeorge:
rebasing on main after merge of #3096

weresch (changes requested):
issue (blocking): There was feedback on the v8 PR #3097 that applies here. Please include the additional commits from it.

command/v7/shared/app_summary_displayer_test.go
```diff
@@ -698,10 +698,58 @@ var _ = Describe("app summary displayer", func() {
 		When("there is an active deployment", func() {
 			var LastStatusChangeTimeString = "2024-07-29T17:32:29Z"
 			var dateTimeRegexPattern = `[a-zA-Z]{3}\s\d{2}\s[a-zA-Z]{3}\s\d{2}\:\d{2}\:\d{2}\s[A-Z]{3}\s\d{4}`
+			var maxInFlightDefaultValue = 1
```
a-b, line 701: thought (non-blocking): Should we keep constants in a single place [similar to this](https://github.com/pivotalgeorge/cf-cli/blob/ffa3f52edbee13a40228c66c3f1327a90020ca8c/integration/helpers/constants.go#L3-L4)

joaopapereira (approved):
LGTM

command/v7/shared/app_summary_displayer_test.go
pivotalgeorge: +1 to @joaopapereira's different PR suggestion if we do this. also, I don't think we should extract here because:
- `LastStatusChangeTimeString` and `dateTimeRegexPattern` (and not `maxInFlightDefaultValue`) are the type of helper constants that belong as siblings in the config @a-b linked.
- the three vars here are related to the tests in scope
- we'd be making readers jump around if we break them apart

In general I think we should extract when we see a need for reuse. It's unfortunate that the regex here is off-by-one from [this one in constants.go](https://github.com/pivotalgeorge/cf-cli/blob/ffa3f52edbee13a40228c66c3f1327a90020ca8c/integration/helpers/constants.go#L9)

pivotalgeorge:
updated with newer commit from #3097

joaopapereira (approved):
LGTM

PR #3099, opened 2024-08-13
[main] change WaitGroup to mutex

joaopapereira:
## Description of the Change

The WaitGroup is not behaving as expected, change it to a Mutex to lower the number of flake tests


cf/commands/application/start_test.go
```diff
@@ -152,8 +152,8 @@ var _ = Describe("start command", func() {
 		logRepo = new(logsfakes.FakeRepository)
 		logMessages.Store([]logs.Loggable{})
 
-		closeWait := sync.WaitGroup{}
```
Samze, line 155: **question/non-blocking** Any theory as to why the waitgroup doesn't work here and the mutex does 🤔 ? It looks fine.

joaopapereira: I do not see any particular reason why WaitGroup would not work here. There might be some sort of race condition or another strange thing.

a-b: The `cf` folder is extremely old (last piece of CAPI v2 in this repo) and should be removed as soon as we rework the plugin interface. We've avoided making any changes in this part of the codebase for the last five years.

joaopapereira: Going to resolve this since there as been no movement in the last week

PR #3112, opened 2024-08-15
Change the test to polling instead of just a single sleep [main]

joaopapereira:
## Description of the Change

- Avoid flakyness due to time it takes to bind a service

### DIFF
--- a/resources/droplet_resource.go
+++ b/resources/droplet_resource.go
@@ -1,12 +1,17 @@
 package resources
 
 import (
+	"encoding/json"
+
+	"code.cloudfoundry.org/cli/api/cloudcontroller"
 	"code.cloudfoundry.org/cli/api/cloudcontroller/ccv3/constant"
 )
 
 // Droplet represents a Cloud Controller droplet's metadata. A droplet is a set of
 // compiled bits for a given application.
 type Droplet struct {
+	// AppGUID is the unique identifier of the application associated with the droplet.
+	AppGUID string `json:"app_guid"`
 	//Buildpacks are the detected buildpacks from the staging process.
 	Buildpacks []DropletBuildpack `json:"buildpacks,omitempty"`
 	// CreatedAt is the timestamp that the Cloud Controller created the droplet.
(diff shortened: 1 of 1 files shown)


a-b (approved):
Love this idea, let's see if this helps.

weresch:
issue (blocking): I tried running this and found on occasion it failed, I think it's due to the order the apps can appear in the output. Sometimes APP2 comes first:
```
  OUT: Showing bound apps:
  OUT:    name                                        binding name                           status             message
  OUT:    APP2-1a263f4e-a731-4247-7fea-9dea533cdef0   ca204ee2-67cd-455e-68cf-afe0a557e805   create succeeded
  OUT:    APP1-cdd7753f-8c75-45e8-5cf7-37b5d063e594   e1c2f039-3950-48cf-613b-a09534ddf8d4   create succeeded
```
The test may need to be changed to validate that both appear but not rely on the order of them.

integration/v7/isolated/service_command_test.go
```diff
@@ -171,17 +171,16 @@ var _ = Describe("service command", func() {
...
-					Eventually(session).Should(Exit(0))
-
-					Expect(session).To(SatisfyAll(
-						Say(`Showing bound apps:\n`),
-						Say(`name\s+binding name\s+status\s+message\n`),
-						Say(`%s\s+%s\s+create succeeded\s*\n`, appName1, bindingName1),
-						Say(`%s\s+%s\s+create succeeded\s*\n`, appName2, bindingName2),
-					))
+					Eventually(func(g Gomega) {
+						session := helpers.CF(serviceCommand, serviceInstanceName, "-v")
```
weresch, line 175:
```suggestion
						session := helpers.CF(serviceCommand, serviceInstanceName, "-v").Wait()
```

integration/v7/isolated/service_command_test.go
```diff
@@ -171,17 +171,16 @@ var _ = Describe("service command", func() {
...
+						g.Eventually(session).Should(Exit(0))
```
weresch, line 176:
```suggestion
						g.Expect(session).Should(Exit(0))
```

integration/v7/isolated/service_command_test.go
```diff
@@ -171,17 +171,16 @@ var _ = Describe("service command", func() {
...
+						g.Eventually(session).Should(SatisfyAll(
```
weresch, line 177:
```suggestion
						g.Expect(session).Should(SatisfyAll(
```

integration/v7/isolated/service_command_test.go
```diff
@@ -171,17 +171,16 @@ var _ = Describe("service command", func() {
...
+							Say(`Showing bound apps:\n`),
+							Say(`name\s+binding name\s+status\s+message\n`),
+							Say(`%s\s+%s\s+create succeeded\s*\n`, appName1, bindingName1),
+							Say(`%s\s+%s\s+create succeeded\s*\n`, appName2, bindingName2),
+						))
+					}).WithPolling(5 * time.Second).WithTimeout(3 * time.Minute).Should(Succeed())
```
weresch, line 183:
```suggestion
					}).Should(Succeed())
```

integration/v7/isolated/service_command_test.go
weresch: Eventually blocks when called and attempts an assertion periodically until it passes or a timeout occurs. We can use Expect here and rely on Eventually on line 174 to retry

integration/v7/isolated/service_command_test.go
weresch: This will wait until the process exits before continuing

integration/v7/isolated/service_command_test.go
weresch: I don't think we need to specify polling and timeout here as Eventually on line 174 has some built in

weresch (changes requested):

integration/v7/isolated/service_command_test.go
joaopapereira: Is there any advantage of this approach vs the one that is in the PR?

integration/v7/isolated/service_command_test.go
weresch: Using .Wait() on the command will have it exit before continuing, so it can process the input and try again as needed

integration/v7/isolated/service_command_test.go
weresch: Oh just mentioned below, using .Wait() on the command will have it exit before continuing, so it can process the input and try again as needed

integration/v7/isolated/service_command_test.go
joaopapereira: What I mean is that After a failure if we retry 2 Milliseconds later there is a high probability that the status will not change.

joaopapereira: The majority of the Eventually that we use are done after an execution of a command to which we do not use the `.Wait()` function so it might make sense to have a polling interval that is lower than 5 seconds, but in cases where we want to retry to execution of the command itself, maybe it makes more sense to have a higher polling period, like in this case.
I think that we should configure the global timeouts and polling to the most common case in mind, but I believe the one from this comment is not it.

But if yll feel strongly about it we can create another PR where we do that change globally.

weresch: If we .Wait() for the command execution to complete on line 175, then the polling is limited to the time it takes the command to finish executing and validate output before it immediately tries again. I don't think there's an issue there, CAPI can handle that request rate. This means when the command has the correct output we'll validate it quickly and continue. If we have polling every few seconds, that's a few seconds where the command might be ready to provide correct output but we haven't checked.

integration/v7/isolated/service_command_test.go
```diff
@@ -171,17 +171,16 @@ var _ = Describe("service command", func() {
...
-					))
+					Eventually(func(g Gomega) {
+						session := helpers.CF(serviceCommand, serviceInstanceName, "-v").Wait()
+						g.Expect(session).Should(Exit(0))
+						g.Expect(session).Should(SatisfyAll(
+							Say(`Showing bound apps:\n`),
+							Say(`name\s+binding name\s+status\s+message\n`),
+							ContainSubstring(`%s\s+%s\s+create succeeded\s*\n`, appName1, bindingName1),
+							ContainSubstring(`%s\s+%s\s+create succeeded\s*\n`, appName2, bindingName2),
+						))
```
weresch, line 182:
```suggestion
						))
						sessionContents := string(session.Out.Contents())
						Expect(strings.Contains(sessionContents, fmt.Sprintf(`%s   %s   create succeeded`, appName1, bindingName1))).To(BeTrue())
						Expect(strings.Contains(sessionContents, fmt.Sprintf(`%s   %s   create succeeded`, appName2, bindingName2))).To(BeTrue())
```

weresch: I found the test can fail if app2 is listed first. If we get the session contents and search it for substrings the order doesn't matter.

weresch (changes requested):

integration/v7/isolated/service_command_test.go
joaopapereira: Stopped using the buffer

PR #3100, opened 2024-08-13
[v8] Max in flight status tests

weresch:
## Description of the Change

Adds integration tests to test max-in-flight deployment status changes from #3097

PR #3113, opened 2024-08-15
[main] Max in flight status tests

weresch:
## Description of the Change

* Add copy-source tests to verify max-in-flight output
* Add rollback tests to verify max-in-flight output
* Add restage tests to verify max-in-flight output
* Add restart tests to verify max-in-flight output
* Add push tests to verify max-in-flight output

This is the `main` branch equivalent for #3100

### DIFF
--- a/integration/v7/isolated/app_command_test.go
+++ b/integration/v7/isolated/app_command_test.go
@@ -272,20 +272,6 @@ applications:
 							Eventually(session1).Should(Exit(0))
 						})
 					})
-					When("the deployment is cancelled", func() {
-						It("displays the message", func() {
-							helpers.CF("restart", appName, "--strategy", "rolling")
-							Eventually(func() *Session {
-								return helpers.CF("cancel-deployment", appName).Wait()
-							}).Should(Exit(0))
-
-							Eventually(func(g Gomega) {
-								session := helpers.CF("app", appName).Wait()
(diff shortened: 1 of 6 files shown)


a-b:
Looks like a network issue. I retriggered int tests.

a-b:
re-running flaked int tests

a-b (approved):

integration/v7/isolated/app_command_test.go
```diff
@@ -272,20 +272,6 @@ applications:
 							Eventually(session1).Should(Exit(0))
 						})
 					})
-					When("the deployment is cancelled", func() {
```
joaopapereira, line 275: Any particular reason to remove this test? Are we testing this functionality in a different place?

joaopapereira (approved):
LGTM.
Added a question to make sure that the deletion of the test is not removing coverage.

integration/v7/isolated/app_command_test.go
weresch: There's a small window when `cf app` will report back a rolling deployment is canceling, and even between successive calls this can be missed which leads to a flakey test. I think this is covered by [unit tests for app_summary_displayer](https://github.com/cloudfoundry/cli/blob/main/command/v7/shared/app_summary_displayer_test.go#L825-L874).

PR #3119, opened 2024-08-17
Always show max-in-flight value for active deployments [v8]

weresch:
## Description of the Change

Always show max-in-flight value for active deployments
- show deployment strategy value on its own line


a-b (dismissed):
LGTM

a-b (approved):

PR #3132, opened 2024-08-22
[main] Fix linter workflow to fail when go fmt corrects files

joaopapereira:
## Description of the Change

Fix linter workflow to fail when go fmt corrects files


gururajsh (approved):
LGTM

PR #3155, opened 2024-08-28
Bump stack from cflinuxfs3 to cflinuxfs4 [main]

gururajsh:
Replace cflinuxfs3 stack references with cflinuxfs4
Use CAPI release version repository var if set. If not, get the latest CAPI version
Remove min CAPI jobs
Move ops-files from cli-ci to this repo


a-b (approved):

joaopapereira (approved):
LGTM.
Minor details to address in the future and not related specifically with this PR:
- Our examples code having real versions and stacks will force us to update them everytime they change which is a chore and maybe we can replace them with dummy values
- There is a lot of "cflinuxfs4" strings in our code, specially in tests, do we really need to have that many places where we set the stack?

PR #3156, opened 2024-08-28
Bump stack from cflinuxfs3 to cflinuxfs4 [v8]

gururajsh:
Replace cflinuxfs3 stack references with cflinuxfs4
Use CAPI release version repository var if set. If not, get the latest CAPI version
Remove min CAPI jobs
Move ops-files from cli-ci to this repo


command/v7/app_command_test.go
```diff
@@ -203,7 +203,7 @@ var _ = Describe("app Command", func() {
 						},
 					},
 					CurrentDroplet: resources.Droplet{
-						Stack: "cflinuxfs2",
```
a-b, line 206: question: I've noticed on multiple occasions both `cflinux2` and `cflinux3` were replaced with `cflinux4`. Do we know why we had two different stacks before?

a-b:
Please consider making separate PRs for unrelated changes like cleanup of min CAPI jobs and cflinux stack bumps moving forward.

command/v7/app_command_test.go
gururajsh: Some tests had them. Different branches had different values

joaopapereira (approved):
LGTM.
Minor details to address in the future and not related specifically with this PR:

Our examples code having real versions and stacks will force us to update them everytime they change which is a chore and maybe we can replace them with dummy values
There is a lot of "cflinuxfs4" strings in our code, specially in tests, do we really need to have that many places where we set the stack?

PR #3214, opened 2024-09-18
Remove redis ops-file [v8]

a-b:
Newer cf-deployment has valkey so this ops file is now redundant


a-b:
Failures in integration tests like [this](https://github.com/cloudfoundry/cli/actions/runs/10924736103/job/30324738515?pr=3214#step:11:24) are expected because we're using a destination branch workflow file for security reasons. I recommend proceeding with this merge.

Samze:
LGTM

PR #3241, opened 2024-10-03
Changes on wokflows and CVE checks

joaopapereira:
## Description of the Change

Do not run integration tests when different files/folders are the only change
Remove CVE check on integration/assets


a-b (approved):
LGTM

PR #3249, opened 2024-10-10
[PoC] Add --json flag to cli commands

joaopapereira:
## Description of the Change

This change adds the `--json` flags to the commands `apps`, `app`, and `get-health-check` commands as a PoC for us to understand how complicated this would be.


a-b (approved):
While it is debatable whether JSON belongs to UI, this is a good proof of concept.

Gerg:
I haven't reviewed this PR in detail, but a couple quick thoughts:
1. Is there a way we can handle this flag globally, rather than needing to implement it as a flag for every command? Similar to the `-v` flag or the environment-variable-based configuration.
1. What about `--output=json` instead of `--json` (inspired by [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html))?

joaopapereira:
> 1. Is there a way we can handle this flag globally, rather than needing to implement it as a flag for every command? Similar to the `-v` flag or the environment-variable-based configuration.

That was my first attempt but even when you look at the apps command you can see that the output is split into 3 different calls to the UI Writer, which makes it complicated to abstract out.
Another thing that I noticed is that in some cases changing the name of the keys, in the struct they are called "Name" but in the output we set it to "SpaceName" we also do translations which complicates even further the code.

> 2. What about --output=json instead of --json (inspired by [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html))?

I do not have a strong opinion about that, are we planning on adding TOML or YAML support, or are we just covering our basis here?

Issue #2291, opened 2022-07-07
Can't import code.cloudfoundry.org/cli v8.4.0


olyakrasnozhon:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
I need to start using v3 client instead of v2 in my terraform-provider-cf. After some investigation I see that the new resources and their CRUD APIs I need are placed in code.cloudfoundry.org/cli v8.4.0. 
I can't import this library directly, the error is - 
"go get: code.cloudfoundry.org/cli@v8.4.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v8".

I went over another libraries that I probably could use:
- https://github.com/cloudfoundry-community/go-cfclient (latest)
- https://github.com/cloudfoundry-community/cloudfoundry-cli (latest - v0.0.11-complete-api)
But they don't contains the newest APIs: for instance, go-cfclient contains the resource we need v3ServiceCredentialsBinding, but doesn't contain Create/Delete APIs for it. 

**What happened**
I need your help to find the way to import appropriate library to my terraform-provider-cloudfoundry in order to be able to use all the resources and APIs that implemented in code.cloudfoundry.org/cli@v8.4.0.

**Exact Steps To Reproduce**
1. In any of go-projects run "go get code.cloudfoundry.org/cli@v8.4.0"
2. Expected error is "go get: code.cloudfoundry.org/cli@v8.4.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v8".
3. Look for CreateServiceCredentialsBinding here:
- https://github.com/cloudfoundry-community/cloudfoundry-cli/tree/master/api/cloudcontroller/ccv3
- https://github.com/cloudfoundry-community/go-cfclient
- https://github.com/cloudfoundry/cli/tree/v8.4.0

**Notes regarding V6 and V7 CLI support:** 
- V6:
  - Minimum supported version of CF Deployment: v7.0.0 (CAPI Release: 1.74.0 (APIs 2.128.0 and 3.63.0))
[...]

PR #3256, opened 2024-10-18
upgrade module to v8 standard go version workflow

jcvrabo:

## Description of the Change

This change upgrades the module name according to the new go version workflow as described in https://go.dev/doc/modules/release-workflow

It also has a small fix in the app_summary_displayer_test.go date regex so that it supports up to 4 character time zone abbreviations

The fix is presented as a transitional one, the cli has dependency to the archived project https://github.com/cloudfoundry-incubator/cfnetworking-cli-api which doesn't use go modules and has a recursive dependency back to the cli. This makes it difficult to smoothly upgrade it in the future. The project now has a dependency of the latest vli version without modules (v7.1.0) which is used by the UAAClient to also create a legacy connection and then use it when creating the cf networking client. It's less than optimal.

The way forward is to remove the recursive dependency by either unarchiving the cfnetworking project (and removing the recursive dependency), creating a new library or adding the cf networking client directly to the cli code base. This workaround is a quick fix to move forward with standard versioning.

## Why Is This PR Valuable?

At the moment the development of plugins with go modules is dependent on cli v7.1.0 (which is still not a module). This implies that plugins cannot be updated with any security and version updates that may come from newer versions.

The PR will allow plugins to be updated to the latest cli version.

## Applicable Issues

https://github.com/cloudfoundry/cli/issues/2291

## How Urgent Is The Change?

It has a relative urgency, if you consider the need to update plugins as well as making this change as soon as possible for further progress of the cli

## Other Relevant Parties
  
To my knowledge this only affects plugins (and their maintainers/developers).


jcvrabo:
Hi, would like to know what's the status with reviewing these prs, the longer it takes the harder it will be to maintain it. I was updating it with the repository updates, until it started breaking the tests and I noticed that PRs were approved that themselves were breaking the tests.

The V2 api is getting phased out, and we have been testing that (disabling the v2 api) which ends up failing pretty much all plugins. These PRs became urgent, I would say, if you want to keep on supporting plugins... either that or we all have to start developing plugins without go modules... which is also not acceptable

I don't think I need to refer you to the v2 phasing out, but just in case

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0032-cfapiv2-eol.md
https://github.com/cloudfoundry/community/issues/1001

I will later try to resync the PRs with the head of the branches, but would still like to know what's the ETA for an approval here, as I also won't be fixing tests broken at the head of the branches, which will have the PR ending up with failed checks.

jcvrabo:
Following comment applies to all PR's related to updating the project to follow the standard go versioning workflow

https://github.com/cloudfoundry/cli/pull/3256
https://github.com/cloudfoundry/cli/pull/3260
https://github.com/cloudfoundry/cli/pull/3261

The initial hack based on using the previous non-module cli version that the cfnetworking project (archived) was using, was expected to break sooner or later. It broke sooner for the main and v8 branches as some projects had updates with non-backward compatible versions which basically broke the builds.

I have now copied (and made some small updates) the cfnetworking project into the api package. It was already followin pretty much the same implementation of the other api packages, so even though I don't exactly agree with the approach, I kept it to keep it coherent.

Apparently there was also another merged PR that broke the cf restage timeout test with rolling strategy after adding the --max-in-flight. I split it it two tests, one with the flag and one without, as the end result is different. I basically fixed it so that I could have the PR's green with the integration tests.

I would like to reiterate my plea for a review, as while the pr's are not complex they do touch a lot of files and consequentially will almost always end up having a conflicted status for every other merged PR. I do try to maintain the forked branches with upstream updates, but that also takes time and effort to do so, when the PR's are non-trivial.

theghost5800:
As far as I understood, this change must be merged in order to allow this go project to be used as dependency in other go modules, right?

If the answer of previous question is yes, why this change is not merged already, it's blocking all cf cli plugin maintainers to use latest v3 APIs models.

When this change will be merged or what are the requirements to merge?

jcvrabo:
The answer is yes to your question, but since I haven't gotten any feedback for some time, I stopped maintaining the fork and it now runs in conflict. When I get some positive feedback I'll update the fork with the latest commits and resolve the conflicts

jcvrabo:
tests seem to have failed due to a dns timeout, perhaps the tests could run again

Gerg:
This PR is so large that it's difficult to review natively in GitHub, so I'll post some thoughts here. I'm generally supportive of properly-modularizing the cf CLI, even if it's still not a supported usage.

### cfnetworking-cli-api

This PR's solution appears to be vendoring and modifying the repo: https://github.com/cloudfoundry/cli/pull/3256/commits/65e6e417f8ee2fb82124725e8c6d51936d4b7ade

As another option, we could write an adapter shim. For example: https://github.com/cloudfoundry/cli/blob/gerg-v8-mod/command/v7/shared/new_networking_client.go#L12-L27

In the long run, I think we should still remove the dependency on the ancient, archived module, but we could theoretically break that out as an independent change.

### Breaking Change

I go back and forth on whether properly-modularizing cf v8 is an acceptable breaking change:

- This shouldn't be a problem for folks using the cf CLI as a CLI binary
- BUT people de-facto import the CLI as a module, even if it's technically not supported (esp. for cf CLI plugins)
- Technically, you can't really currently consume the cf CLI as a proper module, since you run into a "major version" error (hence why this GitHub issue & PR exist)
- BUT you can get around it by referencing a commit sha
- So, a consumer would have to change their import paths to `.../cli/v8` when switching what sha they're using (or hopefully switching to tracking versions/tags instead)

The more aggressive option would be to make the change in v8, since it was the correct thing to do for go module major versions in the first place. Anyone consuming the cf CLI as a module is already kinda off-spec, so they could probably roll with the punches (or stay pinned to their current sha).

The more conservative option would be to cut a v9 that is just a properly-modularized v8, with no other breaking changes.
The "main" branch would then become "v10".

jcvrabo:
You are right in most accounts @Gerg 

Yes I'm putting back the cfnetworking library into the project (I'm saying putting back because when I look at cfnetworking I have the distinct impression it lived in the cf cli project in the beginning, it has the exact same structure, if it wasn't there... then indeed, I'm just vendoring it :) )

Yes, I totally agree that the cf networking client should be rewritten, or have that part simply using https://github.com/cloudfoundry/policy_client , for example. But I am basically focusing in modularizing the cli, I have always presented this as a transient change and all other structural changes would be brought afterwards (that would also require a more in-depth code/functionality analysis which I didn't do). The reason why this move was done as part of the modularization process is that cfnetworking is not only archived but it implements a plain anti-pattern of cyclic dependencies, so... cfnetworking has cf cli as a dependency, and that means that if you try to modularize v8 there will be a conflict in some cli packages using v8 and others using it without... so... the simpler solution was to move cfnetworking to cf cli and have its dependencies set to in-project (as I think it was in the beginning).

I do contest the concept of "breaking change", although it kind of is, the breaking part only affects the project itself and it would "break" all existing prs which would have to be updated. This is because either you're developing non-modularized plugins which could be affected by it (haven't really tested that) but I would expect most newer plugins to already be modularized and stuck on the incompatible 7.10 version... updating the module would not affect those plugins and when they would update to v8, it would be a straightforward update. Like you mention... cli is just not currently consumable as a module, so... no-one is in that scenario to be affected (and the versioning workflow also prevents that, as you have to update your version to v8 in order to consume it anyway), and updating should be straightforward.

I don't know when v9 is planned to come out, I'm fine with either option (updating v8 or just v9 and having it as a live version, a pr for main is also available), but regardless... this update will always be a big-bang approach, you can't do it in parts, and it's better to vendor code and keep legacy structure to have less changes to review... in summary... this basically updates all imports to start using v8 and changes cfnetworking dependency to a vendored one, it's a large pr but with little functional content.

Thanks for taking a look

gururajsh:
@jcvrabo branch was outdated, so updated it and rerunning the tests

jcvrabo:
thanks @gururajsh, I see two additional checks were added, will have to see what that will entail

gururajsh:
@jcvrabo 2 checks are failing. CVE one is a known issue. I have a PR open for that. I think codeQL one needs to be addressed.

jcvrabo:
I'll see about solving it

jcvrabo:
Hi @gururajsh, while analyzing the scan (wasn't acquainted with the tool and found it a bit confusing to trace the flow it complains about :) ), I come to the conclusion that they are not only false positives as I expect the same warnings to be raised in the cli repository as it is code that was not touched.

starting by the third error in command/translatableerror/convert_to_translatable_error.go, it complains about a possible leaked passwourd because of passing "sensitive information" in https://github.com/cloudfoundry/cli/blob/3f4eb70658a65e9881bef636576b056457829237/command/v7/auth_command.go#L160-L160 (passing the passwordMissing variable), however, this is just a boolean identifying that a password is missing and as such no sensitive information is being leaked. a waiver can be added for that finding, an alternative is changing the variable name to something else that doesn't have password in its name, but I don't think that should be donr as the variable name is self-explanatory

The other two warnings on cf/actors/plugininstaller/plugin_downloader.go and cf/commands/plugin/install_plugin.go converge on the same piece of code in https://github.com/cloudfoundry/cli/blob/3f4eb70658a65e9881bef636576b056457829237/cf/util/downloader/file_download.go#L39-L39

the source trace can be tracked to just cf/commands/plugin/install_plugin.go which takes an url (potentially, if a plugin is not installed by name through the cli plugin repo) from the command line to install , but that's the intention, that you provide a url to directly install the plugin. So I would say a waiver for these two can also be raised.

gururajsh:
@jcvrabo makes sense. @Gerg do you have anything else to discuss or are we good to go ahead with merge?

Gerg:
> @jcvrabo makes sense. @Gerg do you have anything else to discuss or are we good to go ahead with merge?

Good to go!

marckhouzam (approved):
I've used this change and can confirm I've had no problem using a v8 module to import the CLI

jcvrabo:
will this merge be promoted then?, I see @Gerg and @gururajsh gave the ok but are still tagged as pending for a review

PR #3261, opened 2024-10-21
[v7] upgrade module to v7 standard go version workflow

jcvrabo:

## Description of the Change

This change upgrades the module name according to the new go version workflow as described in https://go.dev/doc/modules/release-workflow

The fix is presented as a transitional one, the cli has dependency to the archived project https://github.com/cloudfoundry-incubator/cfnetworking-cli-api which doesn't use go modules and has a recursive dependency back to the cli. This makes it difficult to smoothly upgrade it in the future. The project now has a dependency of the latest vli version without modules (v7.1.0) which is used by the UAAClient to also create a legacy connection and then use it when creating the cf networking client. It's less than optimal.

The way forward is to remove the recursive dependency by either unarchiving the cfnetworking project (and removing the recursive dependency), creating a new library or adding the cf networking client directly to the cli code base. This workaround is a quick fix to move forward with standard versioning.

## Why Is This PR Valuable?

At the moment the development of plugins with go modules is dependent on cli v7.1.0 (which is still not a module). This implies that plugins cannot be updated with any security and version updates that may come from newer versions.

The PR will allow plugins to be updated to the latest cli version.

## Applicable Issues

https://github.com/cloudfoundry/cli/issues/2291

## How Urgent Is The Change?

It has a relative urgency, if you consider the need to update plugins as well as making this change as soon as possible for further progress of the cli

## Other Relevant Parties

To my knowledge this only affects plugins (and their maintainers/developers).


Samze:
V7 of the CLI is out of support and no longer being released https://github.com/cloudfoundry/cli/wiki/Versioning-and-Support-Policy. Closing this PR.

PR #3268, opened 2024-10-28
Automate release version generation [v8]

gururajsh:
While releasing CLI, take the input from the user for minor/patch bump. Based on the selection, calculate the next release version. New build version need not be stored in a file going forward.


Makefile
```diff
@@ -3,7 +3,6 @@ NODES ?= 10
 PACKAGES ?= api actor command types util version integration/helpers
 LC_ALL = "en_US.UTF-8"
 
-CF_BUILD_VERSION ?= $$(cat BUILD_VERSION) # TODO: version specific
```
a-b, line 6: question(non-blocking): do we still see some value in keeping [build_data.yml](https://github.com/cloudfoundry/cli/blob/main/build_data.yml) up to date? As it was introduced as an eventual alternative to BUILD_VERSION.

a-b (dismissed):

bin/bump-version
a-b, line 1: praise: yay, removing old code!

Makefile
gururajsh: Good point, even that file can be removed. It was mostly used for min capi tests. Since we don't have any, will remove that as well

gururajsh: Done

joaopapereira (approved):
LGTM

PR #3270, opened 2024-10-28
Switch test environments to cfd bosh lites [v8]

a-b:
## Description of the Change

* Switch environment to cfd-bosh-lite
* Turn off set_kernel_parameters to use cfd-bosh-lite
* Expose shepherd configuration via repo variables
* Update secrets path to bosh-lite
* Expose TEST_FLAKE_ATTEMPTS and TEST_NODES
* Set default FLAKE_ATTEMPTS
* Bump gha test suit timeout from 60 to 120m
* Enter nodes as workflow_dispatch input
* Skip steps if SHEPHERD_LEASE_ID defined
* Skip scaling tests that are testing CAPI
* bosh-light environment default memory is 256Mb
* Fixed the issue with force update of bosh
* Delete created orgs during tests
* Ensure that the orgName is empty for each test
* Prevent double deletion of org
* Lower number of tests running using client credentials to avoid flakes
* Allow user to define namespace of the leased environment
* Pass namespace of the lease to the reusable workflow

### DIFF
--- a/integration/v7/isolated/start_command_test.go
+++ b/integration/v7/isolated/start_command_test.go
@@ -128,7 +128,7 @@ var _ = Describe("start command", func() {
 					Eventually(session).Should(Say(`routes:\s+%s.%s`, appName, helpers.DefaultSharedDomain()))
 					Eventually(session).Should(Say(`type:\s+web`))
 					Eventually(session).Should(Say(`instances:\s+1/1`))
-					Eventually(session).Should(Say(`memory usage:\s+1024M`))
+					Eventually(session).Should(Say(`memory usage:\s+256M`))
 					Eventually(session).Should(Say(`\s+state\s+since\s+cpu\s+memory\s+disk\s+logging\s+cpu entitlement\s+details`))
 					Eventually(session).Should(Say(`#0\s+(starting|running)\s+\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z`))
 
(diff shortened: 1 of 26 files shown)


integration/v7/isolated/start_command_test.go
```diff
@@ -128,7 +128,7 @@ var _ = Describe("start command", func() {
 					Eventually(session).Should(Say(`routes:\s+%s.%s`, appName, helpers.DefaultSharedDomain()))
 					Eventually(session).Should(Say(`type:\s+web`))
 					Eventually(session).Should(Say(`instances:\s+1/1`))
-					Eventually(session).Should(Say(`memory usage:\s+1024M`))
+					Eventually(session).Should(Say(`memory usage:\s+256M`))
 					Eventually(session).Should(Say(`\s+state\s+since\s+cpu\s+memory\s+disk\s+logging\s+cpu entitlement\s+details`))
```
a-b, line 132: TODO: double-check correctness

a-b (commented):
LGTM

joaopapereira (approved):
LGTM

a-b:
Integration tests are expected to fail because they use previous workflow definitions that are getting old env from the pool [here](https://github.com/cloudfoundry/cli/actions/runs/11578122464/job/32231356086?pr=3270#step:4:42). Since new bosh-lite uses different memory defaults 256Mb instead of 1G it will fail.

Issue #3274, opened 2024-10-31
cf tasks <app_name> returns all tasks which is problematic at scale


Samze:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**

[Tasks](https://v3-apidocs.cloudfoundry.org/version/3.165.0/index.html#tasks) allow you to run one-off commands on an app. Below is an example workflow in the CLI:

1. Trigger the task

```
cf run-task app -c <my-command>
Creating task for app static1 in org org-1 / space space-1 as admin...
Task has been submitted successfully for execution.
OK

task name:   b0a3bce1
task id:     1309
OK
```

2. Obtain result of task to see if it was it successful or not.
```
cf tasks static1
...
17     c25fabef   SUCCEEDED   Wed, 30 Oct 2024 16:42:01 UTC   ls
16     c76d2713   SUCCEEDED   Wed, 30 Oct 2024 16:42:01 UTC   ls
15     556f415b   SUCCEEDED   Wed, 30 Oct 2024 16:42:00 UTC   ls
14     c50052eb   SUCCEEDED   Wed, 30 Oct 2024 16:42:00 UTC   ls
13     9afdc979   SUCCEEDED   Wed, 30 Oct 2024 16:41:59 UTC   ls
12     9c52c14c   SUCCEEDED   Wed, 30 Oct 2024 16:41:48 UTC   ls
11     70d94489   SUCCEEDED   Wed, 30 Oct 2024 16:41:47 UTC   ls
10     4e24e204   SUCCEEDED   Wed, 30 Oct 2024 16:41:46 UTC   ls
9      afd1c3cc   SUCCEEDED   Wed, 30 Oct 2024 16:41:46 UTC   ls
8      2344f470   SUCCEEDED   Wed, 30 Oct 2024 16:41:45 UTC   ls
7      e6e83064   SUCCEEDED   Wed, 30 Oct 2024 16:41:45 UTC   ls
6      fdd65417   SUCCEEDED   Wed, 30 Oct 2024 16:41:44 UTC   ls
5      78af57a5   SUCCEEDED   Wed, 30 Oct 2024 16:41:43 UTC   ls
[...]

PR #3275, opened 2024-11-01
[main] Make retrieval of tasks in bulk instead of small increments

joaopapereira:
## Description of the Change

In this PR we change the way tasks are retrieved in CLI to lower the load that it currently has in CAPI when there are a big number of tasks associated with a particular application

Related to #3274


a-b (approved):
Looks like a good improvement. What other commands could benefit from this technique?

PR #3282, opened 2024-11-04
[main] Remove fixtures/applications from grype scan

joaopapereira:
## Description of the Change

Grype is looking at a folder that is just used as a testing fixture, this PR will remove it


gururajsh (approved):
LGTM

Issue #3248, opened 2024-10-10
Create user-provided-service is not idempotent when name is already taken


sethboyles:
**Please fill out the issue checklist below and provide ALL the requested information.**

- [x] I reviewed open and closed github issues that may be related to my problem.
- [x] I tried updating to the latest version of the CF CLI to see if it fixed my problem.
- [x] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [x] I am reporting a bug that others will be able to reproduce.

**Describe the bug and the command you saw an issue with**
Trying to create a user-provided-service that already exists fails with exit code 1 instead of succeeding (like most other commands, such as `create-space` and `create-user`).  I seems like there have been previous efforts to make CF commands idempotent: https://github.com/cloudfoundry/cli/pull/2264

**What happened**
```
$ cf create-user-provided-service myservice
Creating user provided service myservice in org org-1 / space org-1-space-1 as admin...
The service instance name is taken: myservice.
FAILED
# exit code 1
```

**Expected behavior**
```
$ cf create-user-provided-service myservice
Creating user provided service myservice  in org org-1 / space org-1-space-1 as admin...
The service instance already exists: myservice.
OK
# exit code 0
```

**Exact Steps To Reproduce**
Steps to reproduce the behavior; include the exact CLI commands and verbose output:
1. Run `cf create-user-provided-service myservice`
1. Run `cf create-user-provided-service myservice` again

gururajsh:
Fix for this is in and will be available in the next release.

krismarc:
@sethboyles, @gururajsh, @a-b That was actually breaking change for anyone who used this error code to proceed with logic in the script.

PR #3288, opened 2024-11-05
Make cups command idempotent [main]

gururajsh:
When a service name is already taken, create-user-provided-service command fails as mentioned in GH Issue https://github.com/cloudfoundry/cli/issues/3248. This PR fixes that.


joaopapereira (approved):
LGTM

PR #3291, opened 2024-11-08
Optimize units workflow [v8]

a-b:
- Added Go module caching which will significantly speed up builds by reusing downloaded dependencies
- Combined Windows and non-Windows jobs into a single matrix job with conditional steps
- Added dynamic shell selection based on OS
- Simplified overall workflow structure while preserving all functionality


gururajsh (approved):
LGTM

PR #3299, opened 2024-11-15
[main] Add new release validation

joaopapereira:
## Description of the Change

Update the version of Ubuntu used to validate the new releases. This was done because when we changed the binaries location from AWS to Github we had a problem with APT and how it handles the URL Github provides. This is fixed in versions of Ubuntu 22.04 and newer.
We added a new workflow that we can run manually to check if the version can be retrieved to help us diagnose any possible problems in the future.


a-b (approved):
LGTM

PR #3306, opened 2024-11-18
Add nonbase64 sha256 support [main]

gururajsh:
CLI currently only supports base64 encoded sha256 fingerprints. This PR adds non base64 sha256 support.


a-b (approved):
LGTM

Issue #3314, opened 2024-11-21
Implement per-route options in the CF CLI


maxmoehl:
**What's the user value of this feature request?**

Allow users to manage per-route options via the CLI instead of having to talk to the API directly.

**Who is the functionality for?**

Users of the per-route features.

**How often will this functionality be used by the user?**

Unknown.

**Who else is affected by the change?**

The change is not breaking and optional.

**Is your feature request related to a problem? Please describe.**

No.

**Describe the solution you'd like**

Implement the RFC specification.

Specifically I propose that we add a new command `update-route` which allows updates to the route resource. This should be implemented in a generic way to not require a change in the CLI when adding a new per-route feature. I have something like this in mind:

```
$ cf update-route -h
NAME:
   update-route - Update an existing route.

USAGE:
   cf update-route DOMAIN [--hostname HOSTNAME] [--path PATH] [--option OPTION=VALUE] [--remove-option OPTION]

EXAMPLES:
   cf update-route example.com -o lb_algo=round-robin
   cf update-route example.com -o lb_algo=least-connections
   cf update-route example.com -r lb_algo

OPTIONS:
   --hostname, -n      Hostname for the HTTP route (required for shared domains)
   --path              Path for the HTTP route
   --option -o         Set the value of a per-route option, key-value pairs, repeat to set multiple options
   --remove-option -r  Unset a previously set option

SEE ALSO:
   create-route, map-route, routes, unmap-route
```

Plus a link to the documentation describing the currently implemented options, probably we could link directly to the API spec which has to list the available options.

**Describe alternatives you've considered**

We could implement each field as a separate flag but that would require users to update the CLI to use new per-route options.

**Additional context**

* [RFC0027 Generic Per-Route Features](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0027-generic-per-route-features.md)
* [Tracking Issue](https://github.com/cloudfoundry/community/issues/909)

peanball:
There are some questions / topics that need decisions:



## Do we use `cf update-route` as new command, or `cf map-route` as existing command?

Options should be able to be set in both of those.

That said, the options should be set in various commands anyway. From the existing commands:
- `create-route`
- `map-route`

Additionally, we could add `update-route` to update general parameters about the route.

The options don't need to be available on `unmap-route`, as there we just identify the route to unmap.

## User Interface in the CLI

The CLI shows route information in various places.

### `cf routes` - shows the routes

Here we could add a list of the options to the output.

[...]

PR #3328, opened 2024-12-09
RFC0027 CLI command "update-route" to manage generic per-route options

Dariquest:

Extending the API of the Route to contain generic options map.

```
$ cf update-route -h
NAME:

USAGE:

EXAMPLES:

OPTIONS:

SEE ALSO:
```
It should be possible to provide options in the create-route and map-route commands.

```
EXAMPLES:
```

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

No urgency


cf/api/resources/routes.go
```diff
@@ -15,8 +15,14 @@ type RouteEntity struct {
...
 	Apps            []ApplicationResource   `json:"apps"`
 	ServiceInstance ServiceInstanceResource `json:"service_instance"`
+	Options         map[string]string       `json:"options"`
 }
 
+/*type RouteOptions struct {
+	LoadBalancing string `json:"load_balancing"`
+	SessionCookie string `json:"session_cookie"`
+}*/
+
```
Dariquest, line 25: This is left on purpose to trigger a discussion about the type formalization. 
Should we select a struct instead of a map? The map is currently selected, since it is easier to set a value of the map and we could add additional options easier. 
Option validation would be easier when using a struct.

maxmoehl: I'd stick to a map. We initially thought to add validation in the client to improve the UX but this will then require users to upgrade the CLI to make use of the new fields which seems unnecessary.

Dariquest: Thanks, I would also prefer that at this moment.

Dariquest:
Replacing by the https://github.com/cloudfoundry/cli/pull/3338 into the appropriate CF CLI version V8 branch

PR #3338, opened 2024-12-19
RFC0027 CLI command "update-route" to manage generic per-route options

Dariquest:

Extending the API of the Route to contain generic options map.

```
$ cf update-route -h
NAME:

USAGE:

EXAMPLES:

OPTIONS:

SEE ALSO:
```
It should be possible to provide options in the create-route and map-route commands.

```
EXAMPLES:
```

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

No urgency


MarcPaquette:
FYI @gururajsh

a-b (changes requested):
Thank you for your contribution. However, we do not make changes in the `cf` folder because it predates version 6 and is marked as deprecated. We hope to remove it soon, as plugins are the only reason it remains in the code.

a-b:
Also, we noticed that you've made your original request against the `main` branch - as it should be, but later [closed](https://github.com/cloudfoundry/cli/pull/3328#issuecomment-2554275438). What was your motivation for this change?

Dariquest:
> Also, we noticed that you've made your original request against the `main` branch - as it should be, but later [closed](https://github.com/cloudfoundry/cli/pull/3328#issuecomment-2554275438). What was your motivation for this change?

Hi @a-b, I was advised in the cli channel, that I should develop against the version branches. The new command is intended for the v8 and not for the v9 (as far as I understood from the contribution guide, main will become v9 later, and v8 is the target version if I intend to add a v8 command). Have I got it wrong?

Dariquest:
A new PR was created to leave the command folder untouched https://github.com/cloudfoundry/cli/pull/3366

PR #3340, opened 2024-12-20
cf revision command implementation [v8]

gururajsh:
This PR implements the revision command which the revision details of an app when given a revision version. The revision command is no longer experimental.


joaopapereira (approved):
LGTM

PR #3341, opened 2024-12-20
cf revision command implementation [main]

gururajsh:
This PR implements the revision command which the revision details of an app when given a revision version. The revision command is no longer experimental.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -20,6 +20,7 @@ on:
       - "doc/**"
       - ".gitpod.yml"
       - "README.md"
+  workflow_dispatch:
 
 jobs:
   golangci:
--- a/command/v7/revision_command.go
+++ b/command/v7/revision_command.go
@@ -1,8 +1,13 @@
 package v7
 
 import (
-	"code.cloudfoundry.org/cli/command"
+	"fmt"
+	"strconv"
+
+	"code.cloudfoundry.org/cli/actor/v7action"
 	"code.cloudfoundry.org/cli/command/flag"
+	"code.cloudfoundry.org/cli/resources"
+	"code.cloudfoundry.org/cli/types"
 )
 
 type RevisionCommand struct {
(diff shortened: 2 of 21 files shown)


command/v7/revision_command.go
```diff
@@ -14,7 +19,118 @@ type RevisionCommand struct {
...
+func (cmd RevisionCommand) displayMetaData(data map[string]types.NullString) {
+	tableData := [][]string{}
+	for k, v := range data {
+		tableData = append(tableData, []string{fmt.Sprintf("%s:", k), v.Value})
+	}
+	cmd.UI.DisplayKeyValueTable("", tableData, 3)
+
+}
+
+func revisionDeployed(revision resources.Revision, deployedRevisions []resources.Revision) bool {
```
joaopapereira, line 129: Any reason for this function to be the only one that is not associated with the struct?

command/v7/revision_command.go
```diff
@@ -14,7 +19,118 @@ type RevisionCommand struct {
...
+	labels := revision.Metadata.Labels
+	cmd.UI.DisplayWarnings(warnings)
+	if err != nil {
+		return err
+	}
+
+	if len(labels) > 0 {
+		cmd.displayMetaData(labels)
+		cmd.UI.DisplayNewline()
+	}
```
joaopapereira, line 77: *Nit*: Given that we are repeating this in different places I think that maybe the `displayMetaData` function should be the one that has this if statement and the `Execute` function would always call the function.

command/v7/revision_command.go
```diff
@@ -14,7 +19,118 @@ type RevisionCommand struct {
...
+	cmd.UI.DisplayWarnings(warnings)
+	if err != nil {
+		return err
+	}
+
+	revision, warnings, err := cmd.Actor.GetRevisionByApplicationAndVersion(
+		app.GUID,
+		cmd.Version.Value,
+	)
+	cmd.UI.DisplayWarnings(warnings)
```
joaopapereira, line 58: *Question:* If we display the warnings in this line, are we not going to repeat them in lines 69 and 81 as well?

joaopapereira (commented):
The PR looks good in general, added a nit and a couple of questions that we should get answered before merging

command/v7/revision_command.go
gururajsh: Fixed

command/v7/revision_command.go
gururajsh: Fixed

.github/workflows/golangci-lint.yml
```diff
@@ -20,6 +20,7 @@ on:
       - "doc/**"
       - ".gitpod.yml"
       - "README.md"
+  workflow_dispatch:
```
a-b, line 23: **Question**: Are we not obtaining linting results for the current state of the repository with each push or pull request? While I can see why manual triggering can be beneficial for addressing flaky tests, is that necessary for linter?
```suggestion
```

a-b (approved):
LGTM

joaopapereira (approved):
LGTM

PR #3364, opened 2025-01-15
[v8] Newer version of the scan action changed the output location for results

joaopapereira:
## Description of the Change
The new version of the CVE scan action changed the location of the results file causing errors in that workflow for us


a-b (approved):
LGTM

PR #3366, opened 2025-01-17
RFC0027 CLI Support for Generic Per-Route Options [v8]

Dariquest:

Extending the API of the Route to contain per route options and particularly the first one called "loadbalancing", containing a load balancing algorithm of a route. 

Existing commands "create-route", "map-route", "routes", "route", "apps", "app" were also extended to accept and output the per-route options.

```
$ cf update-route -h
NAME:

USAGE:

EXAMPLES:

OPTIONS:

SEE ALSO:
```
It is possible to provide options in the create-route and map-route commands.

```
EXAMPLES:
```

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

No urgency


resources/options_resource.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	err := cloudcontroller.DecodeJSON(data, &ccRouteOptions)
+	if err != nil {
+		return err
+	}
+
+	opts.RouteOption.LoadBalancing = ccRouteOptions.RouteOptions.Loadbalancing
+	return nil
+}
```
Dariquest, line 44: @maxmoehl I went for typed route options and implemented marshalling and unmarshalling to prevent unnecessary requests to the CC with the wrong option keys. The option key validation is then performed on the CLI side

maxmoehl: Let's not split the discussion any further and keep it in #3314.

Dariquest:
Hi team,
the linter fails in the PR, but locally it does not identify any issues:
golangci-lint run --exclude-dirs cf --exclude-dirs fixtures --exclude-dirs plugin --exclude-dirs command/plugin -v

```
INFO [config_reader] Used config file .golangci.json 
INFO [lintersdb] Active 6 linters: [errcheck goimports govet ineffassign staticcheck unused] 
INFO [loader] Go packages loading at mode 575 (types_sizes|files|exports_file|imports|name|compiled_files|deps) took 13.93931s 
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 185.93025ms
```

a-b (approved):

PR #3367, opened 2025-01-17
RFC0027 Enable "update-route", "create-route", "map-route", "routes", "route", "apps" & "app" to manage generic per-route options

Dariquest:

Extending the API of the Route to contain per route options and particularly the first one called "loadbalancing", containing a load balancing algorithm of a route. 

Existing commands "create-route", "map-route", "routes", "route", "apps", "app" were also extended to accept and output the per-route options.

```
$ cf update-route -h
NAME:

USAGE:

EXAMPLES:

OPTIONS:

SEE ALSO:
```
It is possible to provide options in the create-route and map-route commands.

```
EXAMPLES:
```

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

No urgency


Dariquest:
Wrong branch selected

PR #3379, opened 2025-01-27
Unit test: Fix regex for timezone parsing [v8]

hoffmaen:
## Description of the Change

The unit tests failed on my local system with time zone `CEST`, as the output is checked against a regex expression that only accepts time zones with three characters. This pull-request allows time zones with three or four characters.

## Why Is This PR Valuable?

Unit tests will run on CEST systems.

## Applicable Issues

none

## How Urgent Is The Change?

Not urgent, it only fixes test issues on local systems.

## Other Relevant Parties

none


joaopapereira (approved):
LGTM

PR #3385, opened 2025-01-28
Use cf-deployment main branch to for enable v2 api ops file [v7]

gururajsh:
Currently, if a cf-d version is pinned for integration workflow run, we check out that version's branch of cf-deployment. But enable-v2-api.yml ops-file is available only in main branch of cf-deployment. Due env issues, we need to cf-d version to v44.4 for now. So the workflow fails. This change checks out the main branch of cf-deployment explicitly.


joaopapereira (approved):
LGTM

Issue #3368, opened 2025-01-17
Allow CF Authentication based on Tokens - user and client tokens


strehle:
**Please provide details on the following items. Failure to do so may result in deletion of your feature request.** 

Request in short:
Enhance the cf auth command with a parameter --token or --assertion (the name can be defined later). The content of this token should be either a user token in order to perform a jwt-bearer or a client token in order to perform a client_credentials grant with federated trust.

**What's the user value of this feature request?**
Security improvement because we can omit secrets and and we can finally omit passwords. 

**Who is the functionality for?**
CF Authentication. "cf auth" command

Example
With 
`cf auth --assertion <jwt-token> `we can replace `cf auth USERNAME PASSWORD --origin`

Use jwt bearer instead of password grant flow. JWT bearer does not need to have an origin because the origin is in the token

and With
`cf auth --assertion <jwt-token> --client-credentials ` we can replace  `cf auth CLIENT_ID CLIENT_SECRET --client-credentials`

Use federated client credential flow. Trust between external system and CF must be setup before.

Who will use this feature? 

Enterprise users.
All other users who dont want setup their password in a command line tool, but currently must use --sso 

**How often will this functionality be used by the user?**
Each authentication of users and clients. With this we can link github to CF and we can map a github action authentication to a CF client so that you can do cf push within a github action. (you can do this currently but now you need to store CF secret in github)

**Who else is affected by the change?**
No breaking change but an enhancement only

**Is your feature request related to a problem? Please describe.**
The problem, that you need to do --sso for users and you dont have a solution for workload identity authentication, means linking a JWT provider to CF. K8s could be linked to CF.

**Describe the solution you'd like**
See examples with cf auth , Allow cf login based on a token from another Identity Provider without using a password or secret.

**Describe alternatives you've considered**
<na>

**Additional context**
https://github.com/cloudfoundry/uaa/issues/2838
https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token

strehle:
> And as discussed in slack: It would be really usefull to keep the option to have CLIENT_ID in the command to be able to use custom clients (otherwise cf client would be used implicetly if I understand correctly):
> 
> `cf auth CLIENT_ID --assertion <jwt-token> --client-credentials`

[...]

PR #3397, opened 2025-02-11
[main] Allow CF Authentication based on Tokens - user and client tokens

strehle:
## Description of the Change

Enhance the cf auth command with a parameter --assertion. The content of this token should be either a user token in order to perform a jwt-bearer or a client token in order to perform a client_credentials grant with federated trust.

UAA supports JWT bearer since UAA 4.5.0 , see https://docs.cloudfoundry.org/api/uaa/version/77.25.0/index.html#jwt-bearer-token-grant
UAA support the federated client credential flow since 77.25.0

## Why Is This PR Valuable?

CF can be integrated into Github Action without any extra secret setup in Github Repo.
Customer can then decide about using external tokens like github action token for user and/or client authentication.

In a PR you retrieve a id_token from gh action, this can be passed with cf auth --assertion <github id token> so that you are authentication in or to do a cf push ...

## Applicable Issues

* https://github.com/cloudfoundry/cli/issues/3368

## How Urgent Is The Change?

* it is an enhancement but it solves security issues, because CF integrations need to omit secrets and/or client certificates, but integration of github action with CF is only possible if you store a secret in Github

## Other Relevant Parties

Only CF landscapes with a configured trust to external OIDC parties

### DIFF
--- a/actor/v7action/auth.go
+++ b/actor/v7action/auth.go
@@ -25,7 +25,7 @@ func NewDefaultAuthActor(config Config, uaaClient UAAClient) AuthActor {
 }
 
 func (actor defaultAuthActor) Authenticate(credentials map[string]string, origin string, grantType constant.GrantType) error {
-	if grantType == constant.GrantTypePassword && actor.config.UAAGrantType() == string(constant.GrantTypeClientCredentials) {
+	if (grantType == constant.GrantTypePassword || grantType == constant.GrantTypeJwtBearer) && actor.config.UAAGrantType() == string(constant.GrantTypeClientCredentials) {
 		return actionerror.PasswordGrantTypeLogoutRequiredError{}
 	}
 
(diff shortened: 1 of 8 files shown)


strehle:
@a-b Can you give me a timeline about when somebody checks the PR ?  There is one failing test but I dont see a relation to my changes.

The extension is for https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0037-deprecate-passwords.md#cf-cli

I will provide PRs for the other repos and also for cf login after that but first I would like to know if you accept this enhancments

Thanks in advance

a-b:
Hi. I'll look into this shortly.

a-b (approved):

a-b:
We are currently only releasing from the v8 branch. @strehle, are you planning to create a PR for v8?

strehle:
> We are currently only releasing from the v8 branch. @strehle, are you planning to create a PR for v8?

Ok, yes, sure. Thanks. That was now faster than expected. Will do downports to v8.
V7 is out of scope ?

I maybe will also add it to cf login, but step by step.

a-b:
Yes v7 is out of support.

PR #3400, opened 2025-02-12
[main] Restart canary steps

joaopapereira:
## Description of the Change

Add `--instance-steps` to the restart command


integration/v7/isolated/continue_deployment_test.go
```diff
@@ -89,15 +89,29 @@ var _ = Describe("Continue Deployment", func() {
 
 		Context("when the continue is successful", func() {
 			When("There is a canary deployment", func() {
-				It("succeeds", func() {
-					helpers.WithHelloWorldApp(func(appDir string) {
-						helpers.CF("push", appName, "-p", appDir, "--strategy=canary").Wait()
```
a-b, line 94: **question**: are we still testing scenario where `--strategy=canary`, but `--instance-steps=` is not provided?

joaopapereira: I think we do have tests that test that

Samze (approved):

PR #3401, opened 2025-02-12
[main] Address new filepath-securejoin constraint

a-b:
more details https://github.com/cyphar/filepath-securejoin/releases/tag/v0.4.1


moleske (approved):
Tested by copying fix to broken dependabot branch.  Not sure why the fix wasn't just made on the dependabot branch but will merge and then tell dependabot to rebase those branches

PR #3419, opened 2025-02-19
[v7] add missing NetworkPolicyV1Endpoint field to Data struct

metskem:
## Description of the Change

There is a NetworkPolicyV1Endpoint field missing in a Data struct, this change adds that single field
 
## Why Is This PR Valuable?

Currently the NetworkPolicyV1Endpoint field in your $CF_HOME/config.json file gets wiped out if you execute a cf plugin subcommand while your AccessToken is expired. Your AccessToken is renewed and your config.json file is rewritten, but then the NetworkPolicyV1Endpoint is gone, which then causes your next "cf network-policies" command to fail.
You have to either manually edit the config.json file and add the field back, or do a "cf login ...." again.
Mind that this only happens if using plugin subcommands, not with the builtin cf subcommands like cf apps, cf services etc.

## Applicable Issues

None

## How Urgent Is The Change?

No, just a bug under specific circumstances.

## Other Relevant Parties

Anyone using plugins that need the cf AccessToken.

### DIFF
--- a/cf/configuration/coreconfig/config_data.go
+++ b/cf/configuration/coreconfig/config_data.go
@@ -31,6 +31,7 @@ type Data struct {
 	LogCacheEndPoint         string
 	MinCLIVersion            string
 	MinRecommendedCLIVersion string
+	NetworkPolicyV1Endpoint  string
 	OrganizationFields       models.OrganizationFields
 	PluginRepos              []models.PluginRepo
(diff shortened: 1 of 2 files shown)


a-b (approved):

Samze:
V7 of the CLI is out of support and no longer being released. Closing this PR.

PR #3426, opened 2025-02-25
[v8] Fix update repos

a-b:
Since linux awscli package is now deprecated and removed, we're switching to the official way of installing aws cli tools via pip.


joaopapereira (approved):
Lgtm

PR #3429, opened 2025-02-26
[main] feat: add ready indicator to `cf app` command

joaopapereira:
## Description of the Change

CC version 3.144.0 introduced readiness checks and with them a new status field for processes: routable. This commit exposes that new field in the `cf app` command to allow users to easily determine the state of their application instances.

See: https://github.com/cloudfoundry/capi-release/releases/tag/1.158.0


a-b (approved):
LGTM

PR #3433, opened 2025-02-27
Fix null canary object

Samze:
* Do not send null canary object if no steps have been specified

## Description of the Change

CAPI validates that `options.canary` can only be set for strategy canary. See https://github.com/cloudfoundry/cloud_controller_ng/blob/78d93d458dad934ce0f21039f63ddc944fb5e0c9/app/messages/deployment_create_message.rb#L84-L86


joaopapereira (approved):
LGTM

a-b (approved):

PR #3450, opened 2025-03-11
Update runner image to the latest ubuntu

a-b:
- Use dpkg-deb -Zxz to keep backward compatibility
- Improve test to check compatibility with Deb OSes

### DIFF
--- a/.github/workflows/release-build-sign-upload.yml
+++ b/.github/workflows/release-build-sign-upload.yml
@@ -114,7 +114,7 @@ jobs:
     name: Build Linux
     needs:
     - setup
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     env:
       VERSION_BUILD: ${{ needs.setup.outputs.version-build }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-build-sign-upload.yml
```diff
@@ -905,20 +905,29 @@ jobs:
...
       uses: actions/download-artifact@v4
       with:
         name: cf-cli-linux-deb-packages
 
-    - name: Display structure of downloaded files
+    - name: Display structure of downloaded files and test compression
       run: |
         ls -R
+
+        for i in debian:11 debian:12 ubuntu:20.04 ubuntu:22.04 ubuntu; do docker run \
```
Samze, line 920: *question* I dont have context on the compression issue, but how does this test it?

a-b: `dpkg --info` fails when trying to open `.deb` file with unsupported compression

a-b: 👁️  https://github.com/cloudfoundry/cli/pull/3450#discussion_r1989991255

a-b: example https://github.com/a-b/cli/actions/runs/13781247233/job/38539815075#step:3:30

joaopapereira: Can we use the matrix strategy here to change the container image? instead of this for loop?

.github/workflows/release-build-sign-upload.yml
```diff
@@ -905,20 +905,29 @@ jobs:
...
 
     - name: Download Signed Linux Packages
       uses: actions/download-artifact@v4
       with:
         name: cf-cli-linux-deb-packages
 
-    - name: Display structure of downloaded files
+    - name: Display structure of downloaded files and test compression
       run: |
         ls -R
```
joaopapereira, line 918: Do we still need this ls?

joaopapereira (approved):
I am going to approve this PR due to the urgency to get this merged so we can release but I would love if we could answer the 2 questions I left and also create a subsequent PR to address them.

.github/workflows/release-build-sign-upload.yml
a-b: Good call. Yes, we should use matrix!

.github/workflows/release-build-sign-upload.yml
a-b: Good call. Yes, we should use matrix!

PR #3455, opened 2025-03-15
[v8] Allow CF Authentication based on Tokens - user and client tokens 

strehle:
(cherry picked from commit de832082898835f39409aca03be947e41ffc5d07)
PR from main: https://github.com/cloudfoundry/cli/pull/3397

## Description of the Change

Enhance the cf auth command with a parameter --assertion. The content of this token should be either a user token in order to perform a jwt-bearer or a client token in order to perform a client_credentials grant with federated trust.

UAA supports JWT bearer since UAA 4.5.0 , see https://docs.cloudfoundry.org/api/uaa/version/77.25.0/index.html#jwt-bearer-token-grant
UAA support the federated client credential flow since 77.25.0

## Why Is This PR Valuable?

CF can be integrated into Github Action without any extra secret setup in Github Repo.
Customer can then decide about using external tokens like github action token for user and/or client authentication.

In a PR you retrieve a id_token from gh action, this can be passed with cf auth --assertion <github id token> so that you are authentication in or to do a cf push ...

## Applicable Issues

* https://github.com/cloudfoundry/cli/issues/3368

## How Urgent Is The Change?

* it is an enhancement but it solves security issues, because CF integrations need to omit secrets and/or client certificates, but integration of github action with CF is only possible if you store a secret in Github

## Other Relevant Parties

Only CF landscapes with a configured trust to external OIDC parties

### DIFF
--- a/actor/v7action/auth.go
+++ b/actor/v7action/auth.go
@@ -25,7 +25,7 @@ func NewDefaultAuthActor(config Config, uaaClient UAAClient) AuthActor {
 }
 
 func (actor defaultAuthActor) Authenticate(credentials map[string]string, origin string, grantType constant.GrantType) error {
-	if grantType == constant.GrantTypePassword && actor.config.UAAGrantType() == string(constant.GrantTypeClientCredentials) {
+	if (grantType == constant.GrantTypePassword || grantType == constant.GrantTypeJwtBearer) && actor.config.UAAGrantType() == string(constant.GrantTypeClientCredentials) {
 		return actionerror.PasswordGrantTypeLogoutRequiredError{}
 	}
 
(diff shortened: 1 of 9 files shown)


Samze:
Looks like integration tests are failing due to changes to the auth command, the fix https://github.com/cloudfoundry/cli/pull/3458 will need to be ported to this PR branch once it has been merged into main.

strehle:
> Looks like integration tests are failing due to changes to the auth command, the fix #3458 will need to be ported to this PR branch once it has been merged into main.

Thank you, I picked it into this PR

strehle:
@a-b This PR is same as https://github.com/cloudfoundry/cli/pull/3454 but https://github.com/cloudfoundry/cli/pull/3454 was closed after a mistake in rebase. 

I still see not everything green, but I dont see test failures related to the changes

PR #3468, opened 2025-03-26
[v8] Improve integration tests stability

a-b:
This should improve reliability of the integration tests


Makefile
```diff
@@ -20,7 +20,7 @@ UNAME_S := $(shell uname -s)
 POLL_PROGRESS_THRESHOLD=120s
 
 GINKGO_FLAGS ?= -r -randomize-all -require-suite
-GINKGO_INT_FLAGS = $(GINKGO_FLAGS) --poll-progress-after $(POLL_PROGRESS_THRESHOLD)
+GINKGO_INT_FLAGS = $(GINKGO_FLAGS) --poll-progress-after $(POLL_PROGRESS_THRESHOLD) -flake-attempts $(FLAKE_ATTEMPTS)
```
Samze, line 23: **thought** I'm pretty torn about this, I fear this will lead us to never discover and resolve the flakes and to instead live with the consequences of retries and increased runtime and possibly mask real non-deterministic bugs in the CLI.

a-b: So far, we have only manually rerun tests on flake occurrences without taking further action. Eventually, we might analyze these flake attempts and create metrics, including the duration of tests and the count of attempts made.

a-b: **observation** In the past, a significant fraction of flakes were caused by networking between VMs. This is no longer the case, as everything is dockerized within a single VM.

a-b: The biggest downside is a potentially prolonged failure feedback loop. Which IMO an acceptable tradeoff to babysitting pipelines.

Samze (approved):

PR #3481, opened 2025-04-04
Fix double quote error on locale

Samze:
Else this evaluates to ""en_US.UTF-8"" and see this warning
```
bash: warning: setlocale: LC_ALL: cannot change locale ("en_US.UTF-8"): No such file or directory 
```


xandroc (approved):

evanfarrar (approved):
Yep!

PR #3503, opened 2025-05-12
[v8] Remove explicit cli version check from plugin system

Samze:
## Description of the Change

The code does not need to enumerate versions since the code being run is the code that either handles or does not handle the specified library version.


gururajsh (approved):
LGTM

PR #3510, opened 2025-05-23
update golang to 1.24.3

Benjamintf1:

## Description of the Change

## Why Is This PR Valuable?

## Applicable Issues

## How Urgent Is The Change?

## Other Relevant Parties

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module code.cloudfoundry.org/cli
 
-go 1.23.6
+go 1.24.3
 
 require (
 	code.cloudfoundry.org/bytefmt v0.36.0


.github/workflows/golangci-lint.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.61
+          version: v1.64.8
```
marckhouzam, line 38: I believe that using the notation `vX.Y` would use the latest patch version of `vX.Y` (see below for confirmation).
Do you want to lock the version to `v1.64.8`, or do you prefer to keep it to the latest patch of version by using `v1.64`?

**Confirmation:**
This was hard to confirm by looking at previous CI runs because it happens that the `v1.61` version only has a single patch `v1.61.0`.  I went hunting around on Github and I found the following project that uses the same action and specifies a `vX.Y` version and we can see that it picks the latest patch of that `vX.Y` version.

We can see [here](https://github.com/dharasimic/CI-Go-Alura/blob/9e57e15082af19e08a06e2dea78594a993063337/.github/workflows/go.yml#L25C9-L25C23) that the workflow specifies `version: v1.59` but when it runs we can see [here](https://github.com/dharasimic/CI-Go-Alura/actions/runs/13863945847/job/38798711480#step:5:18) that it uses the latest patch version: 
`Requested golangci-lint 'v1.59', using 'v1.59.1', calculation took 42ms`

marckhouzam (commented):
Thanks for you patience in fixing all the linter problems!

.github/workflows/golangci-lint.yml
Benjamintf1: image
https://github.com/cloudfoundry/cli/actions/runs/15215293911/job/42799266551
^ this is how it ran on the first run with v1.64. 

My plan was I would try and bump it to an even later version after merging this, but reducing changes to only bump to the latest 1.64 first. I'm not sure why in the first case it used a later version but when running this version it didn't.

Benjamintf1: oh wait, i see

PR #3526, opened 2025-06-12
Support "trace" log level [v8]

### LINKED ISSUE #3525: Log level does not support the "trace" keyword

- [ ] I attempted to run the command with `CF_TRACE=1` to help debug the issue.
- [ ] I am reporting a bug that others will be able to reproduce.

While working on code changes I noticed that I wasn't able to see logs if I set the log level to `trace`.  It turns out that the log level using the `trace` keyword isn't supported although it has been added to the logger 7 years ago.

**What happened**

Notice that a debug log is being printed when I use the `debug` level
```
$ CF_LOG_LEVEL=debug cf create-package hello
Creating and uploading bits package for app hello in org myorg / space myspace as john.doe@example.com...
DEBU[0000] convert to translatable error                 err="actionerror.ApplicationNotFoundError{GUID:\"\", Name:\"hello\"}"
App 'hello' not found.
FAILED
```
but that if I set the log level to `trace` the debug printout does not appear.
```
$ CF_LOG_LEVEL=trace cf create-package hello
Creating and uploading bits package for app hello in org myorg / space myspace as john.doe@example.com...
App 'hello' not found.
FAILED
```
however, if I set the debug level to 6, which corresponds to `trace` it does show:
```
$ CF_LOG_LEVEL=6 cf create-package hello
Creating and uploading bits package for app hello in org myorg / space myspace as john.doe@example.com...
DEBU[0000] convert to translatable error                 err="actionerror.ApplicationNotFoundError{GUID:\"\", Name:\"hello\"}"
App 'hello' not found.
FAILED
```

**Expected behavior**

I should be able to use `CF_LOG_LEVEL=trace`

**Exact Steps To Reproduce**

See above.

marckhouzam:
## Description of the Change

This PR allows to use the `trace` keyword to set a log level, as supported by the "logrus" library as seen here https://github.com/sirupsen/logrus/blob/d1e6332644483cfee14de11099f03645561d55f8/logrus.go#L111

The PR takes the approach of parsing log levels using a "logrus" library function instead of duplicating the code locally.

## Why Is This PR Valuable?

It allow to use `CF_LOGL_LEVEL=trace` when debugging, which is supported by the log library.
It would also open the door to the CLI code to use `log.Trace()` for an extra level of logs.
 
## Applicable Issues

Fixes #3525 

## How Urgent Is The Change?

Not urgent.

### DIFF
--- a/util/configv3/env.go
+++ b/util/configv3/env.go
@@ -3,10 +3,10 @@ package configv3
 import (
 	"encoding/json"
 	"strconv"
-	"strings"
 	"time"
 
 	"code.cloudfoundry.org/cli/util/trace"
+	log "github.com/sirupsen/logrus"
 )
 
 // EnvOverride represents all the environment variables read by the CF CLI
@@ -115,17 +115,8 @@ func (config *Config) LogLevel() int {
 			return int(envVal)
 		}
 
-		switch strings.ToLower(config.ENV.CFLogLevel) {
-		case "fatal":
-			return 1
-		case "error":
-			return 2
-		case "warn":
-			return 3
-		case "info":
-			return 4
-		case "debug":
-			return 5
+		if level, err := log.ParseLevel(config.ENV.CFLogLevel); err == nil {
+			return int(level)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


marckhouzam:
It was brought up that using the word `trace` might cause confusion with `CF_TRACE` which is a completely different thing.  Let's keep this PR on ice until we have more time to discuss.

gururajsh:
@marckhouzam can we close this if not needed now?

marckhouzam:
OK, let’s close for the moment

PR #3535, opened 2025-07-09
Update cf push rolling and and canary deployment behavior with scaling flags

prkalle:
## Description of the Change

This PR addresses an unexpected behavior where cf push commands using `--strategy rolling`/ `--strategy canary` alongside scaling flags (-i, -m, --disk --log-rate-limit) would first scale the application (potentially restarting all instances) before initiating the incremental rolling deployment.

Problem:
Prior to CAPI changes [CAPI_4364](https://github.com/cloudfoundry/cloud_controller_ng/pull/4264) [CAPI_4247](https://github.com/cloudfoundry/cloud_controller_ng/pull/4247), Deployments do not fully support parameters typically available via the cf scale action. This means that changes to instances, memory, or disk were not rolled out incrementally with Deployments. For example, cf push my app -i 5 -m 1G --strategy rolling would unexpectedly scale to 5 instances and 1G first, before the true rolling deployment commenced. Also, since scale is applied before a deployment, restarting the old web process will cause the new scale parameters to be applied.

## Why Is This PR Valuable?

Users leveraging rolling deployments will experience smoother, truly incremental updates when applying scaling changes.

## How Urgent Is The Change?

Fairly urgent as it helps rolling deployments smoother.


gururajsh (approved):
LGTM

PR #3539, opened 2025-07-14
Move to bosh lites from shepherd envs

gururajsh:
The integration tests used shepherd environments(broadcom hosted) to run. This PR is to create bosh lite environments hosted on CFF GCP.

The PR adds

Workflows to create, delete bosh lite environments.
Modifies integration tests workflow to use these environments.

CATs job is removed from the integration test workflow. Will try and introduced separate workflow for them.


integration/v7/isolated/app_command_test.go
```diff
@@ -355,7 +355,7 @@ applications:
 			})
 
 			When("the app exists", func() {
-				When("isolation segments are available", func() {
+				XWhen("isolation segments are available", func() {
```
Samze, line 358: **question** Any future plans to re-enable this? Does CATs have coverage here, should we just delete this test if

Samze (approved):

PR #3591, opened 2025-10-03
Remove mention of rename-buildpack in related cmds

marckhouzam:
## Description of the Change

"rename-buildpack" no longer exists as a command: it was removed by commit ba9cea0e84ef1320e2d66dbe179780229c9672dc.

## Why Is This PR Valuable?

Showing `rename-buildpack` in the `SEE ALSO` section of the help for `cf stacks` and `cf buildpacks` is confusing to the user since that command does not exist.

## How Urgent Is The Change?

Not urgent.


marckhouzam:
I will fix the integration tests as soon as I find time.

gururajsh (approved):
LGTM

Issue #3350, opened 2025-01-07
Show routable / readiness state in the `cf app` command


maxmoehl:
**What's the user value of this feature request?**

Right now readiness checks can be configured via the manifest but the CLI doesn't give any indication whether an application is ready. This causes confusion as the app is unreachable (404: route not found) but in the CLI it is shown as running with no indication of anything being wrong.

**Who is the functionality for?**

Users pushing apps with readiness checks.

**How often will this functionality be used by the user?**

Often, every time the readiness of app instances needs to be inspected.

**Who else is affected by the change?**

Operators who troubleshoot issues reported by users, anyone who might be parsing the output of the `cf app` command.

**Is your feature request related to a problem? Please describe.**

Yes, it is tedious to use `cf curl`s to find the processes and list their stats to view the routable state.

**Describe the solution you'd like**

An additional column in the process table to view the readiness / routable state:

```
$ cf app my-app
Showing health and status for app my-app in org my-org / space my-space as user@domain.tld...

name:              my-app
requested state:   started
routes:            my-app.cf.example.com
[...]

     state     since                  cpu    memory          disk            logging             cpu entitlement   ready   details
#0   running   2025-01-07T09:11:53Z   2.1%   16.2M of 128M   36.9M of 256M   0B/s of unlimited   106.9%            true

```

**Describe alternatives you've considered**

An additional command but that seems unnecessary for just viewing such information.

**Additional context**

[RFC0020 Readiness Healthchecks](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0020-readiness-healthchecks.md)
[CC v3 API: The process stats object](https://v3-apidocs.cloudfoundry.org/version/3.184.0/index.html#the-process-stats-object)

I'm willing to provide a PR for this.

maxmoehl:
I've built a POC in my fork at [feat/add-ready-column](https://github.com/maxmoehl/cf-cli/tree/feat/add-ready-column).

maxmoehl:
I just noticed that the RFC contains an example on how to implement this part:

```
     state     routable   since                  cpu    memory          disk         logging            details
#0   running   yes        2023-06-27T15:07:14Z   0.6%   46.8M of 192M   179M of 1G   0/s of unlimited
#1   running   no         2023-06-27T15:11:43Z   0.0%   0 of 0          0 of 0       0/s of 0/s
[...]

PR #3615, opened 2025-10-31
feat: add ready indicator to cf app command (v8)

vchrisb:
## Description of the Change

CC version 3.144.0 introduced readiness checks and with them a new status field for processes: routable. This commit exposes that new field in the cf app command to allow users to easily determine the state of their application instances.

See: https://github.com/cloudfoundry/capi-release/releases/tag/1.158.0

I adapted #3429 for the v8 branch to solve #3350.

The routable information is introduced at the end of the row to not break the column structure.


gururajsh (approved):
LGTM

PR #3620, opened 2025-11-06
Bump go dependencies [v8]

gururajsh:
This PR fixes bumps the go dependencies of integration test modules.


moleske:
Not necessary for this pr, but would it be worth it in a future pr to update the [dependabot.yml](https://github.com/cloudfoundry/cli/blob/main/.github/dependabot.yml#L3) file add the integration folder's go mods?

Whoops should have made this comment on the main branch pr.  Definitely don't bother to update dependabot.yml on the v8 branch

gururajsh:
> Not necessary for this pr, but would it be worth it in a future pr to update the [dependabot.yml](https://github.com/cloudfoundry/cli/blob/main/.github/dependabot.yml#L3) file add the integration folder's go mods?
> 
> Whoops should have made this comment on the main branch pr. Definitely don't bother to update dependabot.yml on the v8 branch

I thought about that. But hydrabroker tests are notorious for failing. And we do not run integration tests for dependabot PRs. So went with bumping manually when needed.

PR #3633, opened 2025-11-20
Make droplet's lifecycle information available

benjaminguttmann-avtq:
## Description of the Change

The PR makes the droplet's lifecycle information available as a property of the droplet object.

## Why Is This PR Valuable?

Within cf_exporter we want to specify different actions based on which droplet lifecycle is used. Therefore we need to make that information available. 

## Applicable Issues

none

## How Urgent Is The Change?

not urgent

## Other Relevant Parties

Who else is affected by the change? 

[v8 PR](https://github.com/cloudfoundry/cli/pull/3634)

PR #3634, opened 2025-11-20
Make droplet's lifecycle information available

benjaminguttmann-avtq:
## Description of the Change

The PR makes the droplet's lifecycle information available as a property of the droplet object.

## Why Is This PR Valuable?

Within cf_exporter we want to specify different actions based on which droplet lifecycle is used. Therefore we need to make that information available. 

## Applicable Issues

-

## How Urgent Is The Change?

not urgent

## Other Relevant Parties

### DIFF
--- a/resources/droplet_resource.go
+++ b/resources/droplet_resource.go
@@ -44,9 +44,9 @@ type DropletBuildpack struct {
 }
 
 // An object describing the lifecycle that was used when staging the droplet
-// possible values for type: "buildpack", "cnd", "docker"
+// possible values for type: "buildpack", "cnb", "docker"
 type DropletLifecycle struct {
-	Type string `json:"name"`
+	Type string `json:"type"`
 }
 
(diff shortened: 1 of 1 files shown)


resources/droplet_resource.go
```diff
@@ -41,11 +43,18 @@ type DropletBuildpack struct {
 	Version string `json:"version"`
 }
 
+// An object describing the lifecycle that was used when staging the droplet
+// possible values for type: "buildpack", "cnd", "docker"
+type DropletLifecycle struct {
+	Type string `json:"name"`
```
anujc25: Should we call the json tag as `type` instead of `name here?

anujc25 (commented):
Couple of comments but otherwise looks good.

It might be good to rebase this PR as well so that it can take the improvements done in the integration tests pipeline. (same comments are applicable to #3633 as well)

PR #3647, opened 2025-12-01
[v8] Improve integration test flakiness around servicebroker tests 

anujc25:
## Description of the Change

Fix intermittent integration test failures around servicebroker
* The ensureAppIsDeployed helper was calling appResponds() immediately and passing the result (a boolean) to Eventually. This prevented Gomega from polling the function, causing failures if the app wasn't ready instantly.
* This change passes the function reference appResponds to Eventually so it can be polled correctly, and adds a 1-minute timeout to allow for slower environments.

## Why Is This PR Valuable?

Improves integration test reliability

## Applicable Issues

None

## How Urgent Is The Change?

Not urgent


gururajsh (approved):
LGTM

PR #3655, opened 2025-12-05
[main] Add "--strategy" parameter to "bind-service" command

jochenehret:
## Description of the Change

Enhancement for [rfc-0040-service-binding-rotation](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0040-service-binding-rotation.md#cf-cli).

Add parameter `strategy` to `bind-service` command. Allowed values are `single` (default) and `multiple`. With `multiple`, the CLI sends and additional parameter `"strategy": "multiple"` to the CC v3 API:
https://v3-apidocs.cloudfoundry.org/version/3.207.0/index.html#create-a-service-credential-binding 

## Why Is This PR Valuable?

This PR enhances the CLI so that the multiple service binding feature is enabled for end users. More PRs for other related commands will follow.

## Applicable Issues

https://github.com/cloudfoundry/community/pull/1198

## How Urgent Is The Change?

Not super urgent.

## Other Relevant Parties

Change is backwards compatible, so no regressions expected.


anujc25:
Added some comment on #3654 that applies to this PR as well.

jochenehret:
Updated this PR with review comments from https://github.com/cloudfoundry/cli/pull/3654

PR #3664, opened 2025-12-09
[main] Command "unbind-service" can handle multiple bindings

jochenehret:
## Description of the Change

The "unbind-service" command shall be able to handle multiple bindings per (app, service instance). This PR first retrieves the list of all bindings for a given app and a given service instance name. Then it loops over the list and calls `DELETE /v3/service_credential_bindings/:guid` for each item. A new parameter "--guid" allows to delete only a specific binding

## Why Is This PR Valuable?

Allows users to delete multiple service bindings with one CF CLI command.

## Applicable Issues

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0040-service-binding-rotation.md#cf-cli

## How Urgent Is The Change?

Not super-urgent.


anujc25:
Can you please help rebasing this PR once? We fixed the environment creation issue that we were seeing for integration tests.

anujc25 (approved):
LGTM. Thanks 👍

PR #3678, opened 2026-01-08
[main] Optimize integration test execution to reduce infrastructure load related flakes

anujc25:
## Description of the Change

This PR optimizes the integration test execution in CI to reduce infrastructure load and improve test reliability. 

The main change splits the `integration-tests-full-ci` Makefile target into 3 sequential batches instead of running all 6 test suites together with high parallelism. Additionally, the number of parallel nodes is reduced from 12 to 4 in the GitHub Actions workflow.

Test suites are now grouped logically into batches that run sequentially with controlled parallelism to prevent infrastructure overload while maintaining reasonable test execution time.

## Why Is This PR Valuable?

This change addresses infrastructure load issues that cause test failures due to resource contention. By reducing concurrent load from 12 parallel nodes to 4 and spreading test execution over time through batching, the tests become significantly more stable and reliable on resource-constrained infrastructure.

Users running integration tests in CI environments with limited resources will benefit from more consistent test results and fewer infrastructure-related failures.

## Applicable Issues

## How Urgent Is The Change?

This change improves the reliability of integration tests in CI but is not time-sensitive.

### DIFF
--- a/.github/workflows/tests-integration-reusable.yml
+++ b/.github/workflows/tests-integration-reusable.yml
@@ -18,7 +18,7 @@ on:
 
 env:
   ENV_NAME: ${{ inputs.env-name }}
-  NODES: 12
+  NODES: 4
   FLAKE_ATTEMPTS: ${{ vars.TEST_FLAKE_ATTEMPTS || '5' }}
   BBL_CLI_VERSION: ${{ vars.BBL_CLI_VERSION }}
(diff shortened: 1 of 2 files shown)


anujc25:
@gururajsh can you please help take a look? I got failures twice but I think this will reduce the flakes and possible failures because of the infrastructure limits. WDYT?

PR #3758, opened 2026-04-10
RFC0055 Identity-Aware Routing

rkoster:
## Summary

This PR implements the identity-aware routing RFC ([RFC#1438](https://github.com/cloudfoundry/community/pull/1438)) in the CF CLI. It provides the complete implementation from a prior POC, with two significant improvements:

1. **Terminology rebrand** — updated to match the latest RFC revision ("route policies" instead of "access rules")
2. **User-friendly source flags** — name-based flags (following `add-network-policy` conventions) instead of raw GUID selectors

**Terminology Update (Breaking Change):**
- Commands: `access-rules` → `route-policies`, `add-access-rule` → `add-route-policy`, `remove-access-rule` → `remove-route-policy`
- Flags: `--enforce-access-rules` → `--enforce-route-policies`, `--selector` → `--source`
- API paths: `/v3/access_rules` → `/v3/route_policies`
- Resource types: `AccessRule` → `RoutePolicy`

This is an acceptable breaking change since the feature has not been GA-released.

## Motivation

The original POC implementation used "access rules" terminology, but the RFC was updated to use "route policies" for better alignment with existing CF concepts. Additionally, the commands required users to manually construct GUID-based selectors (e.g., `cf:app:d76446a1-f429-4444-8797-be2f78b75b08`), which was cumbersome and error-prone.

This PR follows CF CLI conventions from commands like `add-network-policy` to provide a user-friendly interface with name-based flags.

### 1. Terminology Rebrand

**Complete rebrand across all affected files:**
- Commands: `route-policies`, `add-route-policy`, `remove-route-policy`
- Flags: `--enforce-route-policies`, `--source`, `--source-app`, `--source-space`, `--source-org`, `--source-any`
- API endpoints: `/v3/route_policies`
- Query parameters: `sources`, `source_guids`
- Resource fields: `RoutePolicy.Source`, `Domain.EnforceRoutePolicies`, `Domain.RoutePoliciesScope`
- Table columns: Updated to show "source" terminology

**Rationale (per RFC commit be8d74c):**
> The term "route policy" is more idiomatic within Cloud Foundry, which already uses "network policies" for app-to-app network access. This naming convention better communicates that these are routing-layer policies that control which sources can access specific routes.

### 2. Command Interface Improvements

**Before:**
```bash
cf add-access-rule RULE_NAME DOMAIN SELECTOR --hostname HOST
cf remove-access-rule RULE_NAME DOMAIN --hostname HOST
```

**After:**
```bash
# add-route-policy examples
cf add-route-policy apps.identity --source-app frontend-app --hostname backend
cf add-route-policy apps.identity --source-space monitoring --hostname api
cf add-route-policy apps.identity --source-org platform --hostname shared-api
cf add-route-policy apps.identity --source-any --hostname public-api

# Cross-space/org access
cf add-route-policy apps.identity --source-app client --source-space prod-space --source-org prod-org --hostname api

# Advanced: raw source still supported
cf add-route-policy apps.identity --source cf:app:GUID --hostname backend

# remove-route-policy
cf remove-route-policy apps.identity --source cf:app:GUID --hostname backend
```

### 3. New Flags

**add-route-policy:**
- `--source-app APP_NAME` - Specify source app by name (resolves to GUID)
- `--source-space SPACE_NAME` - Specify space context for app lookup or create space-level policy
- `--source-org ORG_NAME` - Specify org context for space/app lookup or create org-level policy
- `--source-any` - Allow any authenticated app
- `--source SOURCE` - Raw GUID-based source for advanced users

**remove-route-policy:**
- `--source SOURCE` - Required. Specify the source to remove

**create-shared-domain / create-private-domain:**
- `--enforce-route-policies` - Require route policies for all routes on domain
- `--route-policies-scope [app|space|org]` - Minimum granularity for policies

### 4. RFC Alignment: No User-Provided Names

Per RFC commits 882b69a and 11752f2, route policies no longer have user-provided names. They are identified by their source only, with labels/annotations available for metadata.

**Key Changes:**
- Removed `RULE_NAME` positional argument from `add-route-policy` command
- Removed `Name` field from RoutePolicy API resource
- Updated `route-policies` list command to show 4 columns: route, source, scope, source/name
  - The `source/name` column shows the resolved name of the app/space/org from the source
- Updated `remove-route-policy` to use `--source` flag instead of name
- Renamed `DeleteAccessRule()` → `DeleteRoutePolicyBySource()`

### 5. Enhanced Output

**add-route-policy:**
```
Adding route policy for route backend.apps.identity as admin...
  scope: app, source name: frontend-app
  source: cf:app:d76446a1-f429-4444-8797-be2f78b75b08
OK
```

**route-policies:**
```
route                            source                                         scope   source/name
backend.apps.identity            cf:app:d76446a1-f429-4444-8797-be2f78b75b08    app     frontend-app
api.apps.identity/metrics        cf:space:2b26e210-1b48-4e60-8432-f24bc5927789  space   monitoring
public-api.apps.identity         cf:any                                         any
```

### 6. Validation & Error Handling

- Validates exactly one primary source is specified
- Provides helpful error messages when app not found in current space
- Suggests using `--source-space` and `--source-org` flags for cross-space/org access
- Follows CF CLI patterns from `add-network-policy` command

### 7. Code Quality

- `RoutePolicyArgs` consolidates the positional-argument struct that was independently defined for `add-route-policy` and `remove-route-policy` into a single shared type in `command/flag/arguments.go`

## Testing

New and updated test files:

- `command/v7/add_route_policy_command_test.go` — flags, validation, error handling, success path
- `command/v7/remove_route_policy_command_test.go` — flags, validation, error handling, success path
- `command/v7/route_policies_command_test.go` — listing, column output
- `command/v7/route_policy_source_flags_test.go` — 23 specs covering all source flag validation and resolution paths (internal package tests, no generated fakes needed)
- `command/v7/create_shared_domain_command_test.go` — extended with `--enforce-route-policies`/`--scope` coverage, CAPI version check, shared behaviour helper
- `command/v7/create_private_domain_command_test.go` — same coverage as shared domain
- `command/v7/enforce_route_policies_shared_test.go` — shared Ginkgo behaviour helper for domain command tests
- `actor/v7action/route_policy_test.go` — actor-layer tests
- `api/cloudcontroller/ccv3/route_policy_test.go` — ccv3 client tests
- `actor/v7action/domain_test.go` — extended with scope/enforce contexts using `JustBeforeEach` pattern
- `resources/route_policy_resource_test.go` — marshal/unmarshal tests for the resource type

Coverage: flag validation, source resolution, GUID lookup, error handling, cross-space/org scenarios, CAPI version checks, domain flags.

## Breaking Changes

This is a **complete rebrand** with no backward compatibility:
- All command names changed
- All API paths changed
- All resource field names changed
- All flags renamed

**This is acceptable** since:
1. The RFC has not reached GA
2. Only POC/lab implementations exist
3. The RFC itself was updated with these breaking changes
4. Clean cut-over is better than long deprecation for pre-GA features

## Files Changed

- ~40 files modified
- **Key areas:**
  - Resources: `route_policy_resource.go`
  - API client: `route_policy.go`, `api_routes.go`, `query.go`
  - Actor: `route_policy.go`, error types, interfaces
  - Commands: `route_policies_command.go`, `add_route_policy_command.go`, `remove_route_policy_command.go`
  - Domain commands: Updated flags in `create_shared_domain` and `create_private_domain`
  - Generated fakes: `fake_cloud_controller_client.go`, `fake_actor.go` regenerated for new methods only — bulk `Invocations()` cleanup is in PR #3799

## Checklist

- [x] Tests added/updated and passing
- [x] Documentation (help text) updated
- [x] Breaking changes documented
- [x] Follows CF CLI conventions
- [x] Aligned with RFC terminology rebrand
- [x] Code formatted with gofmt
- [x] Mocks regenerated with counterfeiter (bulk `Invocations()` cleanup in separate PR #3799)
- [x] CLI binary builds successfully
- [ ] Manual testing against live CF deployment (requires CAPI backend support)

## Pending

1. Coordination with CAPI team to ensure API implementation uses new terminology
2. Manual testing against a CF deployment with route policies API support
3. Integration testing with full CF ecosystem

## Related

- RFC PR: https://github.com/cloudfoundry/community/pull/1438
- RFC commit (terminology rebrand): be8d74c1e7c0adf0bbbed7ced64e44bdd0c383b2
- RFC commit (removed names): 882b69a, 11752f2
- Standalone fakes cleanup PR: #3799

Co-authored-by: Ruben Koster <rkoster@starkandwayne.com>


rkoster:
Don't worry about this PR just yet, just doing some more POC work on the RFC: https://github.com/cloudfoundry/community/pull/1438

cppforlife:
> I'll update the RFC with this terminology and push a new commit. Thanks for pushing on this—it's a significant UX improvement.

glad we all agree =)

one nit: 
instead of
```
cf add-route-policy apps.identity --source-app frontend-app --hostname backend
```
i think it will be much more fluent to have
```
cf add-route-policy apps.identity --hostname backend --allow-from-app frontend-app 
cf add-route-policy apps.identity --hostname api --path /metrics --allow-from-space x
...
cf add-route-policy apps.identity --hostname api --path /metrics --allow-from cf:app:...
```

ameowlia:
Issue: add-route-policy and remove-route-policy do not have matching flags.

This works:
```
go run . add-route-policy apps.identity --source-app frontend --hostname backend
```

There is no matching –source-app for the remove policy.

```
go run . remove-route-policy apps.identity --source frontend --hostname backend -f
source must start with one of: cf:app:, cf:space:, cf:org:, or be exactly 'cf:any'
FAILED
exit status 1
```

ameowlia:
I was testing all of this and I accidentally deployed without updating capi, which resulted in this error message:

```
$cf create-shared-domain apps.identity --enforce-route-policies --scope any

Creating shared domain apps.identity as admin...
Unknown field(s): 'enforce_route_policies', 'route_policies_scope'
FAILED
exit status 1
```

Is there a way to give a clearer error message? Maybe checking the capi version? Or saying something like "Route Policies are not supported in this version of CAPI"?

rkoster:
@cppforlife — circling back on this since some time has passed.

> i think it will be much more fluent to have `--allow-from-app frontend-app` / `--allow-from-space x` / `--allow-from cf:app:...`

The `--source-*` naming is intentional — it mirrors CF's existing network policy convention. `cf add-network-policy` uses `SOURCE_APP` as its primary concept:

```
cf add-network-policy SOURCE_APP DESTINATION_APP [-s DESTINATION_SPACE_NAME [-o DESTINATION_ORG_NAME]] [--protocol (tcp | udp) --port RANGE]
```

During RFC review, @ameowlia [explicitly pushed for this feature to follow the network policy CLI pattern](https://github.com/cloudfoundry/community/pull/1438#issuecomment-4077590965):

> If you want to make it mirror c2c you should make these API endpoints that can be wrapped by the CLI.

And @Gerg [noted the family resemblance](https://github.com/cloudfoundry/community/pull/1438#discussion_r3067030904):

> `add-access-rule` kinda sounds like a synonym of `add-network-policy`

Using `--allow-from-*` reads more imperatively but diverges from the `source → destination` mental model that CF networking users already know. Keeping `--source-app` / `--source-space` / `--source-org` makes the two commands feel like siblings, which is the right outcome given the consistency feedback received during RFC review.

rkoster:
> Issue: add-route-policy and remove-route-policy do not have matching flags.
> There is no matching `--source-app` for the remove policy.

Fixed in 0ab3e029a. The source flags (`--source-app`, `--source-space`, `--source-org`, `--source-any`, `--source`) are now extracted into a shared `RoutePolicySourceFlags` struct embedded in both commands. `remove-route-policy` now accepts exactly the same source flags as `add-route-policy`, so the example from the issue now works:

```
cf remove-route-policy apps.identity --source-app frontend --hostname backend
```

rkoster:
Done in fba97d894 — here's what was added:

**`ccversion.MinVersionRoutePolicies`** — placeholder constant `"3.999.0"` in `api/cloudcontroller/ccversion/minimum_version.go`. The comment in-file calls out that it must be replaced with the real CAPI version once the team confirms it.

**Failing test** — `api/cloudcontroller/ccversion/minimum_version_test.go` contains `TestMinVersionRoutePoliciesIsNotPlaceholder` which fails as long as the constant holds `"3.999.0"`, so the TODO won't be silently forgotten.

**Version checks added:**
- `add-route-policy`, `remove-route-policy`, `route-policies` — unconditional `MinimumCCAPIVersionCheck` at top of `Execute()`
- `create-shared-domain`, `create-private-domain` — conditional check (fires only when `--enforce-route-policies` is passed), using `"--enforce-route-policies"` as the feature name in the error message

Let me know what version number to drop in once CAPI has it.

api/cloudcontroller/ccversion/minimum_version.go
```diff
@@ -22,4 +22,10 @@ const (
 	MinVersionEmbeddedProcessInstances = "3.211.0"
 
 	MinVersionUpdateStack = "3.211.0"
+
+	// MinVersionRoutePolicies is a placeholder until the CAPI team confirms the
+	// version that introduces /v3/route_policies and the enforce_route_policies /
+	// route_policies_scope domain fields. Replace "3.999.0" with the real version
+	// once known. The test in minimum_version_test.go will keep failing until then.
+	MinVersionRoutePolicies = "3.999.0"
```
rkoster: The version should be changed once: https://github.com/cloudfoundry/cloud_controller_ng/pull/4910 has been merged and released.

rkoster:
This PR is ready for review. The specs are intentionally failing until the correct CAPI version has been filled in. But let's try and get this PR in a reviewed state, so the changes in this PR can get released closely after the CAPI changes get published.

actor/v7action/route_policy.go
```diff
@@ -0,0 +1,382 @@
...
+
+func (actor Actor) AddRoutePolicy(domainName, source, hostname, path string) (Warnings, error) {
+	allWarnings := Warnings{}
+
+	// Get the domain to ensure it exists and supports route policies
+	domain, warnings, err := actor.GetDomainByName(domainName)
+	allWarnings = append(allWarnings, warnings...)
+	if err != nil {
+		return allWarnings, err
+	}
```
jcvrabo, line 17: I would say... actually check if the domain enforces access rules, or... if reliant on the error returned by the capi, change the comment.

actor/v7action/route_policy.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// Delete the route policy
+	_, deleteWarnings, err := actor.CloudControllerClient.DeleteRoutePolicy(policyGUID)
+	allWarnings = append(allWarnings, Warnings(deleteWarnings)...)
+
+	return allWarnings, err
+}
+
+// GetRoutesByDomain gets routes for a domain with optional hostname and path filters
+func (actor Actor) GetRoutesByDomain(domainGUID, hostname, path string) ([]resources.Route, Warnings, error) {
```
jcvrabo: I would actually add this to route.go

actor/v7action/route_policy.go
```diff
@@ -0,0 +1,382 @@
...
+	if path != "" {
+		queries = append(queries, ccv3.Query{Key: ccv3.PathsFilter, Values: []string{path}})
+	}
+
+	ccv3Routes, warnings, err := actor.CloudControllerClient.GetRoutes(queries...)
+	if err != nil {
+		return nil, Warnings(warnings), err
+	}
+
+	var routes []resources.Route
```
jcvrabo: don't understand the purpose of this block, you're basically copying the response to a new array (ccv3Routes are in fact Routes)

command/v7/add_route_policy_command.go
```diff
@@ -0,0 +1,106 @@
...
+	"code.cloudfoundry.org/cli/v9/api/cloudcontroller/ccversion"
+	"code.cloudfoundry.org/cli/v9/command"
+	"code.cloudfoundry.org/cli/v9/command/flag"
+)
+
+type AddRoutePolicyCommand struct {
+	BaseCommand
+
+	RequiredArgs flag.RoutePolicyArgs `positional-args:"yes"`
+	Hostname     string                  `long:"hostname" required:"true" description:"Hostname for the route"`
```
jcvrabo: short "n" would be handy

command/v7/route_policy_source_flags.go
```diff
@@ -0,0 +1,175 @@
+package v7
```
jcvrabo, line 1: wonder if this package is the right place to have this shared code... perhaps shared actor or some other shared package would be more suitable

command/v7/route_policy_source_flags.go
```diff
@@ -0,0 +1,175 @@
...
+		return "cf:any", "scope: any, source: any authenticated app", allWarnings, nil
+	}
+
+	// --source-app (with optional --source-space / --source-org for cross-space/org lookup)
+	if f.SourceApp != "" {
+		spaceGUID := config.TargetedSpace().GUID
+		spaceName := config.TargetedSpace().Name
+		orgName := config.TargetedOrganization().Name
+
+		if f.SourceSpace != "" {
```
jcvrabo: --source-space can happen both stand-alone and as part of a cross-spaced app. Perhaps this piece of code could either run before checking for an app or having it in a standalong function, to avoid code duplication.

command/v7/route_policy_source_flags.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		return fmt.Sprintf("cf:app:%s", app.GUID), scopeDisplay, allWarnings, nil
+	}
+
+	// --source-space (primary: space-level policy)
+	if f.SourceSpace != "" {
+		orgGUID := config.TargetedOrganization().GUID
+		orgName := config.TargetedOrganization().Name
+		if f.SourceOrg != "" {
```
jcvrabo: same commend for --source-org as it was for --source-space

resources/route_policy_resource.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (r RoutePolicy) MarshalJSON() ([]byte, error) {
+	type alias struct {
+		GUID      string     `json:"guid,omitempty"`
+		CreatedAt *time.Time `json:"created_at,omitempty"`
+		UpdatedAt *time.Time `json:"updated_at,omitempty"`
+		Source    string     `json:"source"`
+		Metadata  *Metadata  `json:"metadata,omitempty"`
+
+		Relationships struct {
```
jcvrabo, line 29: thought it could also have relationships to whatever could be in source, but perhaps that has changed

actor/v7action/route_policy.go
```diff
@@ -0,0 +1,382 @@
...
+	if err != nil {
+		return nil, allWarnings, err
+	}
+
+	if len(routePolicies) == 0 {
+		// No route policies found - return empty list, not an error
+		return []RoutePolicyWithRoute{}, allWarnings, nil
+	}
+
+	// Build route lookup map from included resources
```
jcvrabo: don't know if it wouldn't be more efficient to simply filter the slices first and then build the map to filter the policies

actor/v7action/route_policy.go
```diff
@@ -0,0 +1,382 @@
...
+		for guid, route := range routeByGUID {
+			if route.Path == path {
+				filteredRoutes[guid] = route
+			}
+		}
+		routeByGUID = filteredRoutes
+	}
+
+	// Build domain name cache
+	domainCache := make(map[string]string)
```
jcvrabo: domain cache can be created above and initialized if a domain filter is created, in which case, a cache per se is not even necessary (I mean... to keep the code generic you would simply have a one entry map)

command/v7/remove_route_policy_command.go
```diff
@@ -0,0 +1,89 @@
...
+	"code.cloudfoundry.org/cli/v9/command"
+	"code.cloudfoundry.org/cli/v9/command/flag"
+)
+
+type RemoveRoutePolicyCommand struct {
+	BaseCommand
+
+	RequiredArgs flag.RoutePolicyArgs `positional-args:"yes"`
+	RoutePolicySourceFlags
+	Hostname string `long:"hostname" required:"true" description:"Hostname for the route"`
```
jcvrabo: short "n"

command/v7/route_policies_command.go
```diff
@@ -0,0 +1,94 @@
...
+	"code.cloudfoundry.org/cli/v9/api/cloudcontroller/ccversion"
+	"code.cloudfoundry.org/cli/v9/command"
+	"code.cloudfoundry.org/cli/v9/util/ui"
+)
+
+type RoutePoliciesCommand struct {
+	BaseCommand
+
+	Domain   string `long:"domain" description:"Filter by domain name"`
+	Hostname string `long:"hostname" description:"Filter by hostname"`
```
jcvrabo: short "n"

command/v7/route_policies_command.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"code.cloudfoundry.org/cli/v9/api/cloudcontroller/ccversion"
+	"code.cloudfoundry.org/cli/v9/command"
+	"code.cloudfoundry.org/cli/v9/util/ui"
+)
+
+type RoutePoliciesCommand struct {
+	BaseCommand
+
+	Domain   string `long:"domain" description:"Filter by domain name"`
```
jcvrabo: probably some short version like "d" would be handy

jcvrabo (changes requested):
it's missing a labels command/actor update to support labels update for route policies

api/cloudcontroller/ccversion/minimum_version.go
rkoster: Added a guard after fetching the domain: if `domain.EnforceRoutePolicies` is not set or false, we return a `DomainNotEnforcingRoutePoliciesError` before attempting to create the policy. CAPI already enforces this server-side (`route_policies_controller.rb:129-131`) with a 422, but the client-side check gives a cleaner user-facing error message with the domain name. Tests updated to set `EnforceRoutePolicies: true` on the success path and a new `When("the domain does not enforce route policies")` spec added.

actor/v7action/route_policy.go
rkoster: Done in 316c9af — moved `GetRoutesByDomain` to `route.go` alongside `GetRoutesBySpace` and `GetRoutesByOrg`.

actor/v7action/route_policy.go
rkoster: Fixed in 9adda5db — `ccv3.GetRoutes` already returns `[]resources.Route` directly, so the loop copying each element was an identity conversion. Collapsed to a single return.

command/v7/add_route_policy_command.go
rkoster: Added `short:"n"` in ab71adf.

command/v7/remove_route_policy_command.go
rkoster: Added `short:"n"` in ab71adf.

command/v7/route_policies_command.go
rkoster: Added `short:"n"` in ab71adf.

command/v7/route_policy_source_flags.go
rkoster: Fair question. The positional args (`RoutePolicyArgs`) already follow the `command/flag` convention. The reason `RoutePolicySourceFlags` stays in `command/v7` is that `resolveSource` takes a `v7.Actor` and `command.Config` — moving it to `command/flag` would create a circular import. Splitting the struct from its resolution logic would leave `command/flag` with a type it can't fully own. Keeping both together in their own file within `command/v7` is the least-bad option given those constraints.

command/v7/route_policy_source_flags.go
rkoster: Done in 0d8afe7 — extracted `resolveOrgGUID` and `resolveSpaceGUID` helpers. Both the `--source-app` and `--source-space` branches now call the same functions. Added dedicated unit tests for each helper.

command/v7/route_policy_source_flags.go
rkoster: Same fix as [the `--source-space` thread](https://github.com/cloudfoundry/cli/pull/3758#discussion_r3450764153) — the `--source-org` duplication in the `--source-space` branch is eliminated by `resolveOrgGUID`.

command/v7/route_policy_source_flags.go
rkoster: Follow-up in f73911a93: switched the test file to `package v7_test` with `v7fakes.FakeActor` to align with the rest of `command/v7`. The direct `resolveOrgGUID` and `resolveSpaceGUID` unit test blocks are gone since those helpers are unexported and inaccessible from `v7_test`, but all their code paths remain covered by the `resolveSource` tests exercising the full flag-combination matrix through `Execute()`.

command/v7/route_policy_source_flags.go
rkoster: The `--source-space + --source-org` error path test is still present after the refactor in f73911a93 — it's in the `error propagation` context under `Describe("resolveSource")`.

resources/route_policy_resource.go
rkoster: `source` in the CAPI response is a plain string GUID field (not a JSON:API relationship), so it cannot live in `Relationships`. To resolve source names without N extra API calls, cb0db1bbd switches to `?include=route,source` — CAPIs `IncludeRoutePolicySourceDecorator` batches all referenced app/space/org GUIDs into one query and returns them inline in the `included` section. `sourceInfoFromIncluded` then does a pure map lookup with no further API calls.

actor/v7action/route_policy.go
rkoster: Done — lines 19-21 check `domain.EnforceRoutePolicies.IsSet && domain.EnforceRoutePolicies.Value` and return `DomainNotEnforcingRoutePoliciesError` if the flag is absent or false, so no CAPI error is relied upon.

actor/v7action/route_policy.go
rkoster: Done in 1ee6df8c7. The three filter conditions now narrow `includedResources.Routes` down to a `filteredRoutes` slice first; `routeByGUID` is then built only from that filtered slice, and the domain-cache loop also iterates only over `filteredRoutes`.

actor/v7action/route_policy.go
rkoster: Done in 1ee6df8c7. `domainCache` is now created before the filter block and pre-populated with `domainCache[domain.GUID] = domain.Name` at the point where `GetDomainByName` is called for the domain filter. After filtering, all routes in the result share that GUID, so the subsequent cache-fill loop is a no-op for the filtered case — effectively a one-entry map as suggested.

command/v7/route_policies_command.go
rkoster: Added `short:"d"` in 1ee6df8c7.

actor/v7action/label.go
```diff
@@ -178,3 +179,55 @@ func (actor *Actor) updateResourceMetadata(resourceType string, resourceGUID str
 
 	return warnings, nil
 }
+
+// resolveRoutePolicyGUID finds the GUID of the route policy to operate on.
+// If source is non-empty, it finds the policy with that source.
+// If source is empty, it requires exactly one policy (returns ambiguity or not-found error).
+func (actor *Actor) resolveRoutePolicyGUID(routeURL, spaceGUID, source string) (string, *resources.Metadata, Warnings, error) {
```
jcvrabo: this function would live better in route_policy.go

command/v7/route_policy_source_flags.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	return nil
+}
+
+// resolveSource resolves source flags to a raw source string and a human-readable
+// scope description. Returns (source, scopeDisplay, warnings, error).
+func resolveSource(f RoutePolicySourceFlags, actor Actor, config command.Config) (string, string, v7action.Warnings, error) {
+	var allWarnings v7action.Warnings
```
jcvrabo, line 72: indeed the filter functions were added, but I also had in mind just a generic optimization which was based on checking the source (if not any) from org to space where your final return value would be something like fmt.Sprinf("cf:%s:%s", scope, sourceGuid). First check if org is present and if so set scope to org and sourceGuid to resolved orgGuid, and set variable orgGuid with the resolvedGuid, if no org flag, orgGuid is current orgGuid, then check space, following the same strategy (scope set to space if found and update sourceGuid with the resolved space GUID, which will use the orgGuid to find the space) and the same ida for the space flag... if no scope is set, then there's an error (no flags set for source)... more of a suggestion than anything else, but think it would make code cleaner too

command/v7/route_policy_source_flags.go
```diff
@@ -0,0 +1,187 @@
...
+			allWarnings, nil
+	}
+
+	return "", "", allWarnings, fmt.Errorf("no source specified")
+}
+
+// resolveOrgGUID returns the GUID and name of the org identified by --source-org,
+// falling back to the targeted org if --source-org is not set.
+func resolveOrgGUID(f RoutePolicySourceFlags, actor Actor, config command.Config) (string, string, v7action.Warnings, error) {
+	if f.SourceOrg == "" {
```
jcvrabo: if the suggested optimization would be implemented these helper functions would also not be required, as the guid resolving would occur only once

jochenehret (commented):
There's a problem with the new flag `--enforce-access-rules`:
```
cf create-shared-domain apps.identity --enforce-access-rules --scope any
Incorrect Usage: unknown flag `enforce-access-rules'
```

command/v7/set_label_command.go
jcvrabo, line 1: the Resources() function should also have the route-policy entry

command/v7/unset_label_command.go
jcvrabo, line 1: the Resources() function should also have the route-policy entry

rkoster:
> There's a problem with the new flag --enforce-access-rules:

Flag was renamed to `--enforce-route-policies`.

```
> cf create-shared-domain --help
NAME:
   create-shared-domain - Create a domain that can be used by all orgs (admin-only)

USAGE:
   main create-shared-domain DOMAIN [--router-group ROUTER_GROUP_NAME | --internal] [--enforce-route-policies [--scope (any|org|space)]]

OPTIONS:
   --router-group                Routes for this domain will use routers in the specified router group
   --internal                    Applications that use internal routes communicate directly on the container network
   --enforce-route-policies      Enable platform-enforced access control for routes on this domain (requires mTLS domain configuration in GoRouter)
   --scope                       Operator-level scope boundary for route policies: 'any', 'org', or 'space' (only valid with --enforce-route-policies)

SEE ALSO:
   add-route-policy, create-private-domain, domains, route-policies

> There's a problem with the new flag `--enforce-access-rules`:
> 
> ```
> cf create-shared-domain apps.identity --enforce-access-rules --scope any
> Incorrect Usage: unknown flag `enforce-access-rules'
> ```

File: command/v7/remove_route_policy_command.go [hunk 18]
```diff
@@ -0,0 +1,89 @@
...
+	"code.cloudfoundry.org/cli/v9/command/flag"
+)
+
+type RemoveRoutePolicyCommand struct {
+	BaseCommand
+
+	RequiredArgs flag.RoutePolicyArgs `positional-args:"yes"`
+	RoutePolicySourceFlags
+	Hostname string `long:"hostname" short:"n" required:"true" description:"Hostname for the route"`
+	Path     string `long:"path" description:"Path for the route"`
```
jcvrabo: perhaps a short -p

command/v7/add_route_policy_command.go
```diff
@@ -0,0 +1,106 @@
...
+	"code.cloudfoundry.org/cli/v9/command"
+	"code.cloudfoundry.org/cli/v9/command/flag"
+)
+
+type AddRoutePolicyCommand struct {
+	BaseCommand
+
+	RequiredArgs flag.RoutePolicyArgs `positional-args:"yes"`
+	Hostname     string                  `long:"hostname" short:"n" required:"true" description:"Hostname for the route"`
+	Path         string                  `long:"path" description:"Path for the route"`
```
jcvrabo: perhaps a short p

command/v7/label_updater.go
```diff
@@ -167,6 +172,14 @@ func (cmd *LabelUpdater) validateFlags() error {
 		}
 	}
 
+	if cmd.targetResource.RoutePolicySource != "" && resourceType != RoutePolicy {
+		return translatableerror.ArgumentCombinationError{
+			Args: []string{
```
jcvrabo, line 177: I think -s could also be used here, in plrinciple it wouldn't conflict with stack as it would be for a different resource

command/v7/labels_command.go
```diff
@@ -193,12 +199,20 @@ func (cmd LabelsCommand) validateFlags() error {
 		}
 	}
 
+	if cmd.RoutePolicySource != "" && cmd.canonicalResourceTypeForName() != RoutePolicy {
+		return translatableerror.ArgumentCombinationError{
+			Args: []string{
+				cmd.RequiredArgs.ResourceType, "--source",
```
jcvrabo, line 205: same applies here, for route policy source, in principle, you could also define -s without conflicting with stack

jcvrabo (approved):
don't forget to also make a pr for the V8 branch, perhaps after this one is cleared

jcvrabo (changes requested):
mistakenly submmited previous review with approval instead of requesting changes

actor/v7action/label.go
rkoster: Done in `4ba1b4f` — moved `resolveRoutePolicyGUID` to `route_policy.go`, alongside the other route-policy actor methods. `GetRoutePolicyLabels`/`UpdateRoutePolicyLabels` stay in `label.go` to match the existing convention: every `Get*/Update*Labels` wrapper lives in `label.go` and delegates resolution to the resource's own file (e.g. `GetRouteLabels`→`GetRoute`, `GetDomainLabels`→`GetDomainByName`). `resolveRoutePolicyGUID` is that resolver. Also dropped the now-unused `ccv3` import from `label.go`.

command/v7/route_policy_source_flags.go
rkoster: Done in `37374a1` — `resolveSource` is now a single org→space→app cascade. Each level resolves its GUID once (org → space within that org → app within that space), the most specific flag provided becomes the scope, and the result is built once as `fmt.Sprintf("cf:%s:%s", scope, guid)`. The three separate `--source-*` branches are gone. Verbose scope output, warning order, and the app-not-found TIP error are unchanged — covered by the existing command specs (84 route-policy specs green).

command/v7/route_policy_source_flags.go
rkoster: Right — done in `37374a1`. With the single-pass cascade, GUID resolution happens once per level inline, so `resolveOrgGUID`/`resolveSpaceGUID` are no longer needed and have been removed.

command/v7/set_label_command.go
rkoster: Done in `dbccb8c` — added `route-policy` to `set-label`'s `Resources()` list, placed after `route` to match the ordering already used by the `labels` command.

command/v7/unset_label_command.go
rkoster: Done in `dbccb8c` — same fix applied to `unset-label`'s `Resources()` list.

command/v7/remove_route_policy_command.go
rkoster: Done in `b4fcb44` — `--path` now has a `-p` short alias on `remove-route-policy`. No conflict here: the command's other shorts are `-n` (hostname) and `-f` (force).

command/v7/add_route_policy_command.go
rkoster: Done in `b4fcb44` — added the `-p` short alias for `--path` on `add-route-policy` as well (the only other short here is `-n` for hostname). I also tidied the pre-existing struct-tag alignment on the neighboring `--hostname` line so the file is gofmt-clean.

command/v7/label_updater.go
rkoster: Good thought, but `-s` isn't free on these commands: `set-label` and `unset-label` already bind `-s` to `--stack` (e.g. `cf set-label buildpack go_buildpack go=1.12 -s cflinuxfs4`).

go-flags (v1.6.1) validates the whole option set per command at parse time — not per resource type — so giving `--source` `short:"s"` makes the parser reject the command on every invocation:

```
option `-s, --source' uses the same short name as option `-s, --stack'
```

That error fires even when neither `-s` nor `--source` is passed, so it would break `set-label`/`unset-label` entirely (not just for non-route-policy resources). I've left `--source` without a short to preserve the existing `-s`/`--stack`. Happy to wire up a different (non-`s`) short here if you'd prefer one.

command/v7/labels_command.go
rkoster: Same constraint here as on set-/unset-label: `labels` already binds `-s` to `--stack`, and go-flags (v1.6.1) rejects two options sharing a short name across the whole command at parse time — independent of the positional resource type:

```
option `-s, --source' uses the same short name as option `-s, --stack'
```

Because the option set is validated on every invocation (the resource is a positional arg, not a separate subcommand, so `--stack` and `--source` share one option group), this would break `labels` even when `--source` isn't used. So I've left `--source` without a short. Can wire up a different (non-`s`) short if you'd prefer one.

jcvrabo (approved):
lgtm

api/cloudcontroller/ccversion/minimum_version.go
```diff
@@ -22,4 +22,10 @@ const (
 	MinVersionEmbeddedProcessInstances = "3.211.0"
 
 	MinVersionUpdateStack = "3.211.0"
+
+	// MinVersionRoutePolicies is a placeholder until the CAPI team confirms the
+	// version that introduces /v3/route_policies and the enforce_route_policies /
+	// route_policies_scope domain fields. Replace "3.999.0" with the real version
+	// once known. The test in minimum_version_test.go will keep failing until then.
+	MinVersionRoutePolicies = "3.221.0"
```
anujc25: Is this the final CAPI version? if yes, can you please update the comment here and the unit test file?

anujc25 (approved):
LGTM. Thanks for all these efforts.

api/cloudcontroller/ccversion/minimum_version.go
rkoster: Addressed in: https://github.com/cloudfoundry/cli/pull/3758/commits/60b892f6b88426f37a263fe0f1481b1e1e19e501

PR #3798, opened 2026-06-16
[v8][cherry-pick]: Fix CI client-creds integration tests

prkalle:
## Description of the Change

This PR is cherry-pick of https://github.com/cloudfoundry/cli/pull/3797

Fixed the CI client-creds integration tests

- Added--flake-attempts to `integration-tests-ci-client-creds` to retry transient BOSH Lite staging failures
- Also fix a pre-existing regression in `IsMinimumAPIVersionMet`: when the current API version is empty (as in the selfcontained Kubernetes auth test, which sets up config manually without calling `cf api`),`semver.Make("")` was called and returned "Version string empty", causing`cf apps` to fail. Guard against empty `current` by returning false (minimum not met, use fallback path) instead of erroring. The test failure was introduced by #3725 (embedded process instances for `cf apps`summary), which added the `IsMinimumAPIVersionMet(actor.Config.APIVersion(), ...)` call but did not account for contexts where APIVersion is unset.

## Why Is This PR Valuable?

This helps to fix the CI checks.

## Applicable Issues

## How Urgent Is The Change?

Fairly urgent to fix the CI checks

## Other Relevant Parties

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,13 @@ integration-selfcontained: build install-test-deps
 
 integration-tests: build integration-cleanup integration-isolated integration-push integration-global integration-selfcontained ## Run all isolated, push, selfcontained, and global integration tests
 
-integration-tests-ci-client-creds: build integration-cleanup integration-push integration-global integration-selfcontained
+integration-tests-ci-client-creds: build install-test-deps integration-cleanup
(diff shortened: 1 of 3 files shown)


anujc25 (approved):
LGTM. Thanks

PR #3799, opened 2026-06-17
[main] Regenerate counterfeiter fakes with v6.12.2

rkoster:
## Summary

- Runs `go generate ./...` to bring all generated fakes in sync with the counterfeiter version pinned in `go.mod` (`v6.12.2`)
- The main change is that `Invocations()` no longer emits per-field `RLock`/`RUnlock` calls — this is the output of the newer counterfeiter version
- 180 files changed, no functional difference — purely a codegen sync

## Why

The fakes in `main` were generated with an older version of counterfeiter. This causes unrelated fake changes to show up in feature PRs that happened to run `go generate`. This PR cleans up `main` so future PRs only show fakes changes that are actually relevant to their interface changes.


jcvrabo (approved):
This looks good to me, basically doing what it states (removal of the extra locks for the invocations method)

anujc25 (approved):
LGTM. (Can you also help create similar change on `v8` branch as well?)

PR #3800, opened 2026-06-18
[main] Fix(ci): prevent delete-env failure due to busy persistent disk on te…

prkalle:
## Description of the Change

This PR fixes the transient delete-env CI failure caused by busy persistent disk during BOSH Lite teardown

## Why Is This PR Valuable?

This PR ensures BOSH Lite environments are reliably torn down after integration tests complete. Previously, the `delete-env` job intermittently failed, leaving orphaned GCP resources (director VM, persistent disk) and dangling GCS state that required manual cleanup. It also caused the CI run to report a job failure even when all integration tests had passed.

## Applicable Issues

## How Urgent Is The Change?

Fairly urgent

## Other Relevant Parties


anujc25 (approved):
LGTM. Thanks

PR #3825, opened 2026-07-13
[v8] Regenerate counterfeiter fakes with v6.12.2

anujc25:
## Description of the Change

* Runs go generate ./... to bring all generated fakes in sync with the counterfeiter version pinned in go.mod (v6.12.2)
* The main change is that Invocations() no longer emits per-field RLock/RUnlock calls — this is the output of the newer counterfeiter version
* no functional difference — purely a codegen sync

similar to one done on main branch: https://github.com/cloudfoundry/cli/pull/3799

## Why Is This PR Valuable?

Cleanup only

## Applicable Issues

None

## How Urgent Is The Change?

Not urgent.


prkalle (approved):
LGTM.

PR #3838, opened 2026-08-06
[v8] Fix(curl): surface auth errors and avoid panic on nil HTTP response

prkalle:
## Description of the Change

This PR fixes `cf curl` command to surface auth errors and avoid panic on nil HTTP response

## Why Is This PR Valuable?

Thiis fixes the panic and surfaces underlying auth errors to users so the user sees a real message (e.g. "Bad credentials") 

## Applicable Issues

## How Urgent Is The Change?

Medium 

## Other Relevant Parties


anujc25 (dismissed):
LGTM. Thanks.

anujc25 (approved):

prkalle:
Integration tests passed on earlier run: https://github.com/cloudfoundry/cli/actions/runs/31523051859

PR #3859, opened 2026-09-03
Stream async job warnings as they arrive for delete commands

johha:
## Description of the Change

`cf delete`, `cf delete-org`, and `cf delete-space` kick off an asynchronous Cloud Controller job and then poll it to completion. Previously the CLI aggregated every warning returned across all poll ticks and dumped them at the very end, after the job reached a terminal state. Because Cloud Controller re-sends the same warnings on every poll, a single warning could be printed dozens of times.

This PR migrates those three commands from the aggregated `CloudControllerClient.PollJob` path to a streaming poll: the actor returns a `chan PollJobEvent` (via `PollJobToEventStream`) that `command/v7/shared.WaitForResult` drains, printing warnings **as they arrive** instead of at the end.

Key changes:
- **Streaming actors** - `DeleteApplicationByNameAndSpace`, `DeleteOrganization`, and `DeleteSpace` now return `(chan PollJobEvent, Warnings, error)`. For `cf delete -r`, the app-delete job and each route-delete job are merged into a single stream in the actor, so the whole operation streams through one channel.
- **Run-scoped warning dedup** - `WaitForResult` prints each distinct warning at most once per operation, collapsing the tick-by-tick repeats (including warnings that interleave across ticks).
- The "Waiting for the operation to complete" banner now ends with a newline so streamed warnings and progress dots start on their own line.

### Example: `cf delete-org` against a job that returns warnings on every poll

Before (warnings aggregated and repeated per poll tick):

```
$ cf delete-org my-org -f
Deleting org my-org as admin...
Deletion of service instance my-db is in progress.
Deletion of service instance my-db is in progress.
Endpoint deprecated: use the /v3 API instead.
Deletion of service instance my-db is in progress.
Endpoint deprecated: use the /v3 API instead.
... (repeats per poll tick for the same few warnings) ...
OK
```

After (each distinct warning shown once, streamed live with progress dots):

```
$ cf delete-org my-org -f
Deleting org my-org as admin...
Waiting for the operation to complete
Deletion of service instance my-db is in progress.
.Endpoint deprecated: use the /v3 API instead.
....Service broker returned a warning: quota not fully reclaimed.
.....

OK
```

`cf delete APP -r` (delete app + mapped routes) and `cf delete-space` behave the same way - warnings from the app job and each route/space job stream through as the jobs progress.

## Why Is This PR Valuable?

Users deleting orgs, spaces, or apps get feedback while the operation runs rather than a wall of duplicated text at the end. Warnings that Cloud Controller emits mid-operation (progress hints, content warnings) now surface promptly and legibly - each distinct message once - instead of being buffered until the job finishes or the poll window times out.

This is part of the ongoing effort to make recursive/async delete operations more intuitive (cloudfoundry/cloud_controller_ng#3589). The corresponding Cloud Controller work that emits these async-delete warnings is landing on the server side; this PR is the CLI half that surfaces them well.

## Applicable Issues

- Related: cloudfoundry/cloud_controller_ng#3589 - Make recursive delete operations more intuitive

## How Urgent Is The Change?

Not urgent 

## Other Relevant Parties

_None._

---
_This PR was drafted with the help of Claude (Opus)._


johha:
Backport for v8: #3860

command/v7/shared/result_waiter.go
```diff
@@ -13,16 +13,17 @@ func WaitForResult(stream chan v7action.PollJobEvent, ui command.UI, waitForComp
 	}
 
 	if waitForCompletion {
-		fmt.Fprint(ui.Writer(), "Waiting for the operation to complete")
+		fmt.Fprintln(ui.Writer(), "Waiting for the operation to complete")
```
prkalle, line 16: This newline moves the progress dots onto their own line, and `WaitForResult` is shared by **14** commands — the 3 delete commands added here plus 11 pre-existing service commands (`bind-service`, `unbind-service`, `bind-route-service`, `unbind-route-service`, `create-service`, `update-service`, `upgrade-service`, `delete-service`, `create-service-key`, `delete-service-key`, `cleanup-outdated-service-bindings`).

The unit tests were correctly updated for all of them:

```diff
-Say(`Waiting for the operation to complete\.\.\.\n`),
+Say(`Waiting for the operation to complete\n\.\.\.`),
```

But the  identical assertions in `integration/v7/isolated/` were not and this PR touches no files under `integration/`. These 11 assertions expect dots immediately after `complete`, which no longer holds:

Worth calling out in the PR description as well: the change alters output for 11 service commands, not only the three delete commands the title describes. Anyone parsing `cf bind-service --wait` output is affected.

command/v7/shared/result_waiter.go
```diff
@@ -13,16 +13,17 @@ func WaitForResult(stream chan v7action.PollJobEvent, ui command.UI, waitForComp
...
 		defer func() {
 			ui.DisplayNewline()
 			ui.DisplayNewline()
 		}()
 	}
 
+	seen := map[string]bool{}
 	for event := range stream {
-		ui.DisplayWarnings(event.Warnings)
+		ui.DisplayWarnings(dedupeSeenWarnings(event.Warnings, seen))
```
prkalle: 🤖 AI Review
---
🟡 **Should Fix**

`seen` starts empty here, but each of the three commands has already printed the actor's synchronous warnings via `cmd.UI.DisplayWarnings(warnings)` before calling `WaitForResult`. Those aren't in the dedup set, so a warning that appears both pre-job and during polling still prints twice — which is exactly the case the PR sets out to fix.

This is not hypothetical: the deprecation warning in your own before/after example rides on every CC response, including the initial `DELETE`, so it lands in both buckets. I ran `delete-org` with the same warning emitted synchronously and twice in the stream:

```
=== STDERR (warnings) ===
Endpoint deprecated: use the /v3 API instead.
Endpoint deprecated: use the /v3 API instead.

=== STDOUT ===
Deleting org some-org as banana...
Waiting for the operation to complete
...

OK
```

The two stream copies collapsed correctly — the pre-job one did not. The "After" block in the description shows this warning once, so the intended behaviour looks like one line total.

Seeding `seen` from the warnings the command already displayed would close it. Since the pre-job warnings are displayed by the caller rather than here, that likely means either having `WaitForResult` display them too, or passing the already-shown set in.

actor/v7action/application.go
```diff
@@ -48,39 +46,44 @@ func (actor Actor) DeleteApplicationByNameAndSpace(name, spaceGUID string, delet
...
+					if _, ok := err.(ccerror.ResourceNotFoundError); ok {
+						stream <- PollJobEvent{Warnings: Warnings(deleteWarnings)}
+						continue
+					}
+					stream <- PollJobEvent{Warnings: Warnings(deleteWarnings), Err: err}
+					return
+				}
+				stream <- PollJobEvent{Warnings: Warnings(deleteWarnings)}
+
+				for event := range actor.PollJobToEventStream(jobURL) {
```
prkalle, line 76: Polling each route's job to completion inside the per-route loop serializes what used to be concurrent. Currently, every `DeleteRoute` was issued first and the job URLs collected into `jobQueue`, so Cloud Controller worked on all of them at once; only then were they polled:

```go
for _, route := range routes {
    jobURL, ... := DeleteRoute(route.GUID)   // all issued up front
    jobQueue = append(jobQueue, jobURL)
}
for _, job := range jobQueue {
    PollJob(job)                              // then polled
}
```
Apps with several mapped routes will notice.

Issuing all the `DeleteRoute` calls first and then streaming each job's events in turn preserves the old concurrency while keeping the streaming behaviour — the events would still arrive in route order, just without blocking the next DELETE on the previous poll.

Minor related note: the route deletions are now side effects of a goroutine that only runs as the consumer drains the channel, so a caller that abandons the stream early leaves later routes undeleted. `WaitForResult` drains correctly today, so this isn't a live bug, but it's a sharper contract than the old synchronous version and worth a comment on the exported method.

command/v7/shared/result_waiter.go
```diff
@@ -36,3 +37,18 @@ func WaitForResult(stream chan v7action.PollJobEvent, ui command.UI, waitForComp
 
 	return true, nil
 }
+
+// dedupeSeenWarnings prints each distinct warning at most once per operation:
```
prkalle: 🤖 AI Review
---
🟢 **Suggestion**

Two scope notes on the dedup, neither blocking:

**One `seen` map spans every job in the operation.** For `cf delete APP -r` the app job and all route jobs share a single stream, so an identical warning string produced while deleting route A suppresses the one for route B. That's the right call when the text is generic (`Endpoint deprecated: ...`), but it silently drops multiplicity when a warning is per-resource yet not self-identifying. If CC's async-delete warnings always name the resource, this is a non-issue — worth a moment's thought against the server-side messages landing in cloudfoundry/cloud_controller_ng#3589.

**Dedup now also applies to the 11 service commands** that call `WaitForResult`. Previously a service command that repeated a warning across poll ticks printed it every tick; now it prints once. That's almost certainly an improvement, but it's a behaviour change outside the three commands in the title and isn't mentioned in the description.

prkalle:
@johha I noticed the integration tests are currently failing. Could you take a look at fixing those while I review the rest of the PR?

command/v7/shared/result_waiter.go
johha: Fixed. Updated all affected files under integration/v7/isolated/ with the same regex adjustment already applied to the unit tests.

actor/v7action/application.go
johha: Good catch. Fixed by restoring the two-phase approach: all DeleteRoute calls are now issued first before any polling starts, so CC can process them in parallel as before.

johha:
> @johha I noticed the integration tests are currently failing. Could you take a look at fixing those while I review the rest of the PR?

Thanks for the review @prkalle. The integration tests are skipped now. The CVE job seems to fail for almost all jobs. Also addressed all findings.

Also adjust the v8 PR #3860

integration/v7/isolated/update_service_command_test.go
```diff
@@ -183,7 +183,7 @@ var _ = Describe("update-service command", func() {
 				Expect(session.Out).To(SatisfyAll(
 					Say(`Updating service instance %s in org %s / space %s as %s\.\.\.\n`, serviceInstanceName, orgName, spaceName, username),
 					Say(`\n`),
-					Say(`Update of service instance %s complete\.\n`, serviceInstanceName),
+					Say(`Update of service instance %s complete\n\.\n`, serviceInstanceName),
```
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

Thanks for updating the integration tests — all 11 banner assertions are now correct. But the replace appears to have been applied to `complete\.` generally rather than to the banner string specifically, and it caught a second, unrelated message.

`Update of service instance X complete.` is the command's **final success line**, emitted here:

```go
// command/v7/update_service_command.go:74
cmd.UI.DisplayTextWithFlavor("Update of service instance {{.ServiceInstance}} complete.", cmd.serviceInstanceName())
```

This PR does not change that line, so the assertion should still be `complete\.\n`. As written it expects `complete` + newline + a literal dot + newline, which the command never produces.

Three sites in this file:

- **L186** — `It("can update tags, parameters and plan")`
- **L210** — `It("can update tags alone")`
- **L336** — `It("accepts the --wait flag to wait for completion")`

L186 and L210 are the clearest tell: neither invokes `--wait`, so `Waiting for the operation to complete` never appears in those tests at all — there was nothing there for this PR to affect.

All three want the original:

```suggestion
					Say(`Update of service instance %s complete\.\n`, serviceInstanceName),
```

Same pattern hit `upgrade-service` — commented separately.

Since integration tests need a live foundation, unit CI won't catch this; I built and ran the unit suites against `777b6cb` and they pass, so this is only visible on an integration run.

integration/v7/isolated/upgrade_service_command_test.go
```diff
@@ -151,9 +151,9 @@ var _ = Describe("upgrade-service command", func() {
 					Expect(session.Out).To(SatisfyAll(
 						Say(`Upgrading service instance %s in org %s / space %s as %s\.\.\.\n`, serviceInstanceName, orgName, spaceName, username),
 						Say(`\n`),
-						Say(`Waiting for the operation to complete\.+\n`),
+						Say(`Waiting for the operation to complete\n\.+`),
 						Say(`\n`),
-						Say(`Upgrade of service instance %s complete\.\n`, serviceInstanceName),
+						Say(`Upgrade of service instance %s complete\n\.\n`, serviceInstanceName),
```
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

Same collateral edit as in `update_service_command_test.go` (see the comment there for the full explanation).

`Upgrade of service instance X complete.` is the terminal success message, emitted at `command/v7/upgrade_service_command.go:65`, and is unaffected by this PR. The banner assertion two lines above is correct; this one should go back to the original:

```suggestion
						Say(`Upgrade of service instance %s complete\.\n`, serviceInstanceName),
```

command/v7/shared/result_waiter.go
prkalle: 🤖 AI Review
---
🟡 **Should Fix**

Raising this fresh — it was in my earlier pass but didn't make it into the review that was submitted, so it's new to you rather than something you skipped.

`seen` starts empty here, but each of the three commands has already printed the actor's synchronous warnings via `cmd.UI.DisplayWarnings(warnings)` before reaching `WaitForResult`. Those aren't in the dedup set, so a warning that appears both pre-job and during polling still prints twice — the exact duplication this PR sets out to remove.

Not hypothetical: the deprecation warning in your own before/after example rides on every CC response, including the initial `DELETE`, so it lands in both buckets. Against `777b6cb`, with that warning emitted once synchronously and once in the stream:

```
=== STDERR ===
Endpoint deprecated: use the /v3 API instead.
Endpoint deprecated: use the /v3 API instead.
```

The stream copy was correctly suppressed; the pre-job one wasn't. The "After" block in the description shows this warning once, which I read as the intended behaviour.

Closing it means the pre-job warnings need to go through the same `seen` set — either by having `WaitForResult` display them, or by seeding the set from what the caller already showed.

actor/v7action/application.go
```diff
@@ -48,39 +46,53 @@ func (actor Actor) DeleteApplicationByNameAndSpace(name, spaceGUID string, delet
...
 
-	for _, job := range jobQueue {
-		pollWarnings, err := actor.CloudControllerClient.PollJob(job)
-		allWarnings = append(allWarnings, pollWarnings...)
-		if err != nil {
-			return allWarnings, err
+		if deleteRoutes {
+			type routeJob struct {
+				jobURL   ccv3.JobURL
+				warnings Warnings
```
prkalle: 🤖 AI Review
---
🟢 **Suggestion**

The two-phase restore looks right — I confirmed all `DeleteRoute` calls are now issued before any polling begins, matching the pre-PR ordering.

One leftover: `routeJob.warnings` is written at the append below but never read. The warnings are already streamed on the preceding line via `stream <- PollJobEvent{Warnings: Warnings(deleteWarnings)}`, so the field is dead once set and `routeJob` can just be the job URL:

```go
var routeJobURLs []ccv3.JobURL
...
routeJobURLs = append(routeJobURLs, jobURL)
```

Harmless as-is, but a future reader may reasonably wonder whether those stored warnings were meant to be emitted later during the polling phase.

integration/v7/isolated/update_service_command_test.go
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

Thanks for updating the integration tests — all 11 banner assertions are now correct. But the replace appears to have been applied to `complete\.` generally rather than to the banner string specifically, and it caught a second, unrelated message.

`Update of service instance X complete.` is the command's **final success line**, emitted here:

```go
// command/v7/update_service_command.go:74
cmd.UI.DisplayTextWithFlavor("Update of service instance {{.ServiceInstance}} complete.", cmd.serviceInstanceName())
```

This PR does not change that line, so the assertion should still be `complete\.\n`. As written it expects `complete` + newline + a literal dot + newline, which the command never produces.

Three sites in this file:

- **L186** — `It("can update tags, parameters and plan")`
- **L210** — `It("can update tags alone")`
- **L336** — `It("accepts the --wait flag to wait for completion")`

L186 and L210 are the clearest tell: neither invokes `--wait`, so `Waiting for the operation to complete` never appears in those tests at all — there was nothing there for this PR to affect.

All three want the original:

```suggestion
					Say(`Update of service instance %s complete\.\n`, serviceInstanceName),
```

Same pattern hit `upgrade-service` — commented separately.

Since integration tests need a live foundation, unit CI won't catch this; I built and ran the unit suites against `777b6cb` and they pass, so this is only visible on an integration run.

integration/v7/isolated/upgrade_service_command_test.go
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

Same collateral edit as in `update_service_command_test.go` (see the comment there for the full explanation).

`Upgrade of service instance X complete.` is the terminal success message, emitted at `command/v7/upgrade_service_command.go:65`, and is unaffected by this PR. The banner assertion two lines above is correct; this one should go back to the original:

```suggestion
						Say(`Upgrade of service instance %s complete\.\n`, serviceInstanceName),
```

command/v7/shared/result_waiter.go
prkalle: 🤖 AI Review
---
🟡 **Should Fix**

Raising this fresh — it was in my earlier pass but didn't make it into the review that was submitted, so it's new to you rather than something you skipped.

`seen` starts empty here, but each of the three commands has already printed the actor's synchronous warnings via `cmd.UI.DisplayWarnings(warnings)` before reaching `WaitForResult`. Those aren't in the dedup set, so a warning that appears both pre-job and during polling still prints twice — the exact duplication this PR sets out to remove.

Not hypothetical: the deprecation warning in your own before/after example rides on every CC response, including the initial `DELETE`, so it lands in both buckets. Against `777b6cb`, with that warning emitted once synchronously and once in the stream:

```
=== STDERR ===
Endpoint deprecated: use the /v3 API instead.
Endpoint deprecated: use the /v3 API instead.
```

The stream copy was correctly suppressed; the pre-job one wasn't. The "After" block in the description shows this warning once, which I read as the intended behaviour.

Closing it means the pre-job warnings need to go through the same `seen` set — either by having `WaitForResult` display them, or by seeding the set from what the caller already showed.

actor/v7action/application.go
prkalle: 🤖 AI Review
---
🟢 **Suggestion**

Two-phase restore confirmed — all `DeleteRoute` calls now precede any polling.

Small leftover: `routeJob.warnings` is set below but never read, since those warnings are already streamed on the previous line. `routeJob` can collapse to just the job URL:

```go
var routeJobURLs []ccv3.JobURL
...
routeJobURLs = append(routeJobURLs, jobURL)
```

integration/v7/isolated/update_service_command_test.go
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

All 11 banner assertions look right now. But the replace also caught `Update of service instance X complete.` — that's the command's final success line (`update_service_command.go:74`), not the banner, and this PR doesn't change it. It should stay `complete\.\n`.

Three sites here: **L186**, **L210**, **L336**. L186 and L210 don't even pass `--wait`, so the banner never appears in them.

```suggestion
					Say(`Update of service instance %s complete\.\n`, serviceInstanceName),
```

Same slip in `upgrade_service_command_test.go` — commented there. Integration tests need a live foundation, so unit CI won't catch either.

integration/v7/isolated/upgrade_service_command_test.go
prkalle: 🤖 AI Review
---
🔴 **Must Fix**

Same collateral edit as in `update_service_command_test.go` (see the comment there for the full explanation).

`Upgrade of service instance X complete.` is the terminal success message, emitted at `command/v7/upgrade_service_command.go:65`, and is unaffected by this PR. The banner assertion two lines above is correct; this one should go back to the original:

```suggestion
						Say(`Upgrade of service instance %s complete\.\n`, serviceInstanceName),
```

command/v7/shared/result_waiter.go
prkalle: 🤖 AI Review
---
🟡 **Should Fix**

Raising this fresh — it was in my earlier pass but didn't make it into the review that was submitted, so it's new to you rather than something you skipped.

`seen` starts empty here, but each of the three commands has already printed the actor's synchronous warnings via `cmd.UI.DisplayWarnings(warnings)` before reaching `WaitForResult`. Those aren't in the dedup set, so a warning that appears both pre-job and during polling still prints twice — the exact duplication this PR sets out to remove.

Not hypothetical: the deprecation warning in your own before/after example rides on every CC response, including the initial `DELETE`, so it lands in both buckets. Against `777b6cb`, with that warning emitted once synchronously and once in the stream:

```
=== STDERR ===
Endpoint deprecated: use the /v3 API instead.
Endpoint deprecated: use the /v3 API instead.
```

The stream copy was correctly suppressed; the pre-job one wasn't. The "After" block in the description shows this warning once, which I read as the intended behaviour.

Closing it means the pre-job warnings need to go through the same `seen` set — either by having `WaitForResult` display them, or by seeding the set from what the caller already showed.

actor/v7action/application.go
prkalle: 🤖 AI Review
---
🟢 **Suggestion**

Two-phase restore confirmed — all `DeleteRoute` calls now precede any polling.

Small leftover: `routeJob.warnings` is set below but never read, since those warnings are already streamed on the previous line. `routeJob` can collapse to just the job URL:

```go
var routeJobURLs []ccv3.JobURL
...
routeJobURLs = append(routeJobURLs, jobURL)
```

prkalle (commented):
Thanks for making these changes, @johha!

I re-ran the integration tests in CI and noticed a few failures. I've also left a couple of comments on the code for you to review.

(Note: You can ignore the CVE-related CI failures, I will handle fixing those in a separate PR.)

